PR #11, opened 2019-10-23
Update to SerdeContext

kstich:
Updates the generator to use the `SerdeContext` instead of distinct utils.


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -214,7 +209,7 @@ private void writeSerde() {
                 .indent()
                     .write("output: $T,", applicationProtocol.getResponseType())
                     .write("protocol: string,")
-                    .write("utils?: Utils")
+                    .write("context?: SerdeContext")
```
srchase:
```suggestion
                    .write("context: SerdeContext")
```
Same here.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -205,7 +200,7 @@ private void writeSerde() {
                 .indent()
                     .write("input: $L,", inputType)
                     .write("protocol: string,")
-                    .write("utils?: Utils")
+                    .write("context?: SerdeContext")
```
srchase:
```suggestion
                    .write("context: SerdeContext")
```
This should not be optional.

PR #13, opened 2019-10-30
Improve complex serialization support

kstich:
This commit adds support for many new complex serializations. This
includes support for HTTP payloads, the prefixHeaders trait, and
nested lists, sets, maps, unions, and structures as well as their
respective HTTP binding location modifications. It also generalizes
the method for generating serde function names across the language.

This commit relies upon the changes in awslabs/smithy#193 which have not been published yet. To use it, you'll need to run `publishToMavenLocal` on an updated version of Smithy.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -17,29 +17,45 @@ package software.amazon.smithy.typescript.codegen.integration;
 
 import static software.amazon.smithy.model.knowledge.HttpBinding.Location;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.codegen.core.SymbolReference;
 import software.amazon.smithy.model.knowledge.HttpBinding;
 import software.amazon.smithy.model.knowledge.HttpBindingIndex;
+import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
+import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.neighbor.Walker;
 import software.amazon.smithy.model.shapes.BlobShape;
 import software.amazon.smithy.model.shapes.BooleanShape;
 import software.amazon.smithy.model.shapes.CollectionShape;
+import software.amazon.smithy.model.shapes.DocumentShape;
+import software.amazon.smithy.model.shapes.ListShape;
+import software.amazon.smithy.model.shapes.MapShape;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.NumberShape;
 import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.SetShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeIndex;
+import software.amazon.smithy.model.shapes.ShapeVisitor;
+import software.amazon.smithy.model.shapes.SimpleShape;
 import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.TimestampShape;
+import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.HttpTrait;
-import software.amazon.smithy.model.traits.TimestampFormatTrait;
+import software.amazon.smithy.model.traits.TimestampFormatTrait.Format;
 import software.amazon.smithy.typescript.codegen.ApplicationProtocol;
 import software.amazon.smithy.typescript.codegen.TypeScriptWriter;
+import software.amazon.smithy.utils.ListUtils;
 import software.amazon.smithy.utils.OptionalUtils;
 
 /**
@@ -57,6 +73,13 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
     @Override
     public void generateRequestSerializers(GenerationContext context) {
         TopDownIndex topDownIndex = context.getModel().getKnowledge(TopDownIndex.class);
+        OperationIndex operationIndex = context.getModel().getKnowledge(OperationIndex.class);
+        Walker shapeWalker = new Walker(context.getModel().getKnowledge(NeighborProviderIndex.class).getProvider());
+        SymbolProvider symbolProvider = context.getSymbolProvider();
+        TypeScriptWriter writer = context.getWriter();
+
+        // Track the shapes we need to generate sub-serializers for.
+        Set<Shape> serializingShapes = new HashSet<>();
         for (OperationShape operation : topDownIndex.getContainedOperations(context.getService())) {
             OptionalUtils.ifPresentOrElse(
                     operation.getTrait(HttpTrait.class),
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -179,36 +300,47 @@ private void generateOperationSerializer(GenerationContext context, OperationSha
...
         } else if (target instanceof CollectionShape) {
-            // TODO: fixme
-            throw new UnsupportedOperationException("Not yet implemented");
+            return getCollectionInputParam(context, bindingType, shape, member, target);
+        } else if (target instanceof StructureShape || target instanceof UnionShape || target instanceof MapShape) {
+            Symbol symbol = symbolProvider.toSymbol(target);
+            String value = ProtocolGenerator.getSerdeFunctionName(symbol, getName(), true) + "(";
+            // Collections map over the input entries, so use that.
+            // TODO See collection cleanup note
+            if (shape instanceof CollectionShape) {
```
mtdowling: This should be removed I think. Line 322 would cause this to never be called.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/ProtocolGenerator.java
```diff
@@ -119,6 +121,32 @@ default ApplicationProtocol resolveApplicationProtocol(
...
 
+    /**
+     * Generates the name of a serde function for shapes of a service.
+     *
+     * @param symbol The symbol the serde function is being generated for.
+     * @param protocol Name of the protocol being generated.
+     * @param isInput Whether the function is for input or output.
+     * @return Returns the generated function name.
+     */
+    static String getSerdeFunctionName(Symbol symbol, String protocol, boolean isInput) {
```
mtdowling: Consider just having two method: getSerFunctionName and getDeserFunctionName. Also, should this actually be a public API, or is this implementation specific to HTTP bindings?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -301,7 +495,56 @@ protected abstract void serializeDocument(
...
     );
 
-    private void generateOperationDeserializer(GenerationContext context, OperationShape operation, HttpTrait trait) {
+    /**
+     * Writes the code needed to serialize a structure in the document of a request.
+     *
+     * @param context The generation context.
+     * @param shape The structured shape being generated.
+     */
+    protected abstract void serializeDocumentStructure(
```
mtdowling: My intention for the `serializeDocument` method was that it would be a black box for handling how all input documents are serialized. How is it being used now? The same way? If so, how is stuff like `serializeDocumentUnion` different?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -301,7 +495,56 @@ protected abstract void serializeDocument(
...
+    protected abstract void serializeDocumentMap(
+            GenerationContext context,
+            MapShape shape
+    );
+
+    private void generateOperationDeserializer(
+            GenerationContext context,
+            OperationShape operation,
+            HttpTrait trait,
+            Set<ShapeId> generatedShapes
```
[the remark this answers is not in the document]
kstich: This would be used in the future to perform the same deduplication as serializers. Though, this may go away given other comments.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -126,6 +149,102 @@ private void generateOperationSerializer(GenerationContext context, OperationSha
...
+            private Void generateFunctionSignature(BiConsumer<GenerationContext, Shape> renderer, Shape shape) {
+                generatedShapes.add(shape.getId());
+                Symbol symbol = symbolProvider.toSymbol(shape);
+                // Use the shape name for the function name.
+                String methodName = ProtocolGenerator.getSerdeFunctionName(symbol, getName(), true);
+                // TODO Collection cleanup point
+                // There's a decent bit of this "collections are different" work in here, meaning
+                // there's likely to also be that in the implementations. There should be both a
+                // centralized way to check for and/or handle these differences.
+                if (!(shape instanceof CollectionShape)) {
```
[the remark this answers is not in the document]
kstich: Collections are handled with native types `Array` and `Set`, which aren't types to be imported. Their members may need importing, but would be imported in a later cycle of the visitor.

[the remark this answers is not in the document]
mtdowling: Not sure if you're using it, but using "$T" in the writer for a collection shape will automatically import the members.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
kstich: This is checking the container being generated for, not the target type like L322. This handles swapping the input parameter on things like this:

```ts
  input.map(entry =>
    serializeAws_restJson1_1Foo(entry, context)
  );
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
kstich: It's the same way (after I move a line back that I forgot to in R1.) The `serializeDocument[Shape]` functions handle the same black box of content, but for the methods that serialize for other document shapes instead of the main document shape.

```ts
// Validate we have input, then get the right serialization for the member target.
writer.write("input &&");
writer.openBlock("input.map(entry =>", ");", () -> {
    writer.write(getInputValue(context, Location.DOCUMENT, shape, shape.getMember(), target));
});
```

However, where the structure function is invoked could be a touch more similar to the current `serializeDocument`.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -57,14 +73,97 @@ public ApplicationProtocol getApplicationProtocol() {
...
                     operation.getTrait(HttpTrait.class),
                     httpTrait -> generateOperationSerializer(context, operation, httpTrait),
                     () -> LOGGER.warning(String.format(
                             "Unable to generate %s protocol request bindings for %s because it does not have an "
                             + "http binding trait", getName(), operation.getId())));
+
+            operationIndex.getInput(operation)
+                    .ifPresent(input -> serializingShapes.addAll(shapeWalker.walkShapes(input).stream()
+                            // Don't generate a sub-serializer for the actual input shape.
+                            .filter(s -> !input.equals(s))
```
mtdowling: What if the input shape is reused in the model and not just as input? Like a batch operation? Do we need something that makes it easy to know if an input or output shape is reused in other places in the model?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -57,14 +73,97 @@ public ApplicationProtocol getApplicationProtocol() {
...
+                            .collect(Collectors.toSet())));
         }
+
+        // TODO Abstraction point
+        // This should almost certainly be abstracted out to something else, because any protocol
+        // with a document body is going to need to generate sub-serializers.
+        // Generate the serializers for shapes within the operation closure.
+        serializingShapes.forEach(shape -> shape.accept(new ShapeVisitor.Default<Void>() {
```
mtdowling: This anonymous visit is a lot to put into a single method. Is this something that can be easily moved into an inner class (it could be non-static if that would be useful).

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -57,14 +73,97 @@ public ApplicationProtocol getApplicationProtocol() {
...
+            @Override
+            protected Void getDefault(Shape shape) {
+                return null;
+            }
+
+            private Void generateFunctionSignature(BiConsumer<GenerationContext, Shape> renderer, Shape shape) {
```
mtdowling: It's usually  best to put things like functions / callbacks in the last argument so that it's not awkward if you need to make the function span multiple lines and you still have trailing arguments to worry about.

Is "renderer" the body of the function? Maybe call it "functionBody" if so to make this more clear?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -57,14 +73,97 @@ public ApplicationProtocol getApplicationProtocol() {
...
+                        + "): any => {", "}", methodName, symbol, () -> {
+                    renderer.accept(context, shape);
+                });
+                writer.write("");
+                return null;
+            }
+
+            @Override
+            public Void listShape(ListShape shape) {
+                // TODO Collection cleanup point
```
mtdowling: Is this able to handle lists of lists?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
kstich: If the input shape is reused on a batch input, it will not be the explicit input shape of the batch operation and won't be filtered in that walk.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
kstich: Yes:

```
list FooListList {
    member: FooList
}

list FooList {
    member: Foo
}

structure Foo {
   bar: String,
   baz: String
}
```

would generate `serializeAws_restJson1_1FooListList`, `serializeAws_restJson1_1FooList`, and `serializeAws_restJson1_1Foo`.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
kstich: Yeah, I think that's part of getting this abstraction right. Unsure exactly what shape it's going to take so wanted to leave that to the author, even if that's me.

PR #17, opened 2019-11-08
Add filename to writer callback

mtdowling:
*Issue #, if available:*

*Description of changes:*


smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/runtimeConfig.browser.ts.template
```diff
@@ -21,5 +21,5 @@ export const ClientRuntimeConfiguration: Required<ClientRuntimeDependencies> = {
   utf8Decoder: fromUtf8,
   utf8Encoder: toUtf8,
   defaultUserAgent: defaultUserAgent(name, version),
-  ${3L@client_additional_runtime_configuration}
```
kstich: Are these indentations actually handled automatically or does the value sit at the beginning of the line?

mtdowling: It just copies and pastes into the string. because of this, and because you need to indent anyways, I just moved it to the beginning of the line to make it consistent when you actually intercept a section.

PR #22, opened 2019-11-13
Add support for SymbolProvider decoration

mtdowling:
*Issue #, if available:*

*Description of changes:*


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/TypeScriptIntegration.java
```diff
@@ -54,14 +54,14 @@ default Model preprocessModel(PluginContext context, TypeScriptSettings settings
      * <p>This can be used to customize the names of shapes, the package
      * that code is generated into, add dependencies, add imports, etc.
      *
-     * @param context Plugin context.
```
mtdowling: I simplified this API to remove the need for the PluginContext and instead provide the Model being generated. This gives the ability for the provided model to be filtered and/or transformed prior to invoking this API and removes any confusion as to which Model to use in a decorated SymbolProvider.

PR #30, opened 2019-11-19
Codegen fixes

srchase:
* use requestHandler instead of httpHandler
* remove protocol class
* add generic type arguments in Commands
* fix smithy-client dependency versiond

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -94,14 +94,8 @@ final class CommandGenerator implements Runnable {
         addInputAndOutputTypes();
 
         String name = symbol.getName();
-        writer.openBlock(
-                "export class $L extends $$Command<$L, $L, $L> {",
-                "}",
-                name,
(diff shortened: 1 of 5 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -94,7 +94,14 @@ public void run() {
         addInputAndOutputTypes();
 
         String name = symbol.getName();
-        writer.openBlock("export class $L extends $$Command<$L, $L> {", "}", name, inputType, outputType, () -> {
+        writer.openBlock(
```
kstich: Not everything for `openBlock` calls that go past the line length needs to be on its own line, [example](https://github.com/awslabs/smithy-typescript/blob/00622f9de152dddaa0f1e6921623ff215a8b84a4/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java#L95). This would be fine:

```
writer.openBlock("export class $L extends $$Command<$L, $L> {", "}", name, inputType, outputType, configType,
                () -> {...
```

Similar consolidation can be done in a couple other places.

PR #37, opened 2019-11-27
Add HTTP RPC protocol support

kstich:
This commit adds support for HTTP RPC protocols through the
`HttpRpcProtocolGenerator` abstract class. Its surface area
is similar to the `HttpBindingProtocolGenerator` but instead
generates POST-based RPC api calls.

Shared code between the two `ProtocolGenerator` implementations
has been moved to a utils class for reuse; implementations and
tests have been updated accordingly.

In the process of building this feature, several bugs and some
documentation issues were resolved.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpRpcProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpRpcProtocolGenerator.java
@@ -1,0 +1,302 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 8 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -0,0 +1,302 @@
...
+    private final Set<Shape> documentSerializingShapes = new TreeSet<>();
+    private final Set<Shape> documentDeserializingShapes = new TreeSet<>();
+
+    @Override
+    public ApplicationProtocol getApplicationProtocol() {
+        return ApplicationProtocol.createDefaultHttpApplicationProtocol();
+    }
+
+    /**
+     * Gets the content-type when for a request body.
```
JordonPhillips: -when

PR #39, opened 2019-11-29
add md5 config

srchase:
Adds md5 hashing constructor to runtime configs.


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
```diff
@@ -43,6 +43,8 @@
     AWS_CRYPTO_SHA256_BROWSER("dependencies", "@aws-crypto/sha256-browser", "^0.1.0-preview.1", true),
     AWS_SDK_HASH_NODE("dependencies", "@aws-sdk/hash-node", "^0.1.0-preview.5", true),
 
+    AWS_SDK_MD5_BROWSER("dependencies", "@aws-sdk/md-js", "^0.1.0-preview.8", true),
```
kstich: Should this be `@aws-sdk/md5-js` - notably with the `5`?

The browser config update has this:

```
import { Md5 } from "@aws-sdk/md5-js"
```

srchase: yes. fixed in e0fd36a65e8134f3039db2357d3284034f7d3241

srchase:
Currently, only S3 and SQS need md5 dependencies.  This should potentially be updated to only pull in md5 dependencies on clients that need them.

mtdowling:
If this is client specific, then it should go in the  AWS SDK for JS v3 repo.

PR #51, opened 2019-12-10
import SmithyException, isa and DocumentType from smithy-client instead of lib/smithy

srchase:
Import SmithyException, isa and DocumentType from smithy-client package.

Remove lib/smithy static content


srchase:
smithy-core package added in https://github.com/aws/aws-sdk-js-v3/pull/546

srchase:
This code could also get added to @aws-sdk/smithy-client.  I don't think smithy-core would be used independently of smithy-client, so may make sense to bundle them into one package.

srchase:
Approved in https://github.com/aws/aws-sdk-js-v3/pull/546 as "smithy-core"

PR #52, opened 2019-12-10
Updates for HTTP serde correctness and safety

kstich:
This commit includes updates to Http ProtocolGenerator components for
correctness and safety.

1. It fixes generation issues for the @httpPrefixHeaders trait in
the HttpBindingProtocolGenerator.
2. It updates SerdeContext to use an import alias for safety.
3. It only generates a local parsedOutput if it will be consumed.
4. It adds the Error type to the list of Typescript reserved words.

-----

Use Model instead of deprecated ShapeIndex. This commit will require using the latest version of Smithy published locally.

-----

Use better name for document body shape serde gen

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -36,6 +36,7 @@ import software.amazon.smithy.model.shapes.CollectionShape;
 import software.amazon.smithy.model.shapes.DocumentShape;
 import software.amazon.smithy.model.shapes.DoubleShape;
 import software.amazon.smithy.model.shapes.FloatShape;
+import software.amazon.smithy.model.shapes.MapShape;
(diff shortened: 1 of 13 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -303,11 +304,13 @@ private void writeHeaders(
             for (HttpBinding binding : bindingIndex.getRequestBindings(operation, Location.PREFIX_HEADERS)) {
                 String memberName = symbolProvider.toMemberName(binding.getMember());
                 writer.openBlock("if (input.$L !== undefined) {", "}", memberName, () -> {
-                    Shape target = index.getShape(binding.getMember().getTarget()).get();
+                    MapShape prefixMap = index.getShape(binding.getMember().getTarget()).get().asMapShape().get();
```
mtdowling: Lots of this can use `expectShape` rather than calling get on an optional. It should also be migrated to making calls directly on Model rather than going through ShapeIndex.

PR #90, opened 2020-01-11
Allow payload serialization to be overridden

kstich:
This commit adds support for implementing protocols to override the
serialization behavior of a member marked with @httpPayload. This is
useful for protocols that work with intermediate objects that need to
be altered at the service level. Adding a top level xmlns attribute,
for example.

This commit also fixes various bugs found in testing behavior developed
on top of the main change.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -362,23 +362,16 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
 
         if (!documentBindings.isEmpty()) {
             // Write the default `body` property.
-            context.getWriter().write("let body: any = {};");
+            writer.write("let body: any = {};");
             serializeInputDocument(context, operation, documentBindings);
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -549,6 +542,41 @@ protected abstract void serializeInputDocument(
...
+     * Writes the code needed to serialize the input payload of a request.
+     *
+     * <p>Implementations of this method are expected to set a value to the
+     * {@code body} variable that will be serialized as the request body.
+     * This variable will already be defined in scope.
+     *
+     * <p>For example:
+     *
+     * <pre>{@code
+     * if (input.body !== undefined) {
```
mtdowling: What is `input`?

kstich:
The only other clear way to handle the scope that this change allows would be to add a single protected method to allow for arbitrarily manipulating the value currently being set in to `body`. This would be a function that does no work by default.

Its purpose would be a little less clear, but there's potentially less chance for a downstream consumer to miss-handle processing of Payloads.

mtdowling:
Ok. I'm still not thrilled about the API surface area, but LGTM for now.

Issue #80, opened 2020-01-07
Duplicate identifier in models/index.ts


trivikr:
**Describe the bug**
A "`Duplicate identifier`" error is thrown in some generated clients:
* https://github.com/aws/aws-sdk-js-v3/issues/664#issuecomment-571402554
* https://github.com/aws/aws-sdk-js-v3/issues/664#issuecomment-571407651

**To Reproduce**
* generate iot client using model defined in https://github.com/aws/aws-sdk-js-v3/blob/65740e3548fbacf93ff9b07fc379fe3731c622c1/codegen/sdk-codegen/aws-models/iot.2015-05-28.json 
* Run `tsc` on it (TypeScript compilation)
* Notice that the following error is thrown:
```console
models/index.ts(10350,3): error TS2300: Duplicate identifier 'CONTINUOUS'.
models/index.ts(10351,3): error TS2300: Duplicate identifier 'SNAPSHOT'.
models/index.ts(11255,3): error TS2300: Duplicate identifier 'CONTINUOUS'.
models/index.ts(11256,3): error TS2300: Duplicate identifier 'SNAPSHOT'.
```

**Expected behavior**
TypeScript compilation succeeds without any issue

trivikr:
This happens because the enum `TargetSelection` is exported twice in the [client-iot/models/index.ts](https://github.com/aws/aws-sdk-js-v3/blob/65740e3548fbacf93ff9b07fc379fe3731c622c1/clients/client-iot/models/index.ts)
* First time https://github.com/aws/aws-sdk-js-v3/blob/smithy-codegen/clients/client-iot/models/index.ts#L10349
* Second time https://github.com/aws/aws-sdk-js-v3/blob/smithy-codegen/clients/client-iot/models/index.ts#L11254

trivikr:
This happens because the same `TargetSelection` is defined under `om.amazonaws.iot.laser` as well as `com.amazonaws.iot.otamanager`

Model: https://github.com/aws/aws-sdk-js-v3/blob/65740e3548fbacf93ff9b07fc379fe3731c622c1/codegen/sdk-codegen/aws-models/iot.2015-05-28.json#L21144,L21169

Proposed solution: break down models/index.ts into smaller files as suggested in https://github.com/awslabs/smithy-typescript/issues/81

mtdowling:
> Proposed solution: break down models/index.ts into smaller files as suggested in #81

[...]

PR #105, opened 2020-01-21
Remove duplicate shapes

srchase:
When generating TypeScript shapes, attempt to first condense any matching shapes to eliminate duplicate generated shapes. Throws a CodegenException if shapes exist that cannot be condensed.

Fixes: #80

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
@@ -144,23 +144,11 @@ class CodegenVisitor extends ShapeVisitor.Default<Void> {
         // Generate models that are connected to the service being generated.
         LOGGER.fine("Walking shapes from " + service.getId() + " to find shapes to generate");
         Set<Shape> serviceShapes = new TreeSet<>(new Walker(nonTraits).walkShapes(service));
-        Map<String, Shape> shapeMap = new TreeMap<>();
 
-        // Check for colliding shapes and prune non-unique shapes
-        for (Shape shape : serviceShapes) {
-            String shapeReference = shape.getType().toString() + shape.getId().asRelativeReference();
+        // Condense duplicate shapes
+        Map<String, Shape> shapeMap = condenseShapes(serviceShapes);
 
-            if (shapeMap.containsKey(shapeReference)) {
-                Shape knownShape = shapeMap.get(shapeReference);
-                if (!shapesMatch(shape, knownShape)) {
-                    throw new CodegenException(("Shape Collision: cannot condense " + shape.toString() + " and "
-                            + knownShape.toString()));
-                }
-            } else {
-                shapeMap.put(shapeReference, shape);
-            }
-        }
-        // Generate models from pruned shapes
+        // Generate models from condensed shapes
         for (Shape shape : shapeMap.values()) {
             shape.accept(this);
         }
@@ -306,10 +294,30 @@ class CodegenVisitor extends ShapeVisitor.Default<Void> {
         return null;
     }
 
-    private boolean shapesMatch(Shape shape, Shape otherShape) {
+    private Map<String, Shape> condenseShapes(Set<Shape> shapes) {
+        Map<String, Shape> shapeMap = new TreeMap<>();
+
+        // Check for colliding shapes and prune non-unique shapes
+        for (Shape shape : shapes) {
+            String shapeReference = shape.getType().toString() + shape.getId().asRelativeReference();
+
+            if (shapeMap.containsKey(shapeReference)) {
+                Shape knownShape = shapeMap.get(shapeReference);
+                if (isShapeCollision(shape, knownShape)) {
+                    throw new CodegenException(("Shape Collision: cannot condense " + shape + " and " + knownShape));
+                }
+            } else {
+                shapeMap.put(shapeReference, shape);
+            }
+        }
+
+        return shapeMap;
+    }
+
+    private boolean isShapeCollision(Shape shape, Shape otherShape) {
         // Check names match.
         if (!shape.getId().getName().equals(otherShape.getId().getName())) {
-            return false;
+            return true;
         }
 
         // Check traits match.
@@ -318,13 +326,13 @@ class CodegenVisitor extends ShapeVisitor.Default<Void> {
         for (ShapeId trait : traits.keySet()) {
             // Ignore box trait when comparing.
             if (!otherTraits.containsKey(trait) && !trait.getName().equals("box")) {
-                return false;
+                return true;
             }
         }
         for (ShapeId trait : otherTraits.keySet()) {
             // Ignore box trait when comparing.
             if (!traits.containsKey(trait) && !trait.getName().equals("box")) {
-                return false;
+                return true;
             }
         }
 
@@ -333,15 +341,15 @@ class CodegenVisitor extends ShapeVisitor.Default<Void> {
         Collection<MemberShape> otherMemberShapes = otherShape.members();
         for (MemberShape memberShape : memberShapes) {
             if (!otherMemberShapes.stream().anyMatch(s -> s.getMemberName().contains(memberShape.getMemberName()))) {
-                return false;
+                return true;
             }
         }
         for (MemberShape otherMemberShape : otherMemberShapes) {
             if (!memberShapes.stream().anyMatch(s -> s.getMemberName().contains(otherMemberShape.getMemberName()))) {
-                return false;
+                return true;
             }
         }
 
-        return true;
+        return false;
     }
 }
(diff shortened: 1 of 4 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -139,7 +144,26 @@ void execute() {
         // Generate models that are connected to the service being generated.
         LOGGER.fine("Walking shapes from " + service.getId() + " to find shapes to generate");
         Set<Shape> serviceShapes = new TreeSet<>(new Walker(nonTraits).walkShapes(service));
-        serviceShapes.forEach(shape -> shape.accept(this));
+        Map<String, Shape> shapeMap = new TreeMap<>();
```
mtdowling: I think moving this into its own method would make it more readable. You could then just return the created shape map.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -139,7 +144,26 @@ void execute() {
...
+
+        // Check for colliding shapes and prune non-unique shapes
+        for (Shape shape : serviceShapes) {
+            String shapeReference = shape.getType().toString() + shape.getId().asRelativeReference();
+
+            if (shapeMap.containsKey(shapeReference)) {
+                Shape knownShape = shapeMap.get(shapeReference);
+                if (!shapesMatch(shape, knownShape)) {
+                    throw new CodegenException(("Shape Collision: cannot condense " + shape.toString() + " and "
```
mtdowling: No need to use `toString` on the colliding shapes

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -281,4 +305,43 @@ public Void serviceShape(ServiceShape shape) {
 
         return null;
     }
+
+    private boolean shapesMatch(Shape shape, Shape otherShape) {
```
mtdowling: Rename to `isShapeCollision`

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -281,4 +305,43 @@ public Void serviceShape(ServiceShape shape) {
...
+        // Check names match.
+        if (!shape.getId().getName().equals(otherShape.getId().getName())) {
+            return false;
+        }
+
+        // Check traits match.
+        Map<ShapeId, Trait> traits = shape.getAllTraits();
+        Map<ShapeId, Trait> otherTraits = otherShape.getAllTraits();
+        for (ShapeId trait : traits.keySet()) {
+            // Ignore box trait when comparing.
```
mtdowling: Why ignore the box trait?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -281,4 +305,43 @@ public Void serviceShape(ServiceShape shape) {
...
     }
+
+    private boolean shapesMatch(Shape shape, Shape otherShape) {
+        // Check names match.
+        if (!shape.getId().getName().equals(otherShape.getId().getName())) {
+            return false;
+        }
+
+        // Check traits match.
+        Map<ShapeId, Trait> traits = shape.getAllTraits();
```
mtdowling: This comparison looks like it could just be: `shape.getAllTraits().equals(otherShape.getAllTraits())`.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -281,4 +305,43 @@ public Void serviceShape(ServiceShape shape) {
...
+            }
+        }
+        for (ShapeId trait : otherTraits.keySet()) {
+            // Ignore box trait when comparing.
+            if (!traits.containsKey(trait) && !trait.getName().equals("box")) {
+                return false;
+            }
+        }
+
+        // Check members match.
```
mtdowling: This also looks like it could just be `shape.members().equals(otherShape.members())`

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/CodegenVisitorTest.java
```diff
@@ -98,6 +100,52 @@ public void generatesServiceClients() {
...
+
+        Assertions.assertTrue(manifest.hasFile("models/index.ts"));
+        assertThat(manifest.getFileString("models/index.ts").get(),
+                containsString("export interface Bar {\n" +
+                        "  __type?: \"Bar\";\n" +
+                        "  baz: string | undefined;\n" +
+                        "}"));
+    }
+
+
```
mtdowling: nit: extra new line

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Doing the check with `shape.members().equals(otherShape.members())` way won't ignore the namespaces:

```
namespace smithy.example

structure Bar {
    @required
    baz: String
}

namespace another.example

structure Bar {
    @required
    baz: String
}
```

The `baz` members are different, so it returns `false`.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: The box trait needs ignored for examples like this:
https://github.com/aws/aws-sdk-js-v3/blob/smithy-codegen/models/iot.2015-05-28.json#L10612:L10620

Those would generate the same TypeScript, so we don't want to throw that they're colliding shapes in this instance.  Since the box trait only applies to primitives, there won't have an issue with duplicate model code being generated.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Fixed in cf9d401

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Fixed in cf9d401

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Fixed in cf9d401

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
mtdowling: :+1:

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
mtdowling: :+1: Can this be simplified to:

```java
Map<ShapeId, Trait> traitsA = new HashMap<>(shape.getAllTraits());
Map<ShapeId, Trait> traitsB = new HashMap<>(shape.getAllTraits());
// Ignore the box trait since it has no effect in JavaScript.
traitsA.remove(BoxTrait.ID);
traitsB.remove(BoxTrait.ID);
if (!traitsA.equals(traitsB)) {
    return false;
}
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Fixed in 66e0ca0

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
srchase: Fixed in 66e0ca0

PR #118, opened 2020-02-04
Initialize body, but do not set to empty object

srchase:
Body should only be initialized, but not set to empty object.

Empty objects that do not get replaced with actual body content cause errors when sending requests.

Part of fix for https://github.com/aws/aws-sdk-js-v3/issues/838

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -374,13 +374,13 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
 
         if (!documentBindings.isEmpty()) {
             // Write the default `body` property.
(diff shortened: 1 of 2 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -374,13 +374,13 @@ private void writeHeaders(
 
         if (!documentBindings.isEmpty()) {
             // Write the default `body` property.
-            writer.write("let body: any = {};");
+            writer.write("let body: any;");
```
kstich: This needs an equivalent fix needed in the `HttpRpcProtocolGenerator`.

srchase: Added in 0bf9184.

I verified that `HttpRequest` will properly handle this as part of testing with aws/aws-sdk-js-v3/pull/844 to confirm fix for aws/aws-sdk-js-v3/issues/838

PR #120, opened 2020-02-05
Export models in index.ts

srchase:
Resolves: aws/aws-sdk-js-v3/issues/847

Export all shapes from models/index.ts in root index.ts


srchase:
After re-generating all the clients (via ```node scripts/generate-clients/index.js -m models/```), I ran ```yarn pretest:all``` and confirmed all packages and clients compiled properly.

PR #124, opened 2020-02-10
Add serde support for application/json String

kstich:
This commit adds support for serializing and deserializing String
shapes as the `LazyJsonString` when they are marked with the `@mediatype`
trait with the value of `"application/json"`.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
@@ -18,6 +18,7 @@ package software.amazon.smithy.typescript.codegen;
 import static java.lang.String.format;
 
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.CodegenException;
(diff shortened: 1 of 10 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
```diff
@@ -235,9 +237,24 @@ public Symbol operationShape(OperationShape shape) {
...
+        if (enumTrait.isPresent()) {
+            return createEnumSymbol(shape, enumTrait.get());
+        }
+
+        // Handle media type generation, defaulting to a string.
+        Optional<MediaTypeTrait> mediaTypeTrait = shape.getTrait(MediaTypeTrait.class);
+        if (mediaTypeTrait.isPresent()) {
+            String mediaType = mediaTypeTrait.get().getValue();
+            switch (mediaType) {
+                case "application/json":
```
mtdowling: - [ ] You could also use the wrapper for any media type that ends with "+json": https://tools.ietf.org/html/rfc6839#section-3.1
- [ ] It looks like this check if performed in a couple places. If we make this work with "+json", then an "isJsonMediaType" utility method or something similar should probably be added.
- [ ] I think this should have a test too.

PR #132, opened 2020-02-12
fix: set errorCode to parsedBody.code in default case

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/issues/851

*Description of changes:*
* set errorCode to parsedBody.name in default case
* removes "UnknownError" assignment, as it's initialized on line 268

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
@@ -295,7 +295,7 @@ final class HttpProtocolGeneratorUtils {
 
                         // Get the protocol specific error location for retrieving contents.
                         String errorLocation = bodyErrorLocationModifier.apply(context, "parsedBody");
-                        writer.write("errorCode = errorCode || \"UnknownError\";");
+                        writer.write("errorCode = errorCode;");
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -295,7 +295,7 @@ static void generateCollectBodyString(GenerationContext context) {
 
                         // Get the protocol specific error location for retrieving contents.
                         String errorLocation = bodyErrorLocationModifier.apply(context, "parsedBody");
-                        writer.write("errorCode = errorCode || \"UnknownError\";");
+                        writer.write("errorCode = parsedBody.code || errorCode;");
```
kstich: This should be updated to use the `errorLocation` variable [like below.](https://github.com/awslabs/smithy-typescript/pull/132/files#diff-b575367de812e857bb3dddf79f734badR302)

It should also check for `.Code` in addition to `.code`

trivikr: Done in https://github.com/awslabs/smithy-typescript/pull/132/commits/6ec5660bce79511c324f585e660ccc3da9f5f9f2

PR #142, opened 2020-03-04
Add support for generating HTTP protocol tests

kstich:
This commit adds support for generating protocol tests based on the
`smithy.test#httpRequestTests` and `smithy.test#httpResponseTests`
traits. Any service that has these traits will have a file generated
in ./tests/functional named after the protocol that is being tested.
These generated tests use the `jest` library and its expect/fail
methods, along with custom mock handlers for testing the requests
and responses.

A custom node/shape visitor is needed to handle TS specific output
typing in the generated code. This handles properly generating Date
types, booleans, sets, and more.

Individual body comparators are included as stubs to be added when
the specific `bodyMediaType` they support are encountered. Some new
dependencies are added when handling query, form-urlencoded, or XML
components.

<details>
  <summary>Expand to view generated test sample from update test files.</summary>

```
import { WeatherClient } from "../../WeatherClient";
import { GetCityCommand } from "../../commands/GetCityCommand";
import { GetCityImageCommand } from "../../commands/GetCityImageCommand";
import { ListCitiesCommand } from "../../commands/ListCitiesCommand";
import { NoSuchResource } from "../../models/index";
import { buildQueryString } from "@aws-sdk/querystring-builder";
import { HttpHandlerOptions, HeaderBag } from "@aws-sdk/types";
import { HttpHandler, HttpRequest, HttpResponse } from "@aws-sdk/protocol-http";
import { Readable } from 'stream';

/**
 * Throws an expected exception that contains the serialized request.
 */
class EXPECTED_REQUEST_SERIALIZATION_ERROR {
    constructor(readonly request: HttpRequest) {}
}

/**
 * Throws an EXPECTED_REQUEST_SERIALIZATION_ERROR error before sending a
 * request. The thrown exception contains the serialized request.
 */
class RequestSerializationTestHandler implements HttpHandler {
    handle(
        request: HttpRequest,
        options: HttpHandlerOptions
    ): Promise<{ response: HttpResponse }> {
        return Promise.reject(new EXPECTED_REQUEST_SERIALIZATION_ERROR(request));
    }
}

/**
 * Returns a resolved Promise of the specified response contents.
 */
class ResponseDeserializationTestHandler implements HttpHandler {
    isSuccess: boolean;
    code: number;
    headers: HeaderBag;
    body: Readable;

    constructor(
        isSuccess: boolean,
        code: number,
        headers?: HeaderBag,
        body?: String
    ) {
        this.isSuccess = isSuccess;
        this.code = code;
        if (headers === undefined) {
            this.headers = {};
        } else {
            this.headers = headers;
        }
        if (body === undefined) {
          body = "";
        }
        this.body = Readable.from([body]);
    }

    handle(
        request: HttpRequest,
        options: HttpHandlerOptions
    ): Promise<{ response: HttpResponse }> {
        return Promise.resolve({
            response: {
                statusCode: this.code,
                headers: this.headers,
                body: this.body
            }
        });
    }
}

interface comparableParts {
  [ key: string ]: string
}

/**
 * Generates a standard map of un-equal values given input parts.
 */
const compareParts = (expectedParts: comparableParts, generatedParts: comparableParts) => {
  const unequalParts: any = {};
  Object.keys(expectedParts).forEach(key => {
    if (generatedParts[key] === undefined) {
      unequalParts[key] = { exp: expectedParts[key], gen: undefined };
    } else if (generatedParts[key] !== expectedParts[key]) {
      unequalParts[key] = { exp: expectedParts[key], gen: generatedParts[key] };
    }
  });

  Object.keys(generatedParts).forEach(key => {
    if (expectedParts[key] === undefined) {
      unequalParts[key] = { exp: undefined, gen: generatedParts[key] };
    } else if (expectedParts[key] !== generatedParts[key]) {
      unequalParts[key] = { exp: expectedParts[key], gen: generatedParts[key] };
    }
  });

  if (Object.keys(unequalParts).length !== 0) {
    return unequalParts;
  }
  return undefined;
};

/**
 * Does something
 */
it("WriteNoSuchResourceAssertions", async () => {
  const client = new WeatherClient({
    requestHandler: new ResponseDeserializationTestHandler(
      false,
      404,
      undefined,
      `{
          "resourceType": "City",
          "message": "Your custom message"
      }`,
    )
  });

  const params: any = {};
  const command = new GetCityImageCommand(params);

  try {
    await client.send(command);
  } catch (err) {
    if (!NoSuchResource.isa(err)) {
      console.log(err);
      fail(`Expected a NoSuchResource to be thrown, got ${err.name} instead`);
    }
    const r: any = err;
    expect(r['$metadata'].httpStatusCode).toBe(404);
    const paramsToValidate: any = {
      "resourceType": "City",
      "message": "Your custom message"
    };
    Object.keys(paramsToValidate).forEach(param => {
      expect(r[param]).toBeDefined();
      expect(r[param]).toBe(paramsToValidate[param]);
    });
    return;
  }
  fail('Expected an exception to be thrown from response');
});

/**
 * Does something
 */
it("WriteListCitiesAssertions", async () => {
  const client = new WeatherClient({
    requestHandler: new RequestSerializationTestHandler()
  });

  const command = new ListCitiesCommand(
  {
    pageSize:
    50,

  },

  );
  try {
    await client.send(command);
    fail('Expected an EXPECTED_REQUEST_SERIALIZATION_ERROR to be thrown');
  } catch (err) {
    if (!(err instanceof EXPECTED_REQUEST_SERIALIZATION_ERROR)) {
      fail(err);
    }
    const r = err.request;
    expect(r.method).toBe("GET");
    expect(r.path).toBe("/cities");

    expect(r.query["nextToken"]).toBeUndefined();

    const queryString = buildQueryString(r.query);
    expect(queryString).toContain("pageSize=50");

    expect(r.body).toBeDefined();
    expect(r.body).toBe("");
  }
});

/**
 * Does something
 */
it("WriteGetCityAssertions", async () => {
  const client = new WeatherClient({
    requestHandler: new RequestSerializationTestHandler()
  });

  const command = new GetCityCommand(
  {
    cityId:
    "123",

  },

  );
  try {
    await client.send(command);
    fail('Expected an EXPECTED_REQUEST_SERIALIZATION_ERROR to be thrown');
  } catch (err) {
    if (!(err instanceof EXPECTED_REQUEST_SERIALIZATION_ERROR)) {
      fail(err);
    }
    const r = err.request;
    expect(r.method).toBe("GET");
    expect(r.path).toBe("/cities/123");

    expect(r.body).toBeDefined();
    expect(r.body).toBe("");
  }
});

/**
 * Does something
 */
it("WriteGetCityResponseAssertions", async () => {
  const client = new WeatherClient({
    requestHandler: new ResponseDeserializationTestHandler(
      true,
      200,
      undefined,
      `{
          "name": "Seattle",
          "coordinates": {
              "latitude": 12.34,
              "longitude": -56.78
          },
          "city": {
              "cityId": "123",
              "name": "Seattle",
              "number": "One",
              "case": "Upper"
          }
      }`,
    )
  });

  const params: any = {};
  const command = new GetCityCommand(params);

  let r: any;
  try {
    r = await client.send(command);
  } catch (err) {
    fail('Expected a valid response to be returned, got err.');
  }
  expect(r['$metadata'].httpStatusCode).toBe(200);
  const paramsToValidate: any = {
    "name": "Seattle",
    "coordinates": {
      "latitude": 12.34,
      "longitude": -56.78
    },
    "city": {
      "cityId": "123",
      "name": "Seattle",
      "number": "One",
      "case": "Upper"
    }
  };
  Object.keys(paramsToValidate).forEach(param => {
    expect(r[param]).toBeDefined();
    expect(r[param]).toBe(paramsToValidate[param]);
  });
});

```
</details>

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
@@ -1,0 +1,490 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.typescript.codegen;
+
+import static java.lang.String.format;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.logging.Logger;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.OperationIndex;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.node.ArrayNode;
+import software.amazon.smithy.model.node.BooleanNode;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.NodeVisitor;
+import software.amazon.smithy.model.node.NullNode;
+import software.amazon.smithy.model.node.NumberNode;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.shapes.CollectionShape;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.traits.IdempotencyTokenTrait;
+import software.amazon.smithy.protocoltests.traits.HttpMessageTestCase;
+import software.amazon.smithy.protocoltests.traits.HttpRequestTestCase;
+import software.amazon.smithy.protocoltests.traits.HttpRequestTestsTrait;
+import software.amazon.smithy.protocoltests.traits.HttpResponseTestCase;
+import software.amazon.smithy.protocoltests.traits.HttpResponseTestsTrait;
+import software.amazon.smithy.utils.IoUtils;
+
+/**
+ * Generates HTTP protocol test cases to be run using Jest.
+ *
+ * <p>Protocol tests are defined for HTTP protocols using the
+ * {@code smithy.test#httpRequestTests} and {@code smithy.test#httpResponseTests}
+ * traits. When found on operations or errors attached to operations, a
+ * protocol test case will be generated that asserts that the protocol
+ * serialization and deserialization code creates the correct HTTP requests
+ * and responses for a specific set of parameters.
+ *
+ * TODO: try/catch and if/else are still cumbersome with TypeScriptWriter.
+ */
+final class HttpProtocolTestGenerator implements Runnable {
+
+    private static final Logger LOGGER = Logger.getLogger(HttpProtocolTestGenerator.class.getName());
+    private static final String TEST_CASE_FILE_TEMPLATE = "tests/functional/%s.spec.ts";
+
+    private final TypeScriptSettings settings;
+    private final Model model;
+    private final ServiceShape service;
+    private final SymbolProvider symbolProvider;
+    private final Symbol serviceSymbol;
+    private final Set<String> additionalStubs = new TreeSet<>();
+
+    /** Vends a TypeScript IFF it's needed. */
+    private final TypeScriptDelegator delegator;
+
+    /** The TypeScript writer that's only allocated once if needed. */
+    private TypeScriptWriter writer;
+
+    HttpProtocolTestGenerator(
+            TypeScriptSettings settings,
+            Model model,
+            SymbolProvider symbolProvider,
+            TypeScriptDelegator delegator
+    ) {
+        this.settings = settings;
+        this.model = model;
+        this.service = settings.getService(model);
+        this.symbolProvider = symbolProvider;
+        this.delegator = delegator;
+        serviceSymbol = symbolProvider.toSymbol(service);
+    }
+
+    @Override
+    public void run() {
+        OperationIndex operationIndex = model.getKnowledge(OperationIndex.class);
+        TopDownIndex topDownIndex = model.getKnowledge(TopDownIndex.class);
+
+        for (OperationShape operation : topDownIndex.getContainedOperations(service)) {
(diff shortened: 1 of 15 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,484 @@
...
+            operation.getTrait(HttpResponseTestsTrait.class).ifPresent(trait -> {
+                for (HttpResponseTestCase testCase : trait.getTestCases()) {
+                    onlyIfProtocolMatches(testCase, () -> generateResponseTest(operation, testCase));
+                }
+            });
+            // 3. Generate test cases for each error.
+            operationIndex.getErrors(operation).stream()
+                    .filter(FunctionalUtils.not(errorsGenerated::contains))
+                    .forEach(error -> {
+                        errorsGenerated.add(error);
```
mtdowling: It looks like this ensures that a given error only generates a single test case. This works for testing the error, but it doesn't provide an end-to-end test that ensures an operation's output parser correctly identifies errors (i.e., it finds the right headers/body members/whatever and then deserializes an error instead of the response). How hard would that be to support?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,484 @@
...
+            case "application/json":
+                additionalStubs.add("protocol-test-json-stub.ts");
+                break;
+            case "application/xml":
+                writer.addDependency(TypeScriptDependency.XML_PARSER);
+                writer.addImport("parse", "xmlParse", "fast-xml-parser");
+                additionalStubs.add("protocol-test-xml-stub.ts");
+                break;
+            default:
+                throw new CodegenException("Unable to compare bodies of media type `" + mediaType + "`");
```
mtdowling: Should this throw or warn? Seems like adding a new comparison would be a breaking change to generators. I don't think we'd want to coordinate these updates across all implementations.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,484 @@
...
+
+            writer.openBlock("Object.keys(paramsToValidate).forEach(param => {", "});", () -> {
+                writer.write("expect(r[param]).toBeDefined();");
+                writer.write("expect(r[param]).toBe(paramsToValidate[param]);");
+            });
+        }
+    }
+
+    /**
+     * Supports writing out
```
mtdowling: Would be nice to say why this is needed

kstich: D'oh!

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
kstich: It'll involve having to compound the test's id with the operation's name so that Jest has these as unique tests, but definitely doable.

kstich: This is also good because it'll enforce using any command specific middleware.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
mtdowling: Yeah -- trying an exact match + warning as a last resource seems good enough to me. At most, I think it could be configurable as to whether it fails or warns 🤷

PR #172, opened 2020-05-01
Use array for SetShape

trivikr:
*Issue #, if available:*
We decided to use Array for SetShape in internal discussions for similar reasons to why Objects are used instead of Map:
* new Set needs to be created by calling `new Set()`, just like `new Map()`
* `setObj.has(value)` is similar to `objMap.get(key)`

Smithy specification for Set https://awslabs.github.io/smithy/spec/core.html#set

```console
Not all languages support set data structures with non-scalar values. Such languages SHOULD represent sets as a custom set data structure that can interpret value hash codes and equality. Alternatively, clients MAY store the values of a set data structure in a list and rely on the service to ensure uniqueness.
```

*Description of changes:*
Use array for SetShape


trivikr:
CodeGen PR: https://github.com/aws/aws-sdk-js-v3/pull/1144

trivikr:
Removed instanceof Set check in https://github.com/awslabs/smithy-typescript/pull/172/commits/e9894d603d76035f7f8ee5fa09ab29c281a4640b

PR #183, opened 2020-06-03
Add $retryable to Error is RetryableTrait is present

trivikr:
*Issue #, if available:*
Follow-up to https://github.com/aws/aws-sdk-js-v3/pull/1222 to consume retryableTrait from model

*Description of changes:*
* Add $retryable to Error is RetryableTrait is present
* Codegen PR at https://github.com/aws/aws-sdk-js-v3/pull/1237

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
@@ -15,6 +15,7 @@
 
 package software.amazon.smithy.typescript.codegen;
 
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import software.amazon.smithy.codegen.core.Symbol;
@@ -23,6 +24,7 @@ import software.amazon.smithy.codegen.core.SymbolReference;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
+import software.amazon.smithy.model.traits.RetryableTrait;
 
 /**
  * Generates normal structures and error structures.
@@ -53,21 +55,26 @@ final class StructureGenerator implements Runnable {
     /**
      * Renders a normal, non-error structure.
      *
-     * <p>For example, given the following Smithy model:
+     * <p>
+     * For example, given the following Smithy model:
      *
-     * <pre>{@code
+     * <pre>
+     * {@code
      * namespace smithy.example
      *
      * structure Person {
-     *     @required
+     *     &#64;required
      *     name: String,
      *     age: Integer,
      * }
-     * }</pre>
+     * }
+     * </pre>
      *
-     * <p>The following TypeScript is rendered:
+     * <p>
+     * The following TypeScript is rendered:
      *
-     * <pre>{@code
+     * <pre>
+     * {@code
      * import { isa as __isa } from "@aws-sdk/smithy-client";
      *
      * export interface Person {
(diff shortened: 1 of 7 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
```diff
@@ -148,18 +161,27 @@ private void renderErrorStructure() {
...
+                        .map(SymbolReference::getAlias))
+                .collect(Collectors.joining(", "));
 
         writer.openBlock("export interface $L extends $L {", symbol.getName(), extendsFrom);
         writer.write("name: $S;", shape.getId().getName());
         writer.write("$$fault: $S;", errorTrait.getValue());
-        StructuredMemberWriter structuredMemberWriter = new StructuredMemberWriter(
-                model, symbolProvider, shape.getAllMembers().values());
+
+        Optional<RetryableTrait> retryableTrait = shape.getTrait(RetryableTrait.class);
```
kstich, line 174: Shouldn't this just be:

```
HttpProtocolGeneratorUtils.writeRetryableTrait(writer, shape);
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
```diff
@@ -53,21 +55,26 @@ public void run() {
     /**
      * Renders a normal, non-error structure.
      *
-     * <p>For example, given the following Smithy model:
+     * <p>
```
kstich, line 58: There're a ton of formatting changes in this file that are at either PR noise or something we don't want changed. Can they be reverted?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
trivikr: Nope, because `StructureGenerator` requires `;` after the member is written while `Http<X>ProtocolGenerator` requires `,`

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
trivikr: Was added by mistake as smithy-typescript doesn't have auto formatting available on save like aws-sdk-js-v3 does. Will remove in next commit

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
trivikr: Formatting removed in [commit](https://github.com/awslabs/smithy-typescript/pull/183/commits/c948214e41ce11721931182283ad491c8cb5adb0)

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
trivikr: Formatting removed in [commit](https://github.com/awslabs/smithy-typescript/pull/183/commits/c948214e41ce11721931182283ad491c8cb5adb0)

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
kstich: Could that just be a `String` parameter to the method? I'd be worried about having separate implementations of how this is added becoming out of sync.

trivikr: Yup, the delimiter parameter can be just a String parameter to the method.
It has not been done in the past for other common params like `name`, `$$fault` though.

Will pass delimiter param for retryableTrait

trivikr: Added in [commit](https://github.com/awslabs/smithy-typescript/pull/183/commits/1349c612cc20d6776612537d5803d1402615b003)

PR #248, opened 2020-12-14
fix: use typeof inputTokenName as token type instead of string

trivikr:
*Issue #, if available:*
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1782

*Description of changes:*
use typeof inputTokenName as token type instead of string


alexforsyth (approved):
LGTM

PR #268, opened 2021-01-29
Update protocol test request body comparison

kstich:
This commit updates the request body comparison generation for
protocol tests to safely handle having multiple `bodyMediaType`
settings in the same suite of protocol tests. It also updates the
methodology for comparing blob bodies to not be a special case in
the generator, as they can have alternate media types.


JordonPhillips (approved):

alexforsyth:
@kstich what are the generated changes?

kstich:
See aws/aws-sdk-js-v3#2049

alexforsyth (approved):
LGTM

PR #270, opened 2021-02-11
Add basic support for server serde in http binding protocols

JordonPhillips:
*Description of changes:*

This adds basic support for server serde in http binding protocols, which is to say it adds support for serializing responses and deserializing requests. Currently there are a few major points missing:

- ~~No deserialization support for members bound to the path~~
- ~~No deserialization support for members bound to the query string~~
- ~~No deserialization support for members bound to the endpoint~~
- No protocol tests (this will need an actual interface being generated first)

@adamthom-amzn
@gosar

This adds two new options to `TypeScriptSettings` which is what the `smithy-build.json` config is mapped to:

* `generateClient` - boolean - default true - triggers client generation
* `generateServerSdk` - boolean - default false - triggers ssdk generation

So to do server only you might have something like:

```json
{
    "version" : "1.0",
    "outputDirectory" : "build/output",
    "projections" : {
        "ts-server" : {
            "plugins": {
                "typescript-codegen" : {
                    "generateClient": false,
                    "generateServerSdk": true,
                    "package" : "@bootleg-service/server-bootleg",
                    "packageVersion": "1.0.0-rc.1"
                }
            }
        }
    }
}
```

This also adds in deser support for the various uri bindings. Here's an example generated deser function:

```typescript
export const deserializeAws_restJson1GetVenueCommandRequest = async(
  output: __HttpRequest,
  context: __SerdeContext
): Promise<GetVenueCommandInput> => {
  const contents: GetVenueCommandInput = {
    city: undefined,
    hostParam: undefined,
    inputHeader: undefined,
    inputSpecific: undefined,
    name: undefined,
    queryParam: undefined,
  };
  const query = output.endpoint.query
  if (query !== undefined && query !== null) {
    if (query['queryP'] !== undefined) {
      contents.queryParam = query['queryP'];
    }
  }
  const pathRegex = new RegExp("/venues/(?<city>.*)/(?<name>.*)/");
  const parsedPath: RegExpMatchArray = output.endpoint.path.match(pathRegex);
  if (parsedPath.groups !== undefined) {
    contents.name = parsedPath.groups.name;
    contents.city = parsedPath.groups.city;
  }
  const hostRegex = new RegExp("^(?<hostParam>.*)\\.bar\\.baz-bam\\.");
  const parsedHost: RegExpMatchArray = output.endpoint.path.match(pathRegex);
  if (parsedHost.groups !== undefined) {
    contents.hostParam = parsedHost.groups.hostParam;
  }
  if (output.headers["x-amz-input-header"] !== undefined) {
    contents.inputHeader = output.headers['x-amz-input-header'];
  }
  const data: any = (await parseBody(output.body, context));
  if (data.hostParam !== undefined && data.hostParam !== null) {
    contents.hostParam = data.hostParam;
  }
  if (data.inputSpecific !== undefined && data.inputSpecific !== null) {
    contents.inputSpecific = data.inputSpecific;
  }
  return Promise.resolve(contents);
}
```

Also note that this pr is merging into the ssdk branch and not main


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -82,6 +86,9 @@ public static TypeScriptSettings from(Model model, ObjectNode config) {
         settings.packageJson = config.getObjectMember(PACKAGE_JSON).orElse(Node.objectNode());
         config.getStringMember(PROTOCOL).map(StringNode::getValue).map(ShapeId::from).ifPresent(settings::setProtocol);
         settings.setPrivate(config.getBooleanMember(PRIVATE).map(BooleanNode::getValue).orElse(false));
+        settings.setGenerateClient(config.getBooleanMember(GENERATE_CLIENT).map(BooleanNode::getValue).orElse(true));
+        settings.setGenerateServerSdk(
+                config.getBooleanMember(GENERATE_SERVER_SDK).map(BooleanNode::getValue).orElse(true));
```
adamthom-amzn, line 91: I thought from the description you wanted this to default to false

JordonPhillips: whoops

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -856,7 +1098,155 @@ private void writeEventBody(GenerationContext context, StructureShape event) {
...
+        List<HttpBinding> pathBindings = bindingIndex.getRequestBindings(operation, Location.LABEL);
+        if (pathBindings.isEmpty()) {
+            return;
+        }
+        StringBuilder pathRegexBuilder = new StringBuilder();
+        for (Segment segment : trait.getUri().getSegments()) {
+            pathRegexBuilder.append("/");
+            if (segment.isLabel()) {
+                // Create a named capture group for the segment so we can grab it later without regard to order.
+                pathRegexBuilder.append(String.format("(?<%s>.*)", segment.getContent()));
```
adamthom-amzn, line 1196: In ... other implementations ... this uses `[^/]+` for non-greedy segments and `.+` for greedy ones. There may be cases with path patterns where this regex causes non-greedy segments to span `/`s:

`/(?<label>.*)/(?<greedy>.*)` would probably not match `/foo/bar/baz` in the way you want.

adamthom-amzn (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
JordonPhillips: Nice catch, thanks!

adamthom-amzn (approved):

PR #277, opened 2021-02-26
Check for only structure shapes for circular dependency

trivikr:
*Issue #, if available:*
N/A

*Description of changes:*
The Set `parents` was added to avoid StackOverflow Error occurring due to circular dependencies in recursion in PR https://github.com/awslabs/smithy-typescript/pull/181
To check for recursions, the value needs to be populated only for Structure shapes.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
@@ -197,24 +197,22 @@ final class StructuredMemberWriter {
             // always call filterSensitiveLog for UnionShape
             return true;
         } else if (memberTarget.isStructureShape()) {
-            parents.add(symbolProvider.toMemberName(member));
-            Collection<MemberShape> structureMemberList = ((StructureShape) memberTarget).getAllMembers().values();
-            for (MemberShape structureMember: structureMemberList) {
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
```diff
@@ -193,11 +193,11 @@ private boolean isMemberOverwriteRequired(MemberShape member, Set<String> parent
         }
 
         Shape memberTarget = model.expectShape(member.getTarget());
-        parents.add(symbolProvider.toMemberName(member));
         if (memberTarget.isUnionShape()) {
             // always call filterSensitiveLog for UnionShape
```
kstich, line 197: Unions can also safely be recursive.

kstich (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
trivikr: We return true for `isMemberOverwriteRequired` for Unions, so recursive check never happens when we encounter Unions.

alexforsyth:
LGTM - i'll leave approval to @kstich

AllanZhengYP (approved):

kstich (approved):

PR #279, opened 2021-03-02
Generate ssdk request protocol tests

JordonPhillips:
*Description of changes:*

This adds test generation for ssdk request deserialization. For example:

```typescript
/**
 * Mixes constant and variable query string parameters
 */
it("RestJsonConstantAndVariableQueryStringMissingOneValue:ServerRequest", async () => {
  let testFunction = jest.fn();
  testFunction.mockReturnValue({});
  const testService: Partial<RestJsonService> = {
    ConstantAndVariableQueryString: testFunction as __Operation<
      ConstantAndVariableQueryStringCommandInput,
      ConstantAndVariableQueryStringCommandOutput
    >,
  };
  const handler = getRestJsonServiceHandler(testService as RestJsonService);
  const request = new HttpRequest({
    method: "GET",
    hostname: "foo.example.com",
    path: "/ConstantAndVariableQueryString",
    query: {
      foo: "bar",
      baz: "bam",
    },
    headers: {},
    body: Readable.from([""]),
  });
  await handler.handle(request);

  expect(testFunction.mock.calls.length).toBe(1);
  let r: any = testFunction.mock.calls[0][0];

  const paramsToValidate: any = [
    {
      baz: "bam",
    },
  ][0];
  Object.keys(paramsToValidate).forEach((param) => {
    expect(r[param]).toBeDefined();
    expect(equivalentContents(r[param], paramsToValidate[param])).toBe(true);
  });
});
```

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
@@ -17,6 +17,9 @@ package software.amazon.smithy.typescript.codegen;
 
 import static java.lang.String.format;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -24,6 +27,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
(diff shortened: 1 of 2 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -217,6 +246,99 @@ private void generateRequestTest(OperationShape operation, HttpRequestTestCase t
...
+        writer.openBlock("it($S, async () => {", "});\n", testName, () -> {
+            // TODO: use the symbol provider when it's ready
+            String serviceName = StringUtils.capitalize(service.getId().getName());
+            String operationName = StringUtils.capitalize(operation.getId().getName());
+            Symbol inputType = operationSymbol.expectProperty("inputType", Symbol.class);
+
+            // Declare "r" to capture the deserialized input
+            writer.write("let r: any;");
+
+            // This doesn't actually satisfy the service's server interface since it only applies to one operation
```
adamthom-amzn, line 271: Do we not have access to a mocking facility that will satisfy the type checker and remove the need to do this manual state capture for verification?

JordonPhillips: I updated to use a mock function there. I can't really get rid of casting, but I did rework it so that we're making better use of the type checker. Now the test service is defined as a `Partial` variable and set to an object literal. So we get excess property checking and checking of the types on the singular function we actually define. I have to as cast it later to the non-partial service, but that seems fine. jest doesn't seem to be able to stub a typescript interface which is why I went that route.

adamthom-amzn (approved):

PR #284, opened 2021-03-12
Use server symbol provider in protocol test gen

JordonPhillips:
This updates the server protocol test gen to use the server symbol provider. It also splits up the server / client "command" generators because having them all in one was causing client code to be generated in the server files. Lastly, I changed the folder for the server operation types to "operation" since typescript special cases "types" and that was causing generation issues.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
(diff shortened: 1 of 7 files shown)


JordonPhillips:
I discovered some issues with the symbol provider change that make this impossible to test, so closing this for now until I get that resolved.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.OperationIndex;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+
+/**
+ * Generates server operation types.
+ */
+final class ServerCommandGenerator implements Runnable {
```
adamthom-amzn, line 34: I don't like the association of the term command with this at all (but no, I don't have a better idea off the top of my head)

adamthom-amzn (approved):

PR #289, opened 2021-03-19
ssdk modeled exception handling

JordonPhillips:
This add exception handling to the SSDK. In doing so I've re-arranged a bunch of things to be a bit more split up. Notably the `getServiceHandler` function has been moved into the service's file and each `OperationSerializer` has been moved into that operation's file.

In addition to re-organizing, this depends on two new additions to the `OperationSerializer`, `isOperationError` and `serializeError`.

```typescript
export interface OperationSerializer<T, K extends keyof T, E extends SmithyException> {
    serialize(input: OperationOutput<T[K]>, ctx: Omit<SerdeContext, 'endpoint'>): Promise<HttpResponse>
    deserialize(input: HttpRequest, ctx: SerdeContext): Promise<OperationInput<T[K]>>
    isOperationError(error: any): error is E
    serializeError(error: E, ctx: Omit<SerdeContext, 'endpoint'>): Promise<HttpResponse>
}
```

`isOperationError` is used to check to see if an exception is in a union of exceptions that the operation expects. TypeScript is neat in that it allows you to specifically type a boolean so that it can know that's what I'm doing. Unfortunately TypeScript won't generate this check for me, so I have to do it myself. This is needed because js has only one catch block and the exception is always an `any` (though I'm assigning it to `unknown` for a bit of extra saftey).

`serializeError` is pretty self-explanitory: it dispatches to the correct serializer based on the `name` of the `SmithyException`. Since it's a simple union, something like an error mux isn't needed.

Here's an example of a generated `OperationSerializer`:

```typescript
export type GetVenue = __Operation<GetVenueServerInput, GetVenueServerOutput>

export type GetVenueServerInput = GetVenueInput;
export type GetVenueServerOutput = GetVenueOutput & __MetadataBearer;

export type GetVenueErrors = NoSuchResource | RequestTimeout

export class GetVenueSerializer implements OperationSerializer<BootlegService, "GetVenue", GetVenueErrors> {
  serialize = serializeGetVenueResponse;
  deserialize = deserializeGetVenueRequest;

  isOperationError(error: any): error is GetVenueErrors {
    // I wish TS could do this for me, but it does not seem to be able to.
    const names: GetVenueErrors['name'][] = ["NoSuchResource", "RequestTimeout"];
    return names.includes(error.name);
  };

  serializeError(error: GetVenueErrors, ctx: Omit<SerdeContext, 'endpoint'>): Promise<__HttpResponse> {
    switch (error.name) {
      case "NoSuchResource": {
        return serializeNoSuchResourceError(error, ctx);
      }
      case "RequestTimeout": {
        return serializeRequestTimeoutError(error, ctx);
      }
      default: {
        // Re-throw any error that we don't know about
        throw error;
      }
    }
  }

}
```

The generated service handler now looks like:

```typescript
export type BootlegServiceOperations = "GetVenue" | "ListVenues";
export interface BootlegService {
  GetVenue: GetVenue
  ListVenues: ListVenues
}
export class BootlegServiceHandler implements ServiceHandler {
  private service: BootlegService;
  private mux: Mux<"Bootleg", BootlegServiceOperations>;
  private serializerFactory: <T extends BootlegServiceOperations>(operation: T) => OperationSerializer<BootlegService, T, SmithyException>;
  private serdeContextBase = {
    base64Encoder: toBase64,
    base64Decoder: fromBase64,
    utf8Encoder: toUtf8,
    utf8Decoder: fromUtf8,
    streamCollector: streamCollector,
    requestHandler: new NodeHttpHandler(),
    disableHostPrefix: true
  };
  /**
   * Construct a BootlegService handler.
   * @param service The {@link BootlegService} implementation that supplies the business logic for BootlegService
   * @param mux The {@link Mux} that determines which service and operation are being invoked by a given {@link HttpRequest}
   * @param serializerFactory A factory for an {@link OperationSerializer} for each operation in BootlegService that
   *                          handles deserialization of requests and serialization of responses
   */
  constructor(service: BootlegService, mux: Mux<"Bootleg", BootlegServiceOperations>, serializerFactory: <T extends BootlegServiceOperations>(op: T) => OperationSerializer<BootlegService, T, SmithyException>) {
    this.service = service;
    this.mux = mux;
    this.serializerFactory = serializerFactory;
  }
  async handle(request: HttpRequest): Promise<HttpResponse> {
    const target = this.mux.match(request);
    if (target === undefined) {
      throw new Error(`Could not match any operation to ${request.method} ${request.path} ${JSON.stringify(request.query)}`);
    }
    switch (target.operation) {
      case "GetVenue" : {
        let serializer = this.serializerFactory("GetVenue");
        try {
          let input = await serializer.deserialize(request, {
            endpoint: () => Promise.resolve(request), ...this.serdeContextBase
          });
          let output = this.service.GetVenue(input, request);
          return serializer.serialize(output, this.serdeContextBase);
        } catch(error: unknown) {
          if (serializer.isOperationError(error)) {
            return serializer.serializeError(error, this.serdeContextBase);
          }
          throw error;
        }
      }
      case "ListVenues" : {
        let serializer = this.serializerFactory("ListVenues");
        try {
          let input = await serializer.deserialize(request, {
            endpoint: () => Promise.resolve(request), ...this.serdeContextBase
          });
          let output = this.service.ListVenues(input, request);
          return serializer.serialize(output, this.serdeContextBase);
        } catch(error: unknown) {
          if (serializer.isOperationError(error)) {
            return serializer.serializeError(error, this.serdeContextBase);
          }
          throw error;
        }
      }
    }
  }
}
```

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
@@ -18,6 +18,7 @@ package software.amazon.smithy.typescript.codegen;
 import static software.amazon.smithy.typescript.codegen.CodegenUtils.getBlobStreamingMembers;
 import static software.amazon.smithy.typescript.codegen.CodegenUtils.writeStreamingMemberType;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import software.amazon.smithy.codegen.core.Symbol;
@@ -26,7 +27,9 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator;
 
 /**
  * Generates server operation types.
@@ -41,19 +44,25 @@ final class ServerCommandGenerator implements Runnable {
     private final OperationIndex operationIndex;
     private final Symbol inputType;
     private final Symbol outputType;
+    private final ProtocolGenerator protocolGenerator;
+    private final ApplicationProtocol applicationProtocol;
 
     ServerCommandGenerator(
             TypeScriptSettings settings,
             Model model,
             OperationShape operation,
             SymbolProvider symbolProvider,
-            TypeScriptWriter writer
+            TypeScriptWriter writer,
+            ProtocolGenerator protocolGenerator,
+            ApplicationProtocol applicationProtocol
     ) {
         this.settings = settings;
         this.model = model;
         this.operation = operation;
         this.symbolProvider = symbolProvider;
         this.writer = writer;
+        this.protocolGenerator = protocolGenerator;
+        this.applicationProtocol = applicationProtocol;
 
         Symbol operationSymbol = symbolProvider.toSymbol(operation);
         operationIndex = OperationIndex.of(model);
@@ -64,6 +73,9 @@ final class ServerCommandGenerator implements Runnable {
     @Override
     public void run() {
         addInputAndOutputTypes();
+        writeErrorType();
+        writeErrorChecker();
+        writeErrorHandler();
     }
 
     private void addInputAndOutputTypes() {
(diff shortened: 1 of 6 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
```diff
@@ -99,4 +113,95 @@ private void writeOutputType(String typeName, Optional<StructureShape> outputSha
...
         }
     }
+
+    private void writeErrorType() {
+        Symbol operationSymbol = symbolProvider.toSymbol(operation);
+
+        if (operation.getErrors().isEmpty()) {
+            writer.write("export type $L = never;", errorsType.getName());
+        } else {
+            writer.writeInline("export type $LErrors = ", operationSymbol.getName());
```
adamthom-amzn, line 123: Shouldn't this use `errorsType`?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
```diff
@@ -99,4 +113,95 @@ private void writeOutputType(String typeName, Optional<StructureShape> outputSha
...
+
+    private void writeErrorHandler() {
+        writer.addImport("SerdeContext", null, "@aws-sdk/types");
+        writer.openBlock("serializeError(error: $T, ctx: Omit<SerdeContext, 'endpoint'>): Promise<$T> {", "}",
+                errorsType, applicationProtocol.getResponseType(), () -> {
+            writer.openBlock("switch (error.name) {", "}", () -> {
+                for (ShapeId errorId : operation.getErrors()) {
+                    writeErrorHandlerCase(errorId);
+                }
+                writer.openBlock("default: {", "}", () -> writer.write("throw error;"));
```
adamthom-amzn, line 192: In the end we will probably not want the framework to throw; instead it will render any error as a serialized InternalFailure (synthetic exception). So either this should default to serializing any error as an InternalFailure, or it should throw and we should have a higher-level catch all and a protocol-level utility function that serializes it.

adamthom-amzn (commented):
We will also need to, instead of throwing an Error when we can't determine the operation being invoked, render a 404 `UnknownOperationException`.

Fine by me if you want to leave generation of our "synthetic" exceptions for another review.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
JordonPhillips: Yeah we will need some synthetic error handling. I omitted that for now to keep the pr focused on modeled errors. In the short term we can rely on APIG to catch any unhandled errors and return a 500.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
JordonPhillips: Yes, yes it should

adamthom-amzn (approved):

PR #311, opened 2021-04-13
Consume httpQueryParams trait

trivikr:
*Issue #, if available:*
N/A

*Description of changes:*
PR to consume smithy [httpQueryParams](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpqueryparams-trait) trait

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -628,7 +628,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
 
         String valueString = getInputValue(context, bindingType, "value", mapMember,
                 model.expectShape(mapMember.getTarget()));
-        return "(" + dataSource + " || {}).reduce("
+        return "Object.entries(" + dataSource + " || {}).reduce("
             + "(acc: any, [key, value]: [string, " + symbolProvider.toSymbol(mapMember) + "]) => ({"
             +   "...acc,"
             +   "[key]: " + valueString + ","
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -301,14 +301,29 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
         TypeScriptWriter writer = context.getWriter();
         SymbolProvider symbolProvider = context.getSymbolProvider();
         List<HttpBinding> queryBindings = bindingIndex.getRequestBindings(operation, Location.QUERY);
+        List<HttpBinding> queryParamsBindings = bindingIndex.getRequestBindings(operation, Location.QUERY_PARAMS);
 
         // Build the initial query bag.
         Map<String, String> queryLiterals = trait.getUri().getQueryLiterals();
-        if (!queryLiterals.isEmpty() || !queryBindings.isEmpty()) {
+        if (!queryLiterals.isEmpty() || !queryBindings.isEmpty() || !queryParamsBindings.isEmpty()) {
             writer.openBlock("const query: any = {", "};", () -> {
                 if (!queryLiterals.isEmpty()) {
                     // Write any query literals present in the uri.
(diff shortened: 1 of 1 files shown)


trivikr:
PR for generated code https://github.com/aws/aws-sdk-js-v3/pull/2246

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -587,6 +608,36 @@ private String getNamedMembersInputParam(
...
+        MemberShape mapMember = target.getValue();
+        SymbolProvider symbolProvider = context.getSymbolProvider();
+
+        String valueString = getInputValue(context, bindingType, "value", mapMember,
+                model.expectShape(mapMember.getTarget()));
+        return "Object.entries(" + dataSource + " || {}).reduce("
+            + "(acc: any, [key, value]: [string, " + symbolProvider.toSymbol(mapMember) + "]) => ({"
+            +   "...acc,"
+            +   "[key]: " + valueString + ","
+            + "}), {})";
```
JordonPhillips, line 638: I get why you did it this way - this should work for any sort of map you throw at it. At the same time, the query params trait can only ever be applied to `map<string, string>` or `map<string, list<string>>` which just so happens to be [the precise type signature of your query map](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/types/src/http.ts#L60). It seems like it would be much simpler to just directly expand it into the query map if it's defined. That would turn this:

```typescript
  const query: any = {
    ...(input.queryParamsMapOfStrings !== undefined &&
      Object.entries(input.queryParamsMapOfStrings || {}).reduce(
        (acc: any, [key, value]: [string, string]) => ({ ...acc, [key]: value }),
        {}
      ))
  }
```

into this:

```typescript
  const query: any = {
    ...(input.queryParamsMapOfStrings !== undefined ? input.queryParamsMapOfStrings : {})
  }
```

or

```typescript
const query: any = {};
if (input.queryParamsMapOfStrings !== undefined) {
  query = { ...input.queryParamsMapOfStrings, ...query }
}
```

It's worth noting that the trait is structurally exclusive, meaning there will only ever be one member bearing it on a given operation.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -299,36 +299,54 @@ private boolean writeRequestQueryString(
...
         // Build the initial query bag.
         Map<String, String> queryLiterals = trait.getUri().getQueryLiterals();
-        if (!queryLiterals.isEmpty() || !queryBindings.isEmpty()) {
+        if (!queryLiterals.isEmpty() || !queryBindings.isEmpty() || !queryParamsBindings.isEmpty()) {
             writer.openBlock("const query: any = {", "};", () -> {
                 if (!queryLiterals.isEmpty()) {
                     // Write any query literals present in the uri.
                     queryLiterals.forEach((k, v) -> writer.write("$S: $S,", k, v));
                 }
+                // Handle any additional query params bindings.
```
JordonPhillips, line 313: nit: probably worth noting that these have to come first because explicit bindings override them and js will take whatever comes last

JordonPhillips (approved):
Approved because this looks like it'll work, though I'd prefer to see the change mentioned.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
trivikr: Fixed in [`45f371c` (#311)](https://github.com/awslabs/smithy-typescript/pull/311/commits/45f371c48552e524f7e5b935ed9e3d811dc5153f)

Codegen the following output code:
```js
  const query: any = {
    ...(input.queryParamsMapOfStrings !== undefined && input.queryParamsMapOfStrings)
  }
```

Verified the new api-gateway client in https://github.com/aws/aws-sdk-js-v3/pull/2249

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
trivikr: Added in [`95f2120` (#311)](https://github.com/awslabs/smithy-typescript/pull/311/commits/95f21204f80c6795f9e21391b31fc102fd4cd972)

AllanZhengYP (approved):

PR #314, opened 2021-04-19
Split server codegen out into a separate plugin

JordonPhillips:
This splits up the server code generation to be gated behind a separate plugin. This lets us generate separate packages for each while still being able to generate both a client and ssdk in a single projection.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
@@ -25,7 +25,6 @@ import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.function.BiFunction;
 import java.util.logging.Logger;
 import software.amazon.smithy.build.FileManifest;
 import software.amazon.smithy.build.PluginContext;
@@ -46,6 +45,7 @@ import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.PaginatedTrait;
+import software.amazon.smithy.typescript.codegen.TypeScriptSettings.ArtifactType;
 import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator;
 import software.amazon.smithy.typescript.codegen.integration.RuntimeClientPlugin;
 import software.amazon.smithy.typescript.codegen.integration.TypeScriptIntegration;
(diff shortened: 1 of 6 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -70,15 +71,18 @@
...
     private final List<TypeScriptIntegration> integrations = new ArrayList<>();
     private final List<RuntimeClientPlugin> runtimePlugins = new ArrayList<>();
     private final ProtocolGenerator protocolGenerator;
     private final ApplicationProtocol applicationProtocol;
 
-    CodegenVisitor(PluginContext context) {
+    CodegenVisitor(
+            PluginContext context,
+            BiFunction<Model, TypeScriptSettings, SymbolProvider> createSymbolProvider,
+            TypeScriptSettings.ArtifactType artifactType
```
adamthom-amzn, line 84: This BiFunction is always static, just make createSymbolProvider a method on the ArtifactType enum and you don't have this ugliness.

adamthom-amzn (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
JordonPhillips: This what you're thinking?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -313,4 +314,29 @@ public ShapeId resolveServiceProtocol(Model model, ServiceShape service, Set<Sha
...
+         * Creates a TypeScript symbol provider suited to the artifact type.
+         *
+         * @param model Model to generate symbols for.
+         * @param settings Settings used by the symbol provider.
+         * @return Returns the created provider.
+         */
+        public SymbolProvider createSymbolProvider(Model model, TypeScriptSettings settings) {
+            return symbolProviderFactory.apply(model, settings);
+        }
+    }
```
adamthom-amzn, line 341: This works, but the canonical, non-everything-has-to-be-a-function solution looks like this
```suggestion
public enum ArtifactType {
    CLIENT {
        @Override
        public SymbolProvider createSymbolProvider(Model model, TypeScriptSettings settings) {
            return new SymbolVisitor(model, settings);
        }
    },
    SSDK {
        @Override
        public SymbolProvider createSymbolProvider(Model model, TypeScriptSettings settings) {
            return new ServerSymbolVisitor(model, new SymbolVisitor(model, settings));
        }
    };

    /**
     * Creates a TypeScript symbol provider suited to the artifact type.
     *
     * @param model Model to generate symbols for.
     * @param settings Settings used by the symbol provider.
     * @return Returns the created provider.
     */
    public abstract SymbolProvider createSymbolProvider(Model model, TypeScriptSettings settings);
}
```

adamthom-amzn (approved):

PR #320, opened 2021-04-28
Generation of low-level validation methods for types

adamthom-amzn:
*Description of changes:*

Validation makes the type's namespace look like this:

```typescript
export namespace CreateVenueInput {
  export const filterSensitiveLog = (obj: CreateVenueInput): any => ({
    ...obj,
  })
  const memberValidators = {
    clientToken: new __NullValidator(),
    name: new __CompositeValidator<string>([
      new __RequiredValidator(),
      new __LengthValidator(1, 128),
      new __PatternValidator("^[\\p{L}\\p{N}]+$"),
    ]),
    city: new __CompositeValidator<string>([
      new __RequiredValidator(),
      new __LengthValidator(1, 128),
    ]),
    capacity: new __CompositeValidator<number>([
      new __RequiredValidator(),
      new __RangeValidator(1, undefined),
    ]),
  };
  export const validate = (obj: CreateVenueInput): __ValidationFailure[] => {
    return [
      ...memberValidators.clientToken.validate(obj.clientToken, "clientToken"),
      ...memberValidators.name.validate(obj.name, "name"),
      ...memberValidators.city.validate(obj.city, "city"),
      ...memberValidators.capacity.validate(obj.capacity, "capacity"),
    ];
  }
}
```


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
```diff
@@ -264,4 +273,83 @@ private String getSanitizedMemberName(MemberShape member) {
...
+            for (MemberShape member : members) {
+                writer.write("...memberValidators.$1L.validate($2L.$1L, $3S),",
+                        getSanitizedMemberName(member), param, member.getMemberName());
+            }
+        });
+    }
+
+    private void writeValidator(TypeScriptWriter writer, Trait trait) {
+        if (trait instanceof RequiredTrait) {
+            writer.addImport("RequiredValidator", "__RequiredValidator", "@aws-smithy/server-common");
```
gosar, line 315: `@aws-smithy/server-common` is repeated in many places. I'm not sure, but wonder if these should be specified in AwsDependency/TypeScriptDependency. I've seen others defined there.

gosar (approved):

smithy-typescript-ssdk-libs/server-common/src/validation/validators.ts
```diff
@@ -13,16 +13,57 @@
...
+            let failure = v.validate(input, memberName);
+            if (failure) {
+                retVal.push(failure);
+            }
+        }
+        return retVal;
+    }
+}
+
+export class NullValidator {
```
JordonPhillips, line 45: `NoOpValidator`? I say this because validating nullability is feasibly something this could be doing

JordonPhillips (changes requested):
This appears to only validate structure members and not those of maps, sets, or lists. Additionally, we need to validate types since js certainly won't (include uniqueness in sets). Basic type validation can probably be in another pr though.

smithy-typescript-ssdk-libs/server-common/src/validation/validators.ts
adamthom-amzn: Isn't that what RequiredValidator does? This is the null object pattern, hence the name, but I have no qualms renaming it to NoOpValidator.

PR #328, opened 2021-05-05
Add typescript top-level tooling and ci

JordonPhillips:
This adds some top level tooling to the typescript package so that everything can be built / linted / tested all in one command. It also adds some github to ci to run a build, run the linter, and ensure the code formatter has been run. Lastly, I fixed some eslint errors.


smithy-typescript-ssdk-libs/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/awslabs/smithy-typescript/tree/main/smithy-typescript-ssdk-libs"
+  },
+  "author": "AWS Smithy Team",
+  "license": "UNLICENSED",
+  "dependencies": {
+    "glob": "^7.1.6",
+    "rimraf": "^3.0.2"
```
adamthom-amzn, line 25: I'm surprised anything is a non-dev-dependency at this level

adamthom-amzn (approved):

smithy-typescript-ssdk-libs/package.json
JordonPhillips: That's a good point - it almost certainly is a dev dependency. I based this off of the aws client repo where it's a non-dev dependency because of some scripts they run. I don't think it really makes a difference either way though since we're not publishing this

PR #338, opened 2021-05-12
fix: Fix path for typesVersions in package.json

rajivshah3:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/2372

*Description of changes:*
Fixes the path of the types for Typescript 3.x. It looks like they should actually be in `dist/types/ts3.4`, not `types/ts3.4`.


trivikr:
The issue was with internal release script.
Updates about the fix: https://github.com/aws/aws-sdk-js-v3/issues/2474#issuecomment-871798569

rajivshah3:
Thanks!

PR #343, opened 2021-05-25
Add yarn to base package.json dev deps

JordonPhillips:
This adds yarn to the base dev dependencies so that the generated packages can run properly without having to pre-install yarn.

### DIFF
--- a/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/base-package.json
+++ b/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/base-package.json
@@ -33,7 +33,8 @@
     "rimraf": "^3.0.0",
     "ts-jest": "^26.4.1",
     "typedoc": "^0.19.2",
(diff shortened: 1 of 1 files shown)


adamthom-amzn (approved):

JordonPhillips:
> Yarn is not required. Without yarn, users can run the script with npm, which comes with Node.js.

Several of the scripts specifically call out yarn though

trivikr:
> Yarn is not required. Without yarn, users can run the script with `npm`, which comes with Node.js.

I agree with @AllanZhengYP that yarn is not required for non-AWS clients.
Using yarn as a package manager was a decision by AWS SDK for JavaScript team in mid-2019 - primarily because of its support in workspaces. We do not want to enforce this decision on non-AWS clients.

The AWS SDK for JavaScript team might switch to yarn 2 (berry), npm v7 or even pnpm in future.

JordonPhillips:
Since a PR is probably not the best place for this discussion, I opened an issue for it

PR #348, opened 2021-05-28
Input validation in SSDKs

adamthom-amzn:
This adds input validation to the SSDKs, meaning that constraints specified in the model will be applied to inputs after deserialization.

Unlike some other frameworks of note, validation is opt-out, and not opt-in. The service developer, in order to ignore invalid input, will need to supply code that does so explicitly.

To use default validation, service authors must add `smithy.framework#ValidationException` to every operation. It will return these exceptions automatically when inputs fall outside of bounds.  An example output:

```json
{
    "fieldList": [
        {
            "message": "Value Canopy Club@ at '/name' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[\\p{L}\\p{N}][\\p{L}\\p{N} ]+[\\p{L}\\p{N}]$",
            "path": "/name"
        },
        {
            "message": "Value -750 at '/capacity' failed to satisfy constraint: Member must be greater than or equal to 1",
            "path": "/capacity"
        }
    ],
    "message": "2 validation errors at 2 paths detected. First failure: Value Canopy Club@ at '/name' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[\\p{L}\\p{N}][\\p{L}\\p{N} ]+[\\p{L}\\p{N}]$"
}
```

To use custom validation, service authors must set `disableDefaultValidation` to `true` in their `smithy-build.json` and then supply a `ValidationCustomizer` to their handler. Doing so looks like this:

```typescript
const testHandler = getTestHandler(
  async () => {
    return { $metadata: {} };
  },
  (context: ValidationContext<"Test">, failures: ValidationFailure[]): CustomValidationError | undefined => {
    if (!failures) {
      return undefined;
    }
    return new CustomValidationErrorError({
      failingMembers: Array.from(new Set(failures.map((f) => f.memberName))),
      totalFailures: failures.length,
    });
  }
);
```

The second argument is the ValidationCustomizer, which is a function type.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
@@ -98,10 +98,21 @@ final class ServerCommandGenerator implements Runnable {
             } else {
                 writeStreamingMemberType(writer, symbolProvider.toSymbol(input), typeName, blobStreamingMembers.get(0));
             }
+            renderNamespace(typeName, input);
         } else {
             // If the input is non-existent, then use an empty object.
             writer.write("export interface $L {}", typeName);
         }
+    }
+
+    private void renderNamespace(String typeName, StructureShape input) {
+        Symbol symbol = symbolProvider.toSymbol(input);
+        writer.openBlock("export namespace $L {", "}", typeName, () -> {
+            writer.addImport("ValidationFailure", "__ValidationFailure", "@aws-smithy/server-common");
+            writer.writeDocs("@internal");
+            writer.write("export const validate: (obj: $L) => __ValidationFailure[] = $T.validate;",
+                    symbol.getName(), symbol);
+        });
     }
 
     private void writeOutputType(String typeName, Optional<StructureShape> outputShape) {
--- a/smithy-typescript-integ-tests/src/default_validation.spec.ts
+++ b/smithy-typescript-integ-tests/src/default_validation.spec.ts
@@ -1,0 +1,59 @@
+/*
+ *  Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *   http://aws.amazon.com/apache2.0
+ *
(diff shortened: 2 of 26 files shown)


JordonPhillips (approved):
Not crazy about the name ValidationCustomizer but I can't think of anything better

smithy-typescript-integ-tests/src/default_validation.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+ *   http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+import { HttpRequest } from "@aws-sdk/protocol-http";
+import { getTestHandler } from "@aws-smithy/typescript-integ-test-types";
```
gosar, line 17: Where is this defined?

gosar (commented):
LGTM, couple of minor comments

smithy-typescript-integ-tests/src/default_validation.spec.ts
adamthom-amzn: `getTestHandler` is a generated method based on the service name, which is `Test`

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServerCommandGenerator.java
```diff
@@ -98,12 +98,23 @@ private void writeInputType(String typeName, Optional<StructureShape> inputShape
...
         } else {
             // If the input is non-existent, then use an empty object.
             writer.write("export interface $L {}", typeName);
         }
     }
 
+    private void renderNamespace(String typeName, StructureShape input) {
+        Symbol symbol = symbolProvider.toSymbol(input);
+        writer.openBlock("export namespace $L {", "}", typeName, () -> {
+            writer.addImport("ValidationFailure", "__ValidationFailure", "@aws-smithy/server-common");
```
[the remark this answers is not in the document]
adamthom-amzn: I think there's a difference between how TypeScriptDependency is used (I need this in package.json) and import (I need this symbol in this file). I don't see a lot of reuse of TypeScriptDependency for imports but maybe I'm not looking hard enough.

PR #358, opened 2021-06-18
Custom contexts for SSDK handlers

adamthom-amzn:
*Description of changes:*
Service authors will want to pass metadata from the originating event through to their business logic. This will likely come from APIGateway metadata, especially metadata related to authentication and authorization. In particular, the calling user, or metadata attached by the custom Lambda authorizer are frequently referenced in service implementations.

Instead of coming up with some idea of context, or potentially coupling ourselves to the way APIGateway thinks about context, this just lets service authors define their own context type and pass it through the system.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
@@ -310,7 +310,7 @@ final class HttpProtocolTestGenerator implements Runnable {
                     writer.write("body: Readable.from([$S]),", body);
                 }
             });
-            writer.write("await handler.handle(request);").write("");
+            writer.write("await handler.handle(request, {});").write("");
 
             // Assert that the function has been called exactly once.
             writer.write("expect(testFunction.mock.calls.length).toBe(1);");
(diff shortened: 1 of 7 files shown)


JordonPhillips (changes requested):
The protocol test generator probably needs updated right?

adamthom-amzn:
There are additional fixes for the SSDK protocol tests where they were broken by validation previously. You can preview them here: https://github.com/adamthom-amzn/smithy-typescript/compare/server-context...adamthom-amzn:fix-validation

I'm too lazy to rebase them out of the context line and post separately, I'll follow up with another PR. Between the two of them, they get us back to working protocol tests (minus the one known failure).

JordonPhillips (approved):

adamthom-amzn:
I ran `yarn generate-clients` and this is the `git status` output in the SDK (the one commit is to move the smithy plugin up to 0.5.3, which is less flaky):

```
On branch main
Your branch is ahead of 'origin/main' by 1 commit.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   protocol_tests/aws-query/tests/functional/awsquery.spec.ts
```

and the change not staged is the one you'd expect:

```
diff --git a/protocol_tests/aws-query/tests/functional/awsquery.spec.ts b/protocol_tests/aws-query/tests/functional/awsquery.spec.ts
index a26a08b573..01b6921209 100644
--- a/protocol_tests/aws-query/tests/functional/awsquery.spec.ts
+++ b/protocol_tests/aws-query/tests/functional/awsquery.spec.ts
@@ -548,9 +548,7 @@ it("QueryInvalidGreetingError:Error:GreetingWithErrors", async () => {
 /**
  * Parses customized XML errors
  */
-// Manually skipping to unblock smithy-1.8.x update.
-// TODO: Consume AWSQueryError trait as a follow-up.
-it.skip("QueryCustomizedError:Error:GreetingWithErrors", async () => {
+it("QueryCustomizedError:Error:GreetingWithErrors", async () => {
   const client = new QueryProtocolClient({
     ...clientParams,
     requestHandler: new ResponseDeserializationTestHandler(
```

alexforsyth (approved):
+1

PR #360, opened 2021-06-21
Add ability to filter tests

JordonPhillips:
This adds the ability to filter certain test cases. Reasons for doing this might include wanting to skip tests for a feature not yet implemented or wanting to skip a test whose asserted http request is only one of several valid options.

smithy-go has [something similar](https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestGenerator.java#L761), but rather than having a functional interface they have a locked down builder where you must be very explicit on what you want to skip. My motivation for going with the functional interface instead was to address complexity around generating for input/output/error and to allow for more easily skipping based on things in the settings (such as client vs ssdk). I could be convinced to go their way though.

see also: https://github.com/aws/aws-sdk-js-v3/pull/2506

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


alexforsyth:
Sweet, we need to run a codegen along with this too

JordonPhillips:
Temporarily closing because I realized that ts doesn't have a hook into the protocol test generation like go does

adamthom-amzn (approved):

alexforsyth (approved):
+1

PR #365, opened 2021-06-23
Add expectations for default member deser

JordonPhillips:
This adds expectations for the default member deserialization so that an error will be thrown if the type being deserialized isn't correct.

See https://github.com/aws/aws-sdk-js-v3/pull/2515 for the other half of this.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/DocumentMemberDeserVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/DocumentMemberDeserVisitor.java
@@ -124,46 +124,48 @@ public class DocumentMemberDeserVisitor implements ShapeVisitor<String> {
 
     @Override
     public String booleanShape(BooleanShape shape) {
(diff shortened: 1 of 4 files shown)


adamthom-amzn (commented):
in services, won't `TypeError` become `InternalFailure` for service callers? Can we reliably turn that into `SerializationException`?

JordonPhillips:
We'd have to handle that separately from the `InternalFailure` wrapping. We could also have InternalFailure embed the error that caused it so we could unpack it.

adamthom-amzn:
I think a TypeError emerging from deserialization code could be wrapped in a SerializationException without having to go through InternalFailure. The only concern would be whether or not we could also reuse the message in the TypeError for the SerializationException (eventually, I didn't define a message at first, whoops). We would have to be sure that nothing else could throw a TypeError with a message we didn't want customers to receive. That consideration may make us choose a different TypeError, or a specialized subclass of it.

adamthom-amzn (approved):

adamthom-amzn (approved):

trivikr (approved):

PR #366, opened 2021-06-25
Correctly generate test cases for list documents

JordonPhillips:
This fixes an issue that is exposed by the new protocol tests added in Smithy 1.9.0 where array nodes were always assumed to be a collection, when in fact they can also be documents.


adamthom-amzn (approved):

JordonPhillips:
pulled into: https://github.com/awslabs/smithy-typescript/pull/367

PR #370, opened 2021-06-25
Fix deserialization of paths and query params for servers

adamthom-amzn:
*Description of changes:*
This decodes potentially-encoded values as early as possible within the
deserialization process so that no downstream deserializer needs to reckon with
encoding.

It also enforces that we don't accept multiple values for query strings that are modeled as single values.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -1683,10 +1683,25 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
         TypeScriptWriter writer = context.getWriter();
         for (HttpBinding binding : directQueryBindings) {
             String memberName = context.getSymbolProvider().toMemberName(binding.getMember());
-            writer.openBlock("if (query['$L'] !== undefined) {", "}", binding.getLocationName(), () -> {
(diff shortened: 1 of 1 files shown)


gosar (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -1685,8 +1685,23 @@ private void readDirectQueryBindings(GenerationContext context, List<HttpBinding
             String memberName = context.getSymbolProvider().toMemberName(binding.getMember());
             writer.openBlock("if (query['$L'] !== undefined) {", "}", binding.getLocationName(), () -> {
                 Shape target = context.getModel().expectShape(binding.getMember().getTarget());
-                String queryValue = getOutputValue(context, binding.getLocation(),
-                        "query['" + binding.getLocationName() + "'] as string",
+                if (target instanceof CollectionShape) {
+                    writer.write("const decoded = Array.isArray(query['$1L']) ? (query['$1L'] as string[])"
```
JordonPhillips, line 1689: nit: `$S` will put in the quotes for you, as well as handling escaping if necessary.

JordonPhillips (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
adamthom-amzn: yeah I don't know why I translated the previous code so literally

alexforsyth (approved):
+1

PR #377, opened 2021-07-07
Fix query parameter values in SSDK protocol tests

adamthom-amzn:
*Description of changes:*
In other commits, we have adjusted the expectation to be that `HttpRequest` will
contain encoded paths and query parameter keys and values, and the generated
protocol deserializers will perform deserialization. The protocol tests still
expected the `HttpRequest` to hold decoded values, which was causing SSDK test
failures.


JordonPhillips (approved):

adamthom-amzn:
Nevermind. This is wrong. Neither the SDKs nor APIGateway use encoded query string values in the request -- just paths.

PR #386, opened 2021-07-19
Calculate content-length for SSDKs

JordonPhillips:
This updates SSDK generation so that content-length is properly calculated. The existing functionality for clients couldn't be re-used because it is designed to function as a middleware that runs just before the request is sent. That makes sense for the client, because a middleware could be modifying the body. But since the SSDK doesn't have middleware, we just embed it directly.

Related PR that makes associated tests run once more (though they'll fail until this is merged): https://github.com/aws/aws-sdk-js-v3/pull/2599

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -506,6 +506,8 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
                         .forEach(serializingDocumentShapes::add);
             }
 
+            calculateContentLength(context);
+
             writer.openBlock("return new $T({", "});", responseType, () -> {
                 writer.write("headers,");
                 writer.write("body,");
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -519,6 +521,20 @@ private void generateOperationResponseSerializer(
...
 
+    private void calculateContentLength(GenerationContext context) {
+        TypeScriptWriter writer = context.getWriter();
+        writer.addDependency(TypeScriptDependency.AWS_SDK_UTIL_BODY_LENGTH_NODE);
+        writer.addImport("calculateBodyLength", null, "@aws-sdk/util-body-length-node");
+        writer.openBlock("if (body && Object.keys(headers).map((str) => str.toLowerCase())"
+                + ".indexOf('content-length') === -1) {", "}", () -> {
+            writer.write("const length = calculateBodyLength(body);");
+            writer.openBlock("if (length !== undefined) {", "}", () -> {
+                writer.write("headers = { ...headers, 'content-length': String(length) };");
```
adamthom-amzn, line 532: why do we reassign headers instead of mutating it?

JordonPhillips: I based this off of [what the client middleware plugin does](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-content-length/src/index.ts#L29-L32). I'm not terribly attached to this approach though, so I can change it.

adamthom-amzn (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
trivikr: The reassigning was done to usable immutability. We'd discovered a bug in during pre-GA because of mutation, which was fixed by using destructuring assignment

trivikr (approved):

PR #388, opened 2021-07-20
Fix dependency versioning of AWS-vended packages

adamthom-amzn:
*Description of changes:*
smithy-aws-typescript-codegen bundles a properties file that has the version of
every client and library that the JS SDK vends. This essentially ties that set
of versions to the published version of the smithy-aws-typescript-codegen
package. Instead of hardcoding specific versions of AWS clients and libraries,
this change parses that file if smithy-aws-typescript-codegen is also on the
classpath, so that we can use compatible versions of these packages when we
generate package.json. If smithy-aws-typescript-codegen is not on the classpath
we fall back to "latest" for these packages.

I tested this by generating a white-label client locally with this change and the latest version of smithy-aws-typescript-codegen. My package.json file had the correct version of every package (3.22.0, in this case) and the client passed TS type checks. It does not without this change.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
@@ -16,9 +16,19 @@
 package software.amazon.smithy.typescript.codegen;
 
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolDependency;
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
```diff
@@ -157,11 +171,44 @@ public Symbol createSymbol(String name) {
...
+                LOGGER.info("Could not read AWS dependency versions from smithy-aws-typescript-codegen, "
+                        + "will use 'latest' for AWS dependencies");
+                tmpVersions = Collections.emptyMap();
+            }
+            VERSIONS = tmpVersions;
+        }
 
-        private SdkVersion() {}
+        private static String getVersion(String packageName) {
+            return VERSIONS.getOrDefault(packageName, "latest");
```
JordonPhillips, line 211: This is very unfortunate as it essentially means you always need to have a dependency on the aws generator even if you're not using any aws stuff. That's probably still better than the current situation though

JordonPhillips (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
adamthom-amzn: Yeah I can't think of another alternative.

AllanZhengYP (approved):

PR #394, opened 2021-07-30
Handle content-type and accept in ssdk

JordonPhillips:
This updates the SSDK to check the content-type and accept headers.

Both accept and content-type headers may be omitted in any case.

If content-type is set and the operation has a modeled content-type,
the content-type sent is validated against what is modeled. If the
operation doesn't have a modeled content-type, any value is rejected.
If the content-type is rejected, a 415 will be returned.

Accept validation works in a similar way, though it looks at the
modeled output type and a 406 will be returned in the case of a
rejection.

Opening as a draft because I'd like to wait for the malformed request headers prs to land so I can do some of those.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -1683,7 +1683,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
      * Writes out handling for the content-type header. The following rules apply:
      *
      *  - The content-type header may always be omitted.
-     *  - If the input shape has a member with the httpPaylaod trait then the following apply:
+     *  - If the input shape has a member with the httpPayload trait then the following apply:
      *      - If the target is a shape with the mediaType trait, the value of the content-type header must
      *        match if present.
      *      - If the target is a blob shape without a media type, the content-type header may have any value.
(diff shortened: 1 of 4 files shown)


adamthom-amzn (commented):
I think this is overall fine, my only concern is blobs that don't have a media type annotation. This code will enforce, I believe, `application/octet-stream` for those incoming requests, but there should be nothing restricting a service from accepting any content type for those requests. Imagine a service that wants to accept any type of image or video or audio upload; you wouldn't want to force the requester to use `application/octet-stream` if it knew the actual content type.

adamthom-amzn (changes requested):
You will also need to update https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/framework-errors.smithy to include the model for these exceptions so we can code gen a serializer for them.

JordonPhillips:
Updated so that blob payloads without a modeled media type don't make any assertions on the content type header. Also, while adding the models I noticed a discrepancy between the model and type for some of the existing ones, so I fixed that.

adamthom-amzn (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -1679,6 +1679,22 @@ private void generateOperationRequestDeserializer(
         writer.write("");
     }
 
+    /**
+     * Writes out handling for the content-type header. The following rules apply:
+     *
+     *  - The content-type header may always be omitted.
+     *  - If the input shape has a member with the httpPaylaod trait then the following apply:
```
adamthom-amzn, line 1686: s/httpPaylaod/httpPayload

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -1717,6 +1733,19 @@ private void handleContentType(
         });
     }
 
+    /**
+     * Writes out handling for the accept header. The following rules apply:
+     *
+     *  - The accept header may always be omitted.
+     *  - If the output shape has a member with the httpPaylaod trait then the following apply:
```
adamthom-amzn, line 1740: also here

adamthom-amzn (approved):

trivikr (approved):

PR #397, opened 2021-08-11
Update float/int parsing and expectations

JordonPhillips:
This updates number parsing to be more strict. It also updates the expectation methods used for numbers to be more specific.

This relies on: https://github.com/aws/aws-sdk-js-v3/pull/2662


adamthom-amzn:
Should you just change int/float? Eventually we're going to have to handle underflow/overflow, and if we put that in the parse method, you're going to have to change this again.

JordonPhillips:
> Should you just change int/float?

Incremental progress is still progress. Sure we'll add parse/expect short/byte/etc later

trivikr (approved):

PR #402, opened 2021-08-20
Use operation symbol name consistently

JordonPhillips:
This updates any usage of an operation name to go through the symbol provider. This avoids issues where the operation name's casing differs from that of the symbol.

Note: this has no impact on client generation, which already exclusively uses the symbol provider name.


trivikr (approved):

gosar (approved):
Hard without a diff of the generated code, but if you've tested against demo SSDK, the change LGTM.

PR #407, opened 2021-08-27
Add warnings when generating for SSDK longs

JordonPhillips:
This adds a warning when generating SSDK longs, because JS numbers can't accurately represent them.


AllanZhengYP (approved):
It looks good from my side. I will defer to you guys on the best way to show the validation. Printing error message makes sense to me.

PR #465, opened 2021-11-11
Be more specific about Smithy versions

adamthom-amzn:
*Description of changes:*
Use a specific version of smithy-cli, and restrict Smithy versions to the
bugfix version range of 1.12, instead of any minor version of 1, so that we
don't introduce mismatches in the Smithy version downstream.


AllanZhengYP (approved):
Nice that we don't need to update the smithy version is multiple files!

PR #476, opened 2021-12-15
feat(codegen): configurable package manager

### LINKED ISSUE #347: Allow build tool to be configurable
Currently generated packages have an implicit dependency on yarn because a number of scripts call out to it, but yarn isn't listed in the dependencies. Ideally one of the following things should happen:

* [`yarn` should be added as a dependency](https://github.com/awslabs/smithy-typescript/pull/343)
* The script runner should be configurable

A reasonable take would be to just use npm by default and make it configurable from there. You can use `packageJson` in the smithy config to override things now. That might be a reasonable way to handle changing to something else honestly.

adamthom-amzn:
*Issue #, if available:*

Resolves #347

*Description of changes:*
Generated packages all use yarn in their package.json, which is not always
available. This adds a configuration to the code generator called
packageManager, which can be set to yarn or npm, or omitted for the default,
which remains yarn for backwards compatibility.


JordonPhillips (approved):
nice

trivikr (approved):

PR #479, opened 2021-12-17
feat(codegen): add a default prepack script

adamthom-amzn:
*Description of changes:*
Doing a clean build in prepack works for both package managers and makes a
simple process of codegen && (npm|yarn) publish work out of the box. Also
adds an "omitPrepack" configuration option to allow the SDK to disable this
script for backwards compatibility.

### DIFF
--- a/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/base-package.json
+++ b/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/base-package.json
@@ -9,7 +9,8 @@
     "build:es": "tsc -p tsconfig.es.json",
     "build:types": "tsc -p tsconfig.types.json",
     "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
-    "clean": "rimraf ./dist-*"
+    "clean": "rimraf ./dist-*",
+    "prepack": "${packageManager} clean && ${packageManager} build"
   },
   "main": "./dist-cjs/index.js",
   "types": "./dist-types/index.d.ts",


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/PackageJsonGenerator.java
```diff
@@ -68,6 +68,10 @@ static void writePackageJson(
             node = node.withMember("private", true);
         }
 
+        if (settings.isOmitPrepack()) {
```
gosar, line 71: If the SDK doesn't need it, I think we should not provide this option.

AllanZhengYP (commented):
The change makes sense, but I'm not a big fan of the name `prepack`. We don't have a `pack` script, so the `pre` hook doesn't make a lot of sense. `pack` is a overloaded verb in JS ecosystem. What does it mean?

adamthom-amzn:
It's automatically run as a part of [npm publish](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-publish) and it's recommended by [yarn](https://yarnpkg.com/advanced/lifecycle-scripts) as a place to do transpilation before `pack`. So you don't have to define `pack` at all - it's a built-in part of the lifecycle of both package managers.

AllanZhengYP (approved):
Since we already defined a `files` in the package.json, which is an allowlist of all the paths to be pulished, do we still need to call the `clean` in the script?

adamthom-amzn:
> Since we already defined a `files` in the package.json, which is an allowlist of all the paths to be pulished, do we still need to call the `clean` in the script?

Wanted to make sure that `npm publish` didn't pick up a JavaScript file in `dist` for a TypeScript file that had been removed from the input.

AllanZhengYP (approved):
Just realize TSC doesn't remove the JS files from previous builds unless with `--clean`: https://github.com/microsoft/TypeScript/issues/30602#issuecomment-483872618

Ship it!

PR #519, opened 2022-02-23
Fix scripts for npm by extracting `run` command out 

gosar:
Currently, for `npm` the concurrently syntax was `concurrently 'npm run-script:build:cjs'` which is wrong and didn't work. `npm:build:cjs` (similar to `yarn:build:cjs`) is good syntax for concurrently.

Extracting run-script out of PackageManager.getCommand and specifying `run` directly in package.json scripts (prepack). `run` is equivalent alias to `run-script` in npm.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
@@ -395,7 +395,7 @@ public final class TypeScriptSettings {
 
     public enum PackageManager {
         YARN("yarn"),
-        NPM("npm run-script");
+        NPM("npm");
 
         private final String command;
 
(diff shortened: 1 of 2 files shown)


adamthom-amzn (approved):

AllanZhengYP (approved):

trivikr (changes requested):
What's the issue with `npm:`?
https://github.com/open-cli-tools/concurrently#usage

gosar:
> What's the issue with `npm:`? https://github.com/open-cli-tools/concurrently#usage

`concurrently 'npm:build:cjs'` would have worked, but https://github.com/awslabs/smithy-typescript/blob/d02ff5910ec9e23ee4344b6fcb50a66784b36e2e/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java#L398 is `npm run-script` which works for other parts like https://github.com/awslabs/smithy-typescript/blob/d02ff5910ec9e23ee4344b6fcb50a66784b36e2e/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/base-package.json#L13

Could be refactored, but seems this syntax is also fine, maybe clearer.

trivikr:
This is incorrect as `npm run-script` is not a packageManager
https://github.com/awslabs/smithy-typescript/blob/d02ff5910ec9e23ee4344b6fcb50a66784b36e2e/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java#L396-L398

We should create a new method on `PackageManager`, say `getRunCommand` which returns empty string (`""`) for yarn and `"run script"` for npm.

The prepack script can be changed to:
```
"prepack": "${packageManager} ${pmRunCommand} clean && ${packageManager} ${pmRunCommand} build"
```

and PackageJsonGenerator can contain the additional line:
```
template = template.replace("${pmRunCommand}", settings.getPackageManager().getRunCommand());
```

trivikr (approved):

PR #528, opened 2022-03-27
fix: missing export for api key auth middleware

eduardomourar:
*Issue #, if available:*

*Description of changes:*


eduardomourar:
@adamthom-amzn , could you help us with this PR, please?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/AddHttpApiKeyAuthPlugin.java
```diff
@@ -142,6 +143,24 @@ public void writeAdditionalFiles(
...
+        boolean isClientSdk = settings.generateClient();
+        ServiceShape service = settings.getService(model);
+        if (isClientSdk && hasEffectiveHttpApiKeyAuthTrait(model, service)) {
+            writer.write("export * from \"./middleware/$1L\";", getMiddlewarePath());
+        }
+    }
+
+    private String getMiddlewarePath() {
+        return "HttpApiKeyAuth";
+    }
```
adamthom-amzn, line 162: I think this is fair as a static constant, don't think it needs a private method

adamthom-amzn:
I'm not familiar with the use of this middleware - is the middleware not added automatically when the model has the trait applied? Do you have an example of why you would need to import something from this middleware in your code?

eduardomourar:
It only adds automatically the configuration for me, not the auth middleware.

Screenshot 2022-03-28 at 23 25 53

Also, I would like to make some customization to it (for instance, use it to populate a bearer token in the authorization header), so I have to use the middleware being autogenerated because there is no public library like [this one](https://github.com/aws/aws-sdk-js-v3/tree/main/packages/middleware-header-default).

trivikr (approved):

Issue #503, opened 2022-02-05
Members on models are always type or `undefined`


jabalsad:
Hello,

The generated typescript types for structures have members that are always either the type or `undefined`.

E.g.
```
export interface Metadata {
  id: string | undefined;
  createdAt: Date | undefined;
  updatedAt: Date | undefined;
  deleted: boolean | undefined;
  lastUpdatedBy: string | undefined;
}
```

Even though these members have a `@required` trait applied to them. Can you elaborate a bit on the motivation here and if there is a way to prevent this from happening?

adamthom-amzn:
It's been considered backwards compatible for Smithy services to remove `@required`. If we didn't include `undefined` in the type, client applications would have to make a code change to compensate when a service dropped `@required`.

We know this is a very annoying thing to have to work around in code. We have a proposal for a trait that allows this change to be made backwards-compatibly without including `undefined` (by falling back to the default value for the type). Unfortunately the TypeScript client code generator does not generate classes/constructors for types (as a size optimization), so there's no clear way for us to include a default, non-null value for members, so there may be no way to adopt this proposal backwards-compatibly when it lands.

agacek:
This unfortunately makes Smithy SDKs tedious to use. For those of us willing to sacrifice backwards compatibility, could there be a flag to allow the more precise SDK generation?

pcholakov:
> It's been considered backwards compatible for Smithy services to remove @required.

This policy only makes sense on input shapes; it's the inverse when it comes to outputs – introducing `@required` is backwards-compatible, removing it is a breaking change. Is the input vs. output delineation something that we should be factoring in? I suspect that most users find the `... | undefined` shapes a pain to work with on the response end of generated code.

To that end, I wonder if @eduardomourar's PR https://github.com/awslabs/smithy-typescript/pull/566 wouldn't be substantially easier to accept if it was limited in scope to just output structures. Am I missing something?

anderha:
I am currently evaluating to use the code generator and came across this behavior. Which seems to be a bug if I do understand the openapi and the smithy specs correctly:

- Openapi distinguishes between required and nullable (ref: https://swagger.io/docs/specification/data-models/data-types/)
  - required: "By default, all object properties are optional. You can specify the required properties"
  - nullable: "OpenAPI 3.0 does not have an explicit null type as in JSON Schema, but you can use nullable: true to specify that the value may be null" and "If the null value needs to be allowed, add nullable: true"
[...]

PR #537, opened 2022-04-26
feat: omit 'undefined' when generating SSDK and @required trait is used

RossWilliams:
*Issue #, if available:*

Addresses issues raised in #503 for SSDK code

*Description of changes:*

Currently when the client SDK is generated, @required shapes have their
types rendered with ' | undefined'. This allows removing @required as
a non-breaking change for client SDK users.

However, the server SDK (SSDK) has no need for backwards compatibility,
it can be updated as the service evolves. When generating SSDK types
the project does not need to include ' | undefined' for required types.

This update changes the structure member writer when the SSDK is 
generated to not include ' | undefined" when it is a required member.
Client SDK remains unaffected.

Server SDK does not appear to have any tests. Test frameworks do
not seem to support SSDK generation at this moment. only implementation
code is included.


RossWilliams:
This causes type issues inside the deserialise functions. Its not feasible to make these changes without a lot of internal plumbing.

PR #547, opened 2022-05-24
Do not use getModelWithoutTraitShapes

gosar:
The Walker does not traverse relationships create by traits anyways.

Ran `yarn generate-clients && yarn test:protocols && yarn generate-clients -s && yarn test:server-protocols` in aws-sdk-js-v3 with this change to confirm there is no change to clients. Though there is some rearrangement of code within a file in protocol tests. Here's the diff https://github.com/aws/aws-sdk-js-v3/pull/3633

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
@@ -50,7 +50,6 @@ import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.PaginatedTrait;
-import software.amazon.smithy.model.transform.ModelTransformer;
 import software.amazon.smithy.model.validation.ValidationEvent;
 import software.amazon.smithy.typescript.codegen.TypeScriptSettings.ArtifactType;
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java
```diff
@@ -180,7 +177,7 @@ void execute() {
         // Generate models that are connected to the service being generated.
         LOGGER.fine("Walking shapes from " + service.getId() + " to find shapes to generate");
         // Walk the tree.
-        Collection<Shape> shapeSet = new Walker(nonTraits).walkShapes(service);
+        Collection<Shape> shapeSet = new Walker(model).walkShapes(service);
 
         Model prunedModel = Model.builder().addShapes(shapeSet).build();
```
kstich, line 182: Is this necessary anymore? It was based upon the `nonTraits` model previously but that's gone, meaning places where this was used can just use `model` instead.

gosar: Still needed. Otherwise, the TopologicalIndex will iterate and visit through more than `shapeSet`.

gosar: Or I'll change it to use the `model` and check if shape is present in `shapeSet` before visiting.

mtdowling (approved):

trivikr (approved):

Issue #192, opened 2020-07-08
Generated TypeScript Client includes body for GET Request


davjohna:
I have a Smithy service definition containing a simple GET operation with no input shape (not sure if API Gateway traits could be relevant, if not, the operation is only the last three lines):

    @readonly
    @http(method: "GET", uri: "/sources")
    @aws.apigateway#integration(
      type: "aws",
      httpMethod: "POST",
      uri: "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${ServiceApiLambda}/invocations",
      credentials: "${ApiLambdaInvokeRoleArn}",
      requestTemplates: {
            "application/json": """
    {
      "requestId": "$context.extendedRequestId",
      "api": "ListLogSources",
      "requester": {
        "federatedUser": "$context.identity.cognitoAuthenticationProvider",
        "callerAccountId": "$context.identity.accountId"
      },
      "inputJson": "$util.escapeJavaScript($input.json('$'))"
    }"""
      },
      responses: {
        "default" : {
          "statusCode" : "200"
        },
        ".*errorType\":\"UnknownOperationException\".*": {
          "statusCode": "404",
          "responseTemplates" : {
            "application/json" : "#set ($errorMessageObj = $util.parseJson($input.path('$.errorMessage')))\n{\n  \"errorType\" : \"$errorMessageObj.errorType\",\n  \"message\" : \"$errorMessageObj.errorMessage\"\n}"
          }
        },
        ".*errorType\":\"UnauthorizedOperationException\".*": {
          "statusCode": "403",
          "responseTemplates" : {
            "application/json" : "#set ($errorMessageObj = $util.parseJson($input.path('$.errorMessage')))\n{\n  \"errorType\" : \"$errorMessageObj.errorType\",\n  \"message\" : \"$errorMessageObj.errorMessage\"\n}"
          }
        }
      }
    )
    operation ListLogSources {
        output: ListLogSourcesOutput
    }

Building this results in a JSON file with an operation that also has no input:

    "com.fooservice#ListSources": {
        "type": "operation",
        "output": {
            "target": "com.fooservice#ListSources"
        },
        "traits": {
            "smithy.api#http": {
                "method": "GET",
                "uri": "/sources",
                "code": 200
            },
            "smithy.api#readonly": {}
        }
    },

My service definition in that same JSON file has these traits:

    "aws.api#service": {
      "sdkId": "Foo",
      "arnNamespace": "execute-api",
[...]

PR #553, opened 2022-05-27
fix: only write default input body if operation has payload bindings

### LINKED ISSUE #552: GET request without input adds 'Content-Type' header and empty string body
Http restJson Operations without any bindings are generating `content-type` headers and empty string bodies, causing 415 server errors.

Test case:
```smithy
@restJson1
service Health {
    version: "2022-03-09",
    operations: [PublicHealthCheck],
}

@readonly
@http(uri: "/public-health", method: "GET")
operation PublicHealthCheck {
    output: PublicHealthCheckOutput,
    errors: [
        ValidationException
    ]
}

@output
structure PublicHealthCheckOutput{
    @required
    sha: String,
}
```

Generated code for serialising the command, notice content-type header and body set to empty string:
```typescript
export const serializeAws_restJson1PublicHealthCheckCommand = async(
  input: PublicHealthCheckCommandInput,
  context: __SerdeContext
): Promise<__HttpRequest> => {
  const {hostname, protocol = "https", port, path: basePath} = await context.endpoint();
  const headers: any = {
    'content-type': "application/json",
  };
  let resolvedPath = `${basePath?.endsWith('/') ? basePath.slice(0, -1) : (basePath || '')}` + "/public-health";
  let body: any;
  body = "";
  return new __HttpRequest({
    protocol,
    hostname,
    port,
    method: "GET",
    headers,
    path: resolvedPath,
    body,
  });
}
```

The error appears to be with the `shouldWriteDefaultBody` returning true in cases where there are no bindings, when in fact it should return false.

The code path I'm hitting is the following [(source)](https://github.com/awslabs/smithy-typescript/blob/5f83ba1fdbd8746fb9dfdbbf1111db640191dffd/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java#L894):
```java
// If we need to write a default body then it needs a content type.
if (!optionalContentType.isPresent() && shouldWriteDefaultBody(context, operationOrError, isInput)) {
    optionalContentType = Optional.of(getDocumentContentType());
}
optionalContentType.ifPresent(contentType -> context.getWriter().write("'content-type': $S,", contentType));
``` 

`shouldWriteDefaultBody` is returning true, causing a content-type to be written. The documentation for this method states: `By default no body will be written if there are no members bound to the input.`. I interpret this to mean that a default body should be written if there are members bound to the input. The code however returns `true` for the opposite case [(source)](https://github.com/awslabs/smithy-typescript/blob/5f83ba1fdbd8746fb9dfdbbf1111db640191dffd/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java#L946)

Note that the body writer also uses `shouldWriteDefaultBody`, creating the empty body string.

**Should `shouldWriteDefaultBody` be switched to return `true` when request bindings do exist?**

Switching the logic, the generated code produces the following, which is correct for my use case and is valid according to the type definitions:
```typescript
export const serializeAws_restJson1PublicHealthCheckCommand = async(
  input: PublicHealthCheckCommandInput,
  context: __SerdeContext
): Promise<__HttpRequest> => {
  const {hostname, protocol = "https", port, path: basePath} = await context.endpoint();
  const headers: any = {
  };
  let resolvedPath = `${basePath?.endsWith('/') ? basePath.slice(0, -1) : (basePath || '')}` + "/public-health";
  let body: any;
  return new __HttpRequest({
    protocol,
    hostname,
    port,
    method: "GET",
    headers,
    path: resolvedPath,
    body,
  });
}
```

EDIT: The proposed change would ADD content-type headers and request bodies if the operation includes only query string parameters. This is a regression and a more complex solution is required.

RossWilliams:
fixes #552

Before this change, when an operation has no input defined, a GET
operation would include a 'content-type' header and an empty body. This
will cause servers to respond with a 415 error, as the server is not
expected content in the body of the request.

The code to determine if a default input body should be written had
faulty logic. It returned true to write an empty body if there are
no request bindings. Instead, a default body should be written if there
are request bindings in the payload.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -935,7 +935,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
 
     /**
      * Given a context and operation, should a default input body be written. By default no body will be written
-     * if there are no members bound to the input.
+     * if there are no members bound to the input in the payload.
      *
      * @param context The generation context.
      * @param operation The operation whose input is being serialized.
@@ -943,7 +943,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
      * @return True if a default body should be generated.
      */
     protected boolean shouldWriteDefaultInputBody(GenerationContext context, OperationShape operation) {
-        return HttpBindingIndex.of(context.getModel()).getRequestBindings(operation).isEmpty();
+        return HttpBindingIndex.of(context.getModel()).getRequestBindings(operation, Location.PAYLOAD).size() > 0;
     }
 
     /**


RossWilliams:
Seems to fix: #192 as well

RossWilliams:
Hello, this has been open for 4 months. If the proposed change is unclear, or if this change is not wanted could you please let me know so that I can pursue alternative methods to fix the issue.

zeyad001:
Any update on the proposed changes?

trivikr:
Code Review is blocked on response to https://github.com/awslabs/smithy-typescript/pull/553#discussion_r938206650

cc author @RossWilliams

Issue #560, opened 2022-06-07
Newer NPM version fails to install generated client SDK


eduardomourar:
Whenever running `npm install` in NPM v8+ for a freshly generated client SDK, we get the following error:

> npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: client-sdk@0.1.0
npm ERR! Found: typescript@4.6.4
npm ERR! node_modules/typescript
npm ERR!   dev typescript@"~4.6.2" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer typescript@"3.9.x || 4.0.x" from typedoc@0.19.2
npm ERR! node_modules/typedoc
npm ERR!   dev typedoc@"^0.19.2" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

trivikr:
Fixed in https://github.com/awslabs/smithy-typescript/pull/561

PR #561, opened 2022-06-07
fix: override typescript version for typedoc

eduardomourar:
*Issue #, if available:* #560

*Description of changes:*

We will add the proper sub-dependencies overrides for both NPM and Yarn. This way we will not need to install with force resolution enabled.


eduardomourar:
Removed yarn `resolutions` property because it only gives a warning during installation and also resolution for peer dependencies is not yet supported (details [here](https://github.com/yarnpkg/berry/issues/4099)).

trivikr (approved):

PR #566, opened 2022-06-28
feat: allow non backward compatible required member

eduardomourar:
*Issue #, if available:* #503

*Description of changes:*

A continuation of PR #537, but we will allow to remove `| undefined` by setting an input configuration called `requiredMemberMode` with the following possible values:

- **nullable**: this is the current behavior and it will be the default. When set, it allows a member that has the `@required` trait applied to be `undefined`. By doing so it can still be considered a backwards compatible change for Smithy services even when the required constraint is dropped from a member .
- **strict**: this will remove `| undefined` from the generated code.

### DIFF
--- a/smithy-typescript-codegen-test/smithy-build.json
+++ b/smithy-typescript-codegen-test/smithy-build.json
@@ -9,7 +9,7 @@
             "packageJson": {
                 "license": "Apache-2.0"
             },
-            "memberNullabilityCompatibilityMode": "strict"
+            "requiredMemberMode": "strict"
         }
     }
 }
--- a/smithy-typescript-codegen-test/smithy-build.json
+++ b/smithy-typescript-codegen-test/smithy-build.json
@@ -8,7 +8,8 @@
             "packageVersion": "0.0.1",
             "packageJson": {
                 "license": "Apache-2.0"
-            }
+            },
+            "compatibilityMode": "strict"
         }
     }
 }
--- a/smithy-typescript-codegen-test/smithy-build.json
+++ b/smithy-typescript-codegen-test/smithy-build.json
@@ -9,7 +9,7 @@
             "packageJson": {
                 "license": "Apache-2.0"
             },
-            "compatibilityMode": "strict"
+            "memberNullabilityCompatibilityMode": "strict"
         }
     }
 }
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
@@ -71,7 +71,7 @@ final class StructureGenerator implements Runnable {
     private final RequiredMemberMode requiredMemberMode;
 
     /**
-     * sets 'includeValidation' to 'false' and backwards compatibility
+     * sets 'includeValidation' to 'false' and requiredMemberMode
      * to {@link RequiredMemberMode#NULLABLE}.
      */
     StructureGenerator(Model model, SymbolProvider symbolProvider, TypeScriptWriter writer, StructureShape shape) {
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
@@ -27,7 +27,7 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
-import software.amazon.smithy.typescript.codegen.TypeScriptSettings.CompatibilityMode;
+import software.amazon.smithy.typescript.codegen.TypeScriptSettings.MemberNullabilityCompatibilityMode;
 import software.amazon.smithy.typescript.codegen.integration.HttpProtocolGeneratorUtils;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
@@ -68,13 +68,15 @@ final class StructureGenerator implements Runnable {
     private final TypeScriptWriter writer;
     private final StructureShape shape;
     private final boolean includeValidation;
-    private final CompatibilityMode compatibilityMode;
+    private final MemberNullabilityCompatibilityMode memberNullabilityCompatibilityMode;
 
     /**
-     * sets 'includeValidation' to 'false' and backwards compatibility to {@link CompatibilityMode#RELAXED}.
+     * sets 'includeValidation' to 'false' and backwards compatibility
+     * to {@link MemberNullabilityCompatibilityMode#RELAXED}.
      */
     StructureGenerator(Model model, SymbolProvider symbolProvider, TypeScriptWriter writer, StructureShape shape) {
-        this(model, symbolProvider, writer, shape, false, CompatibilityMode.RELAXED);
+        this(model, symbolProvider, writer, shape, false,
+            MemberNullabilityCompatibilityMode.RELAXED);
     }
 
     StructureGenerator(Model model,
@@ -82,13 +84,13 @@ final class StructureGenerator implements Runnable {
                        TypeScriptWriter writer,
                        StructureShape shape,
                        boolean includeValidation,
-                       CompatibilityMode compatibilityMode) {
+                       MemberNullabilityCompatibilityMode memberNullabilityCompatibilityMode) {
         this.model = model;
         this.symbolProvider = symbolProvider;
         this.writer = writer;
         this.shape = shape;
         this.includeValidation = includeValidation;
-        this.compatibilityMode = compatibilityMode;
+        this.memberNullabilityCompatibilityMode = memberNullabilityCompatibilityMode;
     }
 
     @Override
@@ -162,7 +164,7 @@ final class StructureGenerator implements Runnable {
         }
 
         StructuredMemberWriter config = new StructuredMemberWriter(
-                model, symbolProvider, shape.getAllMembers().values(), this.compatibilityMode);
+                model, symbolProvider, shape.getAllMembers().values(), this.memberNullabilityCompatibilityMode);
         config.writeMembers(writer, shape);
         writer.closeBlock("}");
         writer.write("");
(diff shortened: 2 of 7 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -65,6 +66,7 @@ public final class TypeScriptSettings {
     private boolean isPrivate;
     private ArtifactType artifactType = ArtifactType.CLIENT;
     private boolean disableDefaultValidation = false;
+    private boolean backwardCompatibleRequiredMember = true;
```
srchase, line 69: Can this be converted to an enum so that we can add additional modes in the future? I think the mode you're adding to remove `| undefined` could be `strict` or something similar.

There's an enum in NullableIndex as an example that has allowed for adding additional check modes as it has evolved: https://github.com/awslabs/smithy/blob/8beb77af9e9dd1ca8cdb822f3a99da5ecff24321/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java#L54-L147

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -295,6 +299,19 @@ public void setDisableDefaultValidation(boolean disableDefaultValidation) {
         this.disableDefaultValidation = disableDefaultValidation;
     }
 
+    /**
+     * Returns whether or not required members are backward-compatible.
+     *
+     * @return true if required members are backward-compatible. Default: true
+     */
+    public boolean isBackwardCompatibleRequiredMember() {
```
srchase, line 307: The name, `BackwardCompatibleRequiredMember`, isn't descriptive of what this is actually doing. Is there another suggestion that will make this more obvious when it shows up in smithy-build.json?

srchase (approved):

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/StructureGeneratorTest.java
```diff
@@ -521,6 +533,7 @@ private String testStructureCodegen(String file, String expectedType) {
                         .withMember("service", Node.from("smithy.example#Example"))
                         .withMember("package", Node.from("example"))
                         .withMember("packageVersion", Node.from("1.0.0"))
+                        .withMember("backwardCompatibleRequiredMember", Node.from(backwardCompatibleRequiredMember))
```
srchase, line 536: Echoing above, with this naming, `backwardCompatibleRequiredMember`, it's hard to understand what this will change in the generated code.

srchase (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
eduardomourar: what about calling it: `backwardCompatibilityMode`?

srchase: That makes more sense. Maybe `memberNullabilityCompatibilityMode`?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
```diff
@@ -67,24 +67,27 @@ final class StructureGenerator implements Runnable {
     private final TypeScriptWriter writer;
     private final StructureShape shape;
     private final boolean includeValidation;
+    private final boolean backwardCompatibleRequiredMember;
 
     /**
      * sets 'includeValidation' to 'false' for backwards compatibility.
```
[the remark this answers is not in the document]
srchase: Can this also emit a warning to the logs when this is used? We want to be sure that someone who changes this mode is understanding the potential impacts on previously generated clients if their model changes later on.

smithy-typescript-codegen-test/smithy-build.json
```diff
@@ -8,7 +8,8 @@
             "packageVersion": "0.0.1",
             "packageJson": {
                 "license": "Apache-2.0"
-            }
+            },
+            "compatibilityMode": "strict"
```
srchase, line 12: Can this be made more specific? Glancing at smithy-build.json, I don't have a good idea what compatibility this is will impact.

srchase (changes requested):

srchase:
> _Issue #, if available:_ #503
> 
> _Description of changes:_
> 
> A continuation of PR #537, but we will allow to remove `| undefined` by setting an input configuration called `backwardCompatibleRequiredMember` to `false`.
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Can the overall PR description get updated with the enums and what impact each has on the generated code?

srchase (dismissed):

smithy-typescript-codegen-test/smithy-build.json
```diff
@@ -8,7 +8,8 @@
             "packageVersion": "0.0.1",
             "packageJson": {
                 "license": "Apache-2.0"
-            }
+            },
+            "memberNullabilityCompatibilityMode": "strict"
```
srchase, line 12: I took some more time to think about this and discuss with @gosar. I think `requiredMemberMode` makes more sense, with enum values of `strict` and `nullable`.

srchase (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -407,6 +438,44 @@ public SymbolProvider createSymbolProvider(Model model, TypeScriptSettings setti
...
+         * This is the current behavior and it will be the default. When set,
+         * it allows a member that has the {@link RequiredTrait} applied to be "undefined".
+         * By doing so it can still be considered a backwards compatible change fo
+         * Smithy services even when the required constraint is dropped from a member .
+         */
+        RELAXED("relaxed"),
+
+        /**
+         * This will dissallow members marked as {@link RequiredTrait} to be "undefined".
+         */
```
gosar, line 455: I think there should be some warning language here too calling out the risks of using this mode. The risks are that when a server drops `@required` from an output shape (and replaces it with `@default`), if the server does not always serialize a value, customer code consuming the client and trying to access this member, may get a NullPointerException. Smithy [spec](https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization) says
> Authoritative model consumers like servers SHOULD always serialize default values to remove any ambiguity about the value of the most up to default value.

So one should use this mode on the client, only if the server is following this.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructureGenerator.java
```diff
@@ -67,24 +68,29 @@ final class StructureGenerator implements Runnable {
     private final TypeScriptWriter writer;
     private final StructureShape shape;
     private final boolean includeValidation;
+    private final RequiredMemberMode requiredMemberMode;
 
     /**
-     * sets 'includeValidation' to 'false' for backwards compatibility.
+     * sets 'includeValidation' to 'false' and backwards compatibility
```
gosar, line 74:
```suggestion
     * sets 'includeValidation' to 'false' and requiredMemberMode
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -295,6 +304,28 @@ public void setDisableDefaultValidation(boolean disableDefaultValidation) {
         this.disableDefaultValidation = disableDefaultValidation;
     }
 
+    /**
+     * Returns the compatibility mode in regards to required members..
+     *
+     * @return the configured compatibility mode in regards to required members..
```
gosar, line 310:
```suggestion
     * Returns the code generation mode for required members.
     *
     * @return the configured mode for required members.
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -295,6 +304,28 @@ public void setDisableDefaultValidation(boolean disableDefaultValidation) {
...
+     * Defaults to {@link RequiredMemberMode#NULLABLE}
+     */
+    public RequiredMemberMode getRequiredMemberMode() {
+        return requiredMemberMode;
+    }
+
+    public void setRequiredMemberMode(
+        RequiredMemberMode requiredMemberMode) {
+            if (requiredMemberMode != RequiredMemberMode.NULLABLE) {
+                LOGGER.warning(String.format("By setting the member nullability compatibility mode to '%s', a"
```
gosar, line 320:
```suggestion
                LOGGER.warning(String.format("By setting the required member mode to '%s', a"
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -295,6 +304,28 @@ public void setDisableDefaultValidation(boolean disableDefaultValidation) {
...
+                    + " member that has the '@required' trait applied CANNOT be 'undefined'. Differrent than"
+                    + " when it is set to '%s', it will be considered a BACKWARDS INCOMPATIBLE change for"
```
gosar, line 322: I think the "Different than when it is set to NULLABLE" is making it bit confusing to read and not useful IMO. I'd just drop that part.

```suggestion
                    + " member that has the '@required' trait applied CANNOT be 'undefined'."
                    + " It will be considered a BACKWARDS INCOMPATIBLE change for"
```
with reformatting.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java
```diff
@@ -407,6 +439,52 @@ public SymbolProvider createSymbolProvider(Model model, TypeScriptSettings setti
         }
     }
 
+    /**
+     * An enum indicating the compatibility mode in regards to required members..
```
gosar, line 443:
```suggestion
     * An enum indicating the code generation mode for required members.
```

srchase (approved):

PR #588, opened 2022-09-13
feat(smithy-typescript-codegen): allow deferred resolution for api key config

eduardomourar:
*Issue #, if available:* N/A

*Description of changes:*

In order to defer resolving the API Key configuration to when we actually make the API call, we need to allow a function to be passed in the `apiKey` property. Now that we are importing the library `@aws-sdk/util-middleware` as part of the API Key middleware, we included it explicitly as a dependency to our generated project.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/AddHttpApiKeyAuthPlugin.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/AddHttpApiKeyAuthPlugin.java
@@ -32,6 +32,7 @@ import software.amazon.smithy.model.traits.HttpApiKeyAuthTrait;
 import software.amazon.smithy.model.traits.OptionalAuthTrait;
 import software.amazon.smithy.typescript.codegen.CodegenUtils;
 import software.amazon.smithy.typescript.codegen.TypeScriptCodegenContext;
+import software.amazon.smithy.typescript.codegen.TypeScriptDependency;
 import software.amazon.smithy.typescript.codegen.TypeScriptSettings;
 import software.amazon.smithy.typescript.codegen.TypeScriptWriter;
 import software.amazon.smithy.utils.IoUtils;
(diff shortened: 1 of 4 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/AddHttpApiKeyAuthPlugin.java
```diff
@@ -123,6 +124,7 @@ public void writeAdditionalFiles(
         writerFactory.accept(
                 Paths.get(CodegenUtils.SOURCE_FOLDER, "middleware", INTEGRATION_NAME, "index.ts").toString(),
                 writer -> {
+                        writer.addDependency(TypeScriptDependency.AWS_SDK_UTIL_MIDDLEWARE);
```
gosar, line 127: If I look at [http-api-key-auth.ts](https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/integration/http-api-key-auth.ts) I don't see @aws-sdk/util-middleware used. What am I missing?

eduardomourar: True. It is because we are working on the bearer token configuration and that needs a Promise to passed instead of a static value.

eduardomourar: I have pushed the changes to `http-api-key-auth.ts` file now.

smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/integration/http-api-key-auth.ts
```diff
@@ -82,10 +84,12 @@ export const httpApiKeyAuthMiddleware =
   async (args) => {
     if (!HttpRequest.isInstance(args.request)) return next(args);
 
+    const apiKey = pluginConfig.apiKey && await pluginConfig.apiKey();
+
     // This middleware will not be injected if the operation has the @optionalAuth trait.
     // We don't know if we're the only auth middleware, so let the service deal with the
     // absence of the API key (or let other middleware do its job).
```
gosar, line 91: couple of updates needed to the documentation of this method.
1. `Must include API key` seem out dated
2. the param are incorrect from the method definition

gosar (commented):
LG pending minor documentation comment

smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/integration/http-api-key-auth.ts
eduardomourar: done

gosar (approved):

kuhe (approved):

PR #590, opened 2022-09-16
feat(endpoint): expose static endpoint param instructions provider

kuhe:
This exposes the static instructions object for each command so that it can be used by other internal packages like s3 presigner and lib-upload that need to resolve the v2 endpoint of e.g. `PutObjectCommand` without necessarily `send`ing an instance of that command.

before:
```js
public resolveMiddleware() {
    this.middlewareStack.use(getEndpointPlugin(configuration, { /* instructions literal object */ })  
}
```

after:
```js
public static instructions(): EndpointParameterInstructions {
    return { /* instructions literal object */ };
}

public resolveMiddleware() {
    this.middlewareStack.use(getEndpointPlugin(configuration, Command.instructions())  
}
```

Since the instructions are not tied to individual Command instances, the function can be static.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -183,7 +183,7 @@ final class CommandGenerator implements Runnable {
         }
         writer.addImport("EndpointParameterInstructions", null, "@aws-sdk/middleware-endpoint");
         writer.openBlock(
-            "public static instructions(): EndpointParameterInstructions {", "}",
+            "public static getEndpointParameterInstructions(): EndpointParameterInstructions {", "}",
             () -> {
                 writer.openBlock(
                     "return {", "};",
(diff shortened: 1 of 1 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -172,32 +177,16 @@ private void generateCommandConstructor() {
...
-                    "getEndpointPlugin",
-                    "@aws-sdk/middleware-endpoint"
-                );
+    private void generateEndpointParameterInstructionProvider() {
+        if (!service.hasTrait(EndpointRuleSetTrait.class)) {
+            return;
+        }
+        writer.addImport("EndpointParameterInstructions", null, "@aws-sdk/middleware-endpoint");
+        writer.openBlock(
+            "public static instructions(): EndpointParameterInstructions {", "}",
```
trivikr, line 186: Be more specific, like `getEndpointParameterInstructions`?
```suggestion
            "public static getEndpointParameterInstructions(): EndpointParameterInstructions {", "}",
```

trivikr (approved):

PR #592, opened 2022-09-21
fix(endpoint): fixes to endpoints codegen from user testing

kuhe:
- codegen fixes for endpoints 2.0


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -73,7 +73,7 @@ private void generateEndpointParameters() {
                     new RuleSetParameterFinder(service).getClientContextParams();
 
                 clientContextParams.forEach((k, v) -> {
-                    writer.write("$L: $L,", k, v);
+                    writer.write("$L?: $L,", k, v);
```
kuhe, line 76: fix: making the clientContextParams optional at the service client constructor level

trivikr (approved):

PR #596, opened 2022-09-28
Populate optional $isError to true when parsing error structures

trivikr:
*Issue #, if available:*
Internal JS-3034

*Description of changes:*
Populate optional `$isError` to true when parsing error structures. This value will allow parse function to populate message for errors.


trivikr:
Superceded by https://github.com/awslabs/smithy-typescript/pull/597

PR #600, opened 2022-09-29
fix(endpoint): misc fixes for endpoints 2.0 based on service unit test cases

kuhe:
- write out more params in the generated interface instead of only client context params
- add param defaults to the generated E2.0 config resolver step

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ParameterGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ParameterGenerator.java
@@ -40,7 +40,6 @@ public class ParameterGenerator {
             .orElseThrow(() -> new RuntimeException("param node is not object node."));
 
         Optional<BooleanNode> requiredNode = paramNode.getBooleanMember("required");
-
         requiredNode.ifPresent(booleanNode -> required = booleanNode.getValue());
 
         Optional<StringNode> type = paramNode.getStringMember("type");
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ParameterGenerator.java
```diff
@@ -63,6 +62,36 @@ public boolean isBuiltIn() {
...
+                break;
+            case "Boolean":
+                buffer += paramNode.expectBooleanMember("default").getValue() ? "true" : "false";
+                break;
+            default:
+                // required by linter
+        }
+
+        buffer += ",";
+        return buffer;
```
trivikr, line 92: nit: Use StringBuilder?

```suggestion
        StringBuilder defaultAsCode = new StringBuilder().append(parameterName)
            .append(": options.").append(parameterName).append(" ?? ");
        ObjectNode paramNode = param.expectObjectNode();
        StringNode type = paramNode.expectStringMember("type");

        switch (type.getValue()) {
            case "String":
                defaultAsCode
                    .append("\"")
                    .append(paramNode.expectStringMember("default").getValue())
                    .append("\"");
                break;
            case "Boolean":
                defaultAsCode
                    .append(paramNode.expectBooleanMember("default").getValue() ? "true" : "false");
                break;
            default:
                // required by linter
        }

        defaultAsCode.append(",");
        return defaultAsCode.toString();
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ParameterGenerator.java
```diff
@@ -76,7 +105,7 @@ public Map.Entry<String, String> getNameAndType() {
     public String toCodeString() {
         String buffer = "";
```
trivikr, line 106: nit: StringBuilder can be used here too, but this diff is not touching the buffer.

trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ParameterGenerator.java
kuhe: If this is for performance and not readability, I would push back on the change. 
I believe string concat is analyzable and optimizable by current Java versions to be effectively identical to appends. 
I would also contend that this is a very small string and this code is not a bottleneck. 

If you prefer the readability of append, then I will accept the change.

PR #601, opened 2022-09-29
Update versions in README example

milesziemer:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ are as follows:
             mavenCentral()
         }
         dependencies {
-            "classpath"("software.amazon.smithy:smithy-cli:[1.19.0,1.20.0[")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,21 +70,23 @@ are as follows:
...
     service ExampleService {
-        version: "2022-01-01",
+        version: "2022-01-01"
         operations: [Echo]
     }
 
     operation Echo {
-        input: EchoInput,
-        output: EchoOutput,
+        input: EchoInput
```
gosar, line 83: let's also update to use inline Input/output

milesziemer: Updated. We could also extract `message: String` to a mixin, but I don't have a strong opinion on whether or not to do that

gosar (approved):

kuhe (approved):

PR #616, opened 2022-10-18
fix(endpoint): deduplicate client context and built-in params with the same name

kuhe:
- fix: dedupe clientContextParams with builtInParams with the same name
- fix for JS conventions: default local param name to `camelCase` if it is `TitleCase`


kuhe:
https://github.com/aws/aws-sdk-js-v3/pull/4051

trivikr (approved):

PR #617, opened 2022-10-19
fix(endpoint): do not modify context param names

kuhe:
don't rename context params like `Bucket` to `bucket`.


kuhe:
tested with `yarn test:e2e` in client-s3 with 
```
AWS_SMOKE_TEST_REGION=... \
AWS_SMOKE_TEST_BUCKET=... \
AWS_SMOKE_TEST_MRAP_ARN=... \
```

trivikr (approved):

PR #639, opened 2022-11-17
fix: missing dom type in tsconfig

eduardomourar:
*Issue #, if available:* https://github.com/aws/aws-sdk-js-v3/issues/2896

*Description of changes:*

After PR #625, the DOM type has been removed from the default tsconfig.json file (before the value for lib included the DOM). This causes the following error during compilation:
> src/models/models_0.ts(1162,67): error TS2304: Cannot find name 'Blob'.

This is change will allow compiling any recent generated Smithy project without modifying the tsconfig (as being done in the AWS SDK JS v3 repo [here](https://github.com/aws/aws-sdk-js-v3/blob/7ac3f19ee0f6e16b825fe4a22177d5e8a4af8b2a/tsconfig.json#L7)). 

### DIFF
--- a/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/tsconfig.json
+++ b/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "@tsconfig/node14/tsconfig.json",
   "compilerOptions": {
+    "lib": ["es2020", "dom"],
     "downlevelIteration": true,
     "importHelpers": true,
     "incremental": true,


eduardomourar:
Most probably the DOM types can be removed from the `tsconfig.es.json` file too.

trivikr:
@eduardomourar Can you create a new [bug report](https://github.com/aws/aws-sdk-js-v3/issues/new?assignees=&labels=bug%2Cneeds-triage&template=bug-report.yml&title=TITLE+FOR+BUG+REPORT) for this issue

From your PR description, I see that the issue exists, but would like to discuss the problem using a minimal repro and compare different solutions.

AndrewFossAWS (approved):

PR #654, opened 2022-12-08
Add validation for intEnum shapes

milesziemer:
This PR adds an IntegerEnumValidator, IntegerEnumValidationFailure and the relevant tests to the ssdk libs. StructuredMemberWriter was updated to use IntegerEnumValidator when generating shape validators for intEnum shapes.

Local testing:
- Update js sdk to use Smithy 1.26.4, which has the protocol tests for intEnum
- Update js sdk to use local version of smithy-typescript-ssdk-libs, specificaly the server protocol test packages
- Run `yarn generate-clients --server-artifacts` and `yarn test:server-protocols`


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
```diff
@@ -473,7 +474,9 @@ private void writeShapeValidator(TypeScriptWriter writer,
                                      Shape shape,
                                      Collection<Trait> constraints,
                                      String trailer) {
-        if (constraints.isEmpty()) {
+        boolean shouldWriteIntEnumValidator = shape.isIntEnumShape();
+
+        if (constraints.isEmpty() && !shouldWriteIntEnumValidator) {
```
haydenbaker, line 479: Are you not able to add it as a constraint instead of adding an override here?

milesziemer: Not from what I can tell. The constraints are Traits, but there is no trait for intEnum. The [EnumTrait](https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumTrait.java) is deprecated too, so ideally we would handle enum and intEnum shapes like the other shapes in [writeMemberValidator](https://github.com/awslabs/smithy-typescript/blob/f13902522ef11ea13212728bdab3185e49755837/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java#L402). The issue with that is, there may still be constraint traits on the shape, and for enum shapes, that will still include the `@enum` trait, so we'd have to like exclude it from the constraints or something. I agree though that handling intEnum like the way I did is a little funky, I'm open to other options. Both intEnum and enum shapes will trigger [this guard](https://github.com/awslabs/smithy-typescript/blob/f13902522ef11ea13212728bdab3185e49755837/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java#L406), which would make it weird to try doing a `shape.isIntEnumShape()` or `shape.isEnumShape()` in `writeMemberValidator`.

srchase (approved):

PR #660, opened 2022-12-20
Fix restXml protocol test for timestampFormat targets

srchase:
Currently, protocol tests generated and run against Smithy 1.26.3 or later result in the following error:
```
src/protocols/Aws_restXml.ts(2199,27): error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.
```
This is due to the protocol test addition in https://github.com/awslabs/smithy/pull/1440.

This PR updates the DocumentMememberSerVisitor to append `.toString()` so that generated input parameter serializers are converted to strings where necessary for restXml protocols.

To test, these smithy-typescript changes were published locally with `./gradlew clean build pTML` before Smithy dependencies were updated in https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/gradle.properties#L1 to `1.26.4` and the following was successfully run: `yarn generate-clients && yarn test:protocols` from the aws-sdk-js-v3 repository which updated the generated `Aws_restXml.ts` file and allowed the protocol tests to succeed.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -1413,8 +1413,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
                 throw new CodegenException("Unexpected named member shape binding location `" + bindingType + "`");
         }
 
-        String baseParam = HttpProtocolGeneratorUtils.getTimestampInputParam(context, dataSource, member, format);
-        return baseParam + ".toString()";
+        return HttpProtocolGeneratorUtils.getTimestampInputParam(context, dataSource, member, format);
     }
 
     /**
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
@@ -82,7 +82,7 @@ public final class HttpProtocolGeneratorUtils {
                 // Use the split to not serialize milliseconds.
                 return "(" + dataSource + ".toISOString().split('.')[0]+\"Z\")";
             case EPOCH_SECONDS:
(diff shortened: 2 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/DocumentMemberSerVisitor.java
```diff
@@ -216,7 +216,11 @@ public final String memberShape(MemberShape shape) {
     public String timestampShape(TimestampShape shape) {
         HttpBindingIndex httpIndex = HttpBindingIndex.of(context.getModel());
         Format format = httpIndex.determineTimestampFormat(shape, Location.DOCUMENT, defaultTimestampFormat);
-        return HttpProtocolGeneratorUtils.getTimestampInputParam(context, dataSource, shape, format);
+        String base = HttpProtocolGeneratorUtils.getTimestampInputParam(context, dataSource, shape, format);
+        if (format.equals(Format.EPOCH_SECONDS)) {
+            base = base + ".toString()";
+        }
```
milesziemer, line 222: I wonder if this logic should be in [HttpProtocolGeneratorUtils.getTimestampInputParam](https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java#L74)

milesziemer: Specifically, it seems like within that function, both DATE_TIME and HTTP_DATE return what would be a string. Unless there's somewhere where EPOCH_SECONDS needs to be a number?

srchase: That makes sense. That will also allow for dropping the `.toString()` that's added [here](https://github.com/awslabs/smithy-typescript/blob/c874a2a0d60a9496ce9b59915f9ed4460e76b5b1/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java#L1417) which would clean up some extra `.toString()` calls that aren't really needed.

milesziemer (approved):

PR #661, opened 2022-12-22
Fix codegen windows

milesziemer:
*Issue #, if available:*
https://github.com/awslabs/smithy-typescript/issues/568
*Description of changes:*
The use of `Paths.get` when creating an import/export path, ie. `import { Foo } from ./path`, caused tests to fail on Windows because the tests expect unix-like paths. For portability and consistency, the places where import/export paths are created in `SymbolVisitor` and `ImportDeclarations` were refactored to always create unix-like path separators regardless of OS.

Additionally, CI was added for Windows (and MacOS since it wasn't there already).

*Testing details:*
- Ran `./gradlew clean build` successfully on Windows and Mac

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,12 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-    name: Java ${{ matrix.java }}
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


milesziemer:
Opened this as a draft because it feels a little flimsy. There's a few places where we have to do the same string manipulation, that I think could be refactored, but I wasn't sure exactly how to do it.

srchase (approved):

PR #664, opened 2023-01-03
Add smithy as codeowners for the whole repo

milesziemer:
*Issue #, if available:*
N/A
*Description of changes:*
Makes Smithy codeowners on the entire repo, rather than just the ssdk stuff. 

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,13 +1,2 @@
-# The JS team owns everything by default
 * @awslabs/aws-sdk-js-team
-
-# These libraries are only intended for use with the SSDK.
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,13 +1,3 @@
 # The JS team owns everything by default
```
srchase, line 1: I'd be good to update this comment to call out why smithy is also a codeowner.

milesziemer: I just removed the comment entirely

trivikr (approved):

kuhe (approved):

PR #667, opened 2023-01-11
feat: replace back re2 with regexp

eduardomourar:
*Issue #, if available:* N/A

*Description of changes:*

This reverts the re2/re2-wasm change (related to #451 and #467) and puts back the default RegExp implementation. Here are the reasons:

* re2: these regular expressions are provided by the teams that designed the API and its Smithy model, so it would make more sense to move the check for vulnerable regular expressions during the Smithy build phase.

* re2-wasm: it adds an extra step during compilation which makes the server deployment more complex (as shown in the AWS sample [here](https://github.com/aws-samples/smithy-server-generator-typescript-sample/blob/7e751f0bef5bf29d08175fb4ff0a35c07eebc327/server/lib/cdk-stack.ts#L38-L40)).

### DIFF
--- a/smithy-typescript-ssdk-libs/server-common/package.json
+++ b/smithy-typescript-ssdk-libs/server-common/package.json
@@ -24,8 +24,7 @@
   "dependencies": {
     "@aws-sdk/protocol-http": "^3.127.0",
     "@aws-sdk/types": "^3.127.0",
-    "tslib": "^1.8.0",
-    "re2-wasm": "^1.0.2"
+    "tslib": "^1.8.0"
   },
   "devDependencies": {
     "@types/jest": "^26.0.22",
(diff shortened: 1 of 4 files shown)


zaccharles (approved):

trivikr (approved):

gosar:
We have not found a library that is guaranteed to _exhaustively_  catch vulnerable regular expressions in a model during a build phase, i.e., without false negatives. I don't think it is worth taking on availability risk for services, to remove this complexity in server setup.

eduardomourar:
@gosar, to really make it usable for us, we would appreciate if you could look into those 2 additional PRs, please?
- https://github.com/awslabs/smithy-typescript/pull/678
- https://github.com/awslabs/smithy-typescript/pull/685

PR #695, opened 2023-02-16
Remove reflected values from validation message

srchase:
This PR updates the `generateValidationMessage` to not return the input value in the validation message.

This change will be paired with updated protocol tests: https://github.com/awslabs/smithy/pull/1622

### DIFF
--- a/smithy-typescript-ssdk-libs/server-common/src/validation/index.ts
+++ b/smithy-typescript-ssdk-libs/server-common/src/validation/index.ts
@@ -103,20 +103,6 @@ export const generateValidationSummary = (failures: readonly ValidationFailure[]
 };
 
 export const generateValidationMessage = (failure: ValidationFailure): string => {
-  let failureValue;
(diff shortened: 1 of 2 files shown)


sugmanue (approved):

AndrewFossAWS (approved):

smithy-typescript-ssdk-libs/server-common/src/validation/index.ts
```diff
@@ -170,5 +153,5 @@ export const generateValidationMessage = (failure: ValidationFailure): string =>
       suffix = "must have unique values";
```
gosar, line 153: looks like we are missing tests for this case. can you add some?

srchase: Added here, and do the protocol tests in https://github.com/awslabs/smithy/pull/1622.

gosar (approved):
Ok to merge this now and update when the protocol test is updated.

PR #711, opened 2023-03-06
Add tests for convertRequest (NodeJS sSDK runtime).

pose:
*Description of changes:* Adding tests for `convertRequest` method for `server-node` runtime.

### DIFF
--- a/smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
+++ b/smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
@@ -1,0 +1,133 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { mkdtemp } from "fs/promises";
+import { createServer, IncomingMessage, request, RequestOptions, Server, ServerResponse } from "http";
+import * as os from "os";
+import * as path from "path";
+import { Readable } from "stream";
+
+import { convertRequest } from "./node";
+
+let socketPath: string;
+let promiseResolve: ([req, res]: [IncomingMessage, ServerResponse]) => void;
+
+let server: Server;
+beforeAll(async () => {
+  server = createServer(function (req, res) {
+    promiseResolve([req, res]);
+    resToEnd = res;
+  });
+  // Create a temporary named pipe where to run the server and obtain a request
+  socketPath = path.join(await mkdtemp(path.join(os.tmpdir(), "named-pipe-for-test-")), "server");
+  // TODO Add support to Windows by using '\\\\?\\pipe'
+  // See: https://nodejs.org/api/net.html#identifying-paths-for-ipc-connections
+  server.listen(socketPath);
+});
+
+let resToEnd: ServerResponse;
+
+function getRequest(options: RequestOptions & { body?: String }): Promise<[IncomingMessage, ServerResponse]> {
+  return new Promise((resolve, _) => {
+    promiseResolve = resolve;
+    request({
+      socketPath,
+      ...options,
+    }).end(Buffer.from(options.body || []));
+  });
+}
+
+afterAll(() => {
+  server?.close();
+});
+
+async function streamToString(stream: Readable) {
+  const chunks = [];
+
+  for await (const chunk of stream) {
+    chunks.push(Buffer.from(chunk));
+  }
+
+  return Buffer.concat(chunks).toString("utf-8");
+}
+
+describe("convertRequest", () => {
+  afterEach(async () => {
+    resToEnd?.end();
+  });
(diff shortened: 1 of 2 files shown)


smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    resToEnd = res;
+  });
+  // Create a temporary named pipe where to run the server and obtain a request
+  socketPath = path.join(await mkdtemp(path.join(os.tmpdir(), "named-pipe-for-test-")), "server");
+  // TODO Add support to Windows by using '\\\\?\\pipe'
+  // See: https://nodejs.org/api/net.html#identifying-paths-for-ipc-connections
+  server.listen(socketPath);
+});
+
+let resToEnd: ServerResponse;
```
kuhe, line 30: what does this variable name mean?

kuhe (approved):

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+let promiseResolve: ([req, res]: [IncomingMessage, ServerResponse]) => void;
+
+let server: Server;
+beforeAll(async () => {
+  server = createServer(function (req, res) {
+    promiseResolve([req, res]);
+    resToEnd = res;
+  });
+  // Create a temporary named pipe where to run the server and obtain a request
+  socketPath = path.join(await mkdtemp(path.join(os.tmpdir(), "named-pipe-for-test-")), "server");
```
everett1992, line 24: Why listen to a socket rather than a port? port 0 would bind to any free port.

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      ...options,
+    }).end(Buffer.from(options.body || []));
+  });
+}
+
+afterAll(() => {
+  server?.close();
+});
+
+async function streamToString(stream: Readable) {
```
everett1992, line 46: @aws-sdk/node-http-handler streamCollector?

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  // Create a temporary named pipe where to run the server and obtain a request
+  socketPath = path.join(await mkdtemp(path.join(os.tmpdir(), "named-pipe-for-test-")), "server");
+  // TODO Add support to Windows by using '\\\\?\\pipe'
+  // See: https://nodejs.org/api/net.html#identifying-paths-for-ipc-connections
+  server.listen(socketPath);
+});
+
+let resToEnd: ServerResponse;
+
+function getRequest(options: RequestOptions & { body?: String }): Promise<[IncomingMessage, ServerResponse]> {
```
everett1992, line 32: Starting a server is really cheap, you could kill all the beforeAll/beforeEach calls and move them into getRequest

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
pose: Thanks for the suggestion. I see the following caveats:
* It returns a `Uint8Array` which I still need to convert to String.
* I'm introducing a new dependency `@aws-sdk/node-http-handler` only for this.

@kuhe what do you think?

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
pose: I think that you are right: I'll give it a go.

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
pose: I'm not a huge fan of tests that depend on the TCP stack in general (probably since I have had clash with already used ports many times). It's also true that I can bind to the socket (as you describe) and it's probably going to be still quite fast. Do you think I'm losing performance or how realistic is the scenario under test?

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
pose: What I tried to convey is that when the test ends, it should end this response. I guess I could use a better name and try what @everett1992 suggested on moving this logic inside the `getRequest` method.

smithy-typescript-ssdk-libs/server-node/src/index.spec.ts
pose: FYI, I'll address this as the second PR for response writing.

srchase (approved):

PR #713, opened 2023-03-08
Remove internal enum values from validation message

srchase:
The enum validation protocol tests were updated in https://github.com/awslabs/smithy/pull/1658 to remove enum values with the `@internal` trait, or which have been tagged with `internal`.

For example:
```
enum Fruit {
    APPLE
    BANANA
    
    @internal
    DURIAN

   @tags(["internal"])
   PAPAYA
}
```

Validation for the above `Fruit` enum will check that input is any of the four enum values, but the validation message returned will only include the non-internal values, `apple` and `banana`.

When https://github.com/aws/aws-sdk-js-v3 updates to use the smithy release that contains https://github.com/awslabs/smithy/pull/1658, a PR will be cut to update the generated validation tests.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
@@ -35,8 +35,10 @@ import software.amazon.smithy.model.shapes.SimpleShape;
 import software.amazon.smithy.model.shapes.StringShape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.model.traits.EnumValueTrait;
 import software.amazon.smithy.model.traits.ErrorTrait;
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/StructuredMemberWriter.java
```diff
@@ -503,6 +505,22 @@ private void writeShapeValidator(TypeScriptWriter writer,
                         });
                     }
 
+                    if (shape.isEnumShape()) {
```
gosar, line 508: In smithy-typescript, the code generator does not convert all enum traits to enum shapes, so we have to check the EnumTrait like before.

srchase: Fixed in [dc90a62](https://github.com/awslabs/smithy-typescript/pull/713/commits/dc90a62abc4bea71c11032154ded1196f27c5a3b).

gosar (approved):

PR #714, opened 2023-03-08
model bucketing: special case for smithy.api#Unit

kuhe:
In an edge case, the only model assigned to the last bucket may be `smithy.api#Unit`. 

This causes the code generator to think that `models_N` exists, but in fact only `models_(N-1)` exists because `Unit` does not generate anything. This causes a compilation error in the generated models index.

```ts
export * from "./models_0";
export * from "./models_1"; // file doesn't exist
```

This proposed fix is to ignore `smithy.api#Unit` when incrementing the bucketing counter.

## testing
tested code generation against the edge case unreleased feature

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
@@ -67,6 +67,7 @@ import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.MediaTypeTrait;
 import software.amazon.smithy.model.traits.StreamingTrait;
+import software.amazon.smithy.model.traits.UnitTypeTrait;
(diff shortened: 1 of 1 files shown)


trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SymbolVisitor.java
```diff
@@ -458,13 +458,20 @@ public String formatModuleName(Shape shape, String name) {
...
             }
             // Add models into buckets no bigger than chunk size.
-            String path = String.join("/", ".", SHAPE_NAMESPACE_PREFIX, "models_" + bucketCount);
-            visitedModels.put(shape, path);
-            currentBucketSize++;
-            if (currentBucketSize == chunkSize) {
-                bucketCount++;
-                currentBucketSize = 0;
+            String path;
+            if (shape.getId().toString().equals("smithy.api#Unit")) {
```
kstich, line 462: Can use `.getId().equals(UnitTypeTrait.UNIT)` instead.

kstich (approved):

PR #722, opened 2023-03-17
Skip code generation of data filter functions for shapes that do not contain sensitive fields (recursive)

kuhe:
Adds a recursive check for sensitive fields before generating a sensitive data filter function. 
Currently all input/output have their own function even if it does nothing. This removes the no-op functions to reduce the code size.

For diff in JSv3 as an example, see https://github.com/aws/aws-sdk-js-v3/pull/4544 
(first commit contains a single client for easier viewing).

- [x] manual testing
- [x] what is the effect on code generation time?
  - for AWS EC2, generation took 4s to 5s both before and after this change.
- [x] use https://smithy.io/2.0/spec/selectors.html#forward-recursive-neighbors like `~> [trait|sensitive]`, look for `Selector.parse` (suggestion from @kstich)

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -46,6 +46,7 @@ import software.amazon.smithy.typescript.codegen.endpointsV2.EndpointsParamNameM
 import software.amazon.smithy.typescript.codegen.endpointsV2.RuleSetParameterFinder;
 import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator;
 import software.amazon.smithy.typescript.codegen.integration.RuntimeClientPlugin;
+import software.amazon.smithy.typescript.codegen.validation.SensitiveDataFinder;
 import software.amazon.smithy.utils.OptionalUtils;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
@@ -73,6 +74,7 @@ final class CommandGenerator implements Runnable {
     private final Symbol outputType;
     private final ProtocolGenerator protocolGenerator;
     private final ApplicationProtocol applicationProtocol;
+    private final SensitiveDataFinder sensitiveDataFinder = new SensitiveDataFinder();
 
     CommandGenerator(
             TypeScriptSettings settings,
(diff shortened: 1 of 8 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -313,20 +315,25 @@ private void generateCommandMiddlewareResolver(String configType) {
...
+                            }
                         },
-                        () -> writer.writeInline("(input: any) => input"));
+                        () -> writer.writeInline("(_: any) => _"));
                 });
                 writer.openBlock("outputFilterSensitiveLog: ", ",", () -> {
                     OptionalUtils.ifPresentOrElse(operationIndex.getOutput(operation),
                         output -> {
+                            if (sensitiveDataFinder.findsSensitiveData(output, model)) {
                             Symbol outputSymbol = symbolProvider.toSymbol(output);
```
trivikr, line 337: nit: need indentation

trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/validation/SensitiveDataFinder.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public class SensitiveDataFinder {
+    private Map<Shape, Boolean> cache = new HashMap<>();
+
+    /**
+     * @param shape - the shape in question.
+     * @param model - model context for the shape, containing its related shapes.
+     * @return whether a sensitive field exists in the shape and its downstream shapes.
+     */
+    public boolean findsSensitiveData(Shape shape, Model model) {
```
srchase, line 32: This method name is a little confusing. Maybe `containsSensitiveData` instead?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/validation/SensitiveDataFinder.java
```diff
@@ -0,0 +1,75 @@
...
+import software.amazon.smithy.model.traits.SensitiveTrait;
+import software.amazon.smithy.model.traits.StreamingTrait;
+
+/**
+ * This validator tells you whether a shape contains sensitive data fields.
+ * This is used to decide whether a sensitive log filter function needs to be generated for
+ * a given shape.
+ */
+public class SensitiveDataFinder {
+    private Map<Shape, Boolean> cache = new HashMap<>();
```
srchase, line 25: How is this cache supposed to work? The public method takes in a model, which could change. I don't think this will work as expected, unless the same model is used every time.

kuhe: I changed the `model` to be a constructor parameter.

srchase (approved):

PR #726, opened 2023-03-24
Export enums as const

srchase:
Currently, enums as generated as the following:

```
export enum Suit {
  club = "club",
  diamond = "diamond",
  heart = "heart",
  spade = "spade",
}
```

This results in the following JavaScript:
```
var Suit;
(function (Suit) {
    Suit["club"] = "club";
    Suit["diamond"] = "diamond";
    Suit["heart"] = "heart";
    Suit["spade"] = "spade";
})(Suit = exports.Suit || (exports.Suit = {}));
```

With this PR, enums are now generated as the following:
```
export const Suit = {
  club: "club",
  diamond: "diamond",
  heart: "heart",
  spade: "spade",
}
```

Which results in this JavaScript:
```
exports.Suit = {
  club: "club",
  diamond: "diamond",
  heart: "heart",
  spade: "spade",
};
```

To make this change work, when enums are used as structure members, `keyof typeof` is inserted:
```
export interface Foo {
  suit?: Suit | string;
}
```

Is updated to the following:

```
export interface Foo {
  suit?: keyof typeof Suit | string;
}
```


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/EnumGenerator.java
```diff
@@ -35,10 +35,11 @@
  * <p>We will generate the following:
  *
  * <pre>{@code
- * export enum TypedYesNo {
+ * export const /* enum *\/ TypedYesNo {
```
syall, line 38: Should this be the following?

```suggestion
 * export const /* enum *\/ TypedYesNo  = {
```

srchase: I added the missing `=`, but the extra slash is needed, otherwise `*/` is interpreted as the close of the multi-line comment.

srchase:
> Are there any backwards compatibility concerns around this?

I added the type export to allow for using these generated consts as types.

kuhe (approved):

syall:
In the example, it shows a double export of the name `Suit`, and was wondering if this is valid?

```typescript
export const /*enum*/ Suit = { // <- Suit export 1
  club: "club",
  diamond: "diamond",
  heart: "heart",
  spade: "spade",
} as const;
export type Suit = typeof Suit[keyof typeof Suit]; // <- Suit export 2
```

kuhe (approved):

PR #728, opened 2023-03-27
improve documentation truncation behavior

kuhe:
You can see that some shapes get quite long, CLI example: https://docs.aws.amazon.com/cli/latest/reference/quicksight/update-template.html

JSv3 API example: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-quicksight/classes/updatetemplatecommand.html

----

This PR aims to improve the recently implemented structural hint doc generation.

- more aggressive truncation for repeated shapes

- truncated item shows the target shape name instead of its entire structure
  - the name appears earlier in the shape hint

example, original shape appearance with name tag and then subsequent repeated shape truncation
```
{
  myShape: { // MyShape
    field1: "STRING_VALUE",
    field2: "STRING_VALUE"
  },
  myShape2: "<MyShape>"
}
```

JSv3 diff in https://github.com/aws/aws-sdk-js-v3/pull/4585/files

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/documentation/StructureExampleGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/documentation/StructureExampleGenerator.java
@@ -18,6 +18,7 @@ import software.amazon.smithy.model.shapes.IntEnumShape;
 import software.amazon.smithy.model.shapes.ListShape;
 import software.amazon.smithy.model.shapes.MapShape;
 import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.SetShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.UnionShape;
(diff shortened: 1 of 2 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/documentation/StructureExampleGenerator.java
```diff
@@ -261,23 +263,41 @@ private static void append(int indentation, StringBuilder buffer, String tail) {
...
+            occurrences.put(shape, occurrences.getOrDefault(shape, 0) + 1);
+        }
+
+        /**
+         * @return whether the shape should be truncated.
+         */
+        public boolean shouldTruncate(Shape shape) {
+            return (shape instanceof MapShape || shape instanceof UnionShape || shape instanceof StructureShape
+                    || shape instanceof ListShape || shape instanceof SetShape)
+                    && (getOccurrenceCount(shape) > 2 || getOccurrenceDepths(shape) > 2);
```
kuhe, line 286: technically occurrence count is greater or equal to depth count, but I left it there in case it needs further adjustments.

syall:
Could test cases be added for both truncating on occurrences and depth?

syall (approved):

PR #730, opened 2023-04-03
shorten internal serde fn names

kuhe:
Shorten generated serde function names. Source code contains a comment with the original long name, distribution uses shorter names.

The premise is that serde function names do not need to contain the protocol -- it's repetitive when the protocol file containing the functions is named after the protocol already. 

JSv3 codegen paired PR https://github.com/aws/aws-sdk-js-v3/pull/4611

Benefits:
reduced code size

Potential Downsides:
- breaking change for anyone is using these internal functions directly (though no one should be). 
- no benefit to anyone who was applying mangling already (most browser applications, some server applications).

Example
```js
/**
 * deserializeAws_restJson1PrefetchConsumption
 */
const de_PrefetchConsumption = ...
```

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/ProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/ProtocolGenerator.java
@@ -114,8 +114,8 @@ public interface ProtocolGenerator {
                     .collect(Collectors.joining(", "));
             throw new CodegenException(String.format(
                     "All of the protocols generated for a service must be runtime compatible, but "
-                    + "protocol `%s` is incompatible with other application protocols: [%s]. Please pick a "
-                    + "set of compatible protocols using the `protocols` option when generating %s.",
+                            + "protocol `%s` is incompatible with other application protocols: [%s]. Please pick a "
+                            + "set of compatible protocols using the `protocols` option when generating %s.",
                     getProtocol().getName(), protocolNames, service.getId()));
         }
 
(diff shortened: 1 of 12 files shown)


kuhe:
Update: no longer shortening the symbol/shape name to prevent thrashing on the function names. Many names

pose (changes requested):
Why is this required and cannot be achieved by properly configuring bundling and minifying with esbuild/webpack?

kuhe:
This is not required, but it benefits users who use the generated clients as they are without minifying. One example is the AWS Lambda provided AWS SDK for JavaScript v3 that users do not bundle into their AWS Lambda applications.

mtdowling:
Can you update the description to show the current state of what this generates? I'd like these names to be stable cross model updates too, which I think you already added. Is it now `[se|de]_ShapeName`?

pose:
Why does the AWS Lambda provided AWS SDK for JavaScript v3 cannot be minified?

kuhe:
Yes, the PR currently is only a change to `[se|de]_ShapeName` which is stable over model changes. I've updated the original description.

> Why does the AWS Lambda provided AWS SDK for JavaScript v3 cannot be minified?

The AWS Lambda provided AWS SDKs are supposedly for development/testing/convenience purposes, [link](https://docs.aws.amazon.com/lambda/latest/operatorguide/sdks-functions.html) and should retain readable stack traces without mangling and minification, but customers often prefer to continue to use them for production workloads. 

We can provide some benefit to those users without needing to change customer behavior. However, if you think the risk of this change is greater than the potential benefits, I'm fine with closing it.

pose (approved):

trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/ProtocolGenerator.java
```diff
@@ -271,7 +292,8 @@ static String getSerdeFunctionSymbolComponent(Symbol symbol, Shape shape) {
      * @param operation the operation shape to retrieve errors for
      * @return map of error names to {@link ShapeId}
      */
-    default Map<String, ShapeId> getOperationErrors(GenerationContext context, OperationShape operation) {
+    default Map<String, ShapeId> getOperationErrors(GenerationContext context,
```
mtdowling, line 295: nit - we wrap at 120

srchase:
Can all the auto-formatting be undone on this? It's hard to read the actual changes and this will be an unnecessarily noisy commit.

srchase (approved):

PR #736, opened 2023-04-17
use aggregated client runtime generator

kuhe:
Switch to using an aggregated client runtime generator instead of codegen.


trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServiceAggregatedClientGenerator.java
```diff
@@ -62,16 +63,24 @@ final class ServiceAggregatedClientGenerator implements Runnable {
         this.aggregateClientName = aggregateClientName;
         this.applicationProtocol = applicationProtocol;
         serviceSymbol = symbolProvider.toSymbol(service);
+        this.barebonesClientName = serviceSymbol.getName();
```
milesziemer, line 66: Nit: this could be scoped to the `run()` method

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServiceAggregatedClientGenerator.java
```diff
@@ -83,43 +92,37 @@ public void run() {
...
-                                 + "  return this.send(command, optionsOrCb);\n"
-                                 + "}", operationSymbol);
-                });
                 writer.write("");
             }
         });
+
+        writer.write("");
+
+        writer.addImport(barebonesClientName + "Config", null, serviceSymbol.getNamespace());
```
milesziemer, line 117: Maybe we should add another method in ServiceBareBonesClientGenerator for getting the config type name, like https://github.com/awslabs/smithy-typescript/blob/a526a73f8fada65c9ef0bddea94a85695229f2c4/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/ServiceBareBonesClientGenerator.java#L90

milesziemer (approved):

PR #740, opened 2023-04-19
TypeScript packaging to top level

srchase:
This PR moves package.json and the associated config files (tsconfig, eslint, prettier) for TypeScript to the top-level.

This PR separates the out changes from https://github.com/awslabs/smithy-typescript/pull/737, allowing for incremental review and steps toward migrating packages to the `/packages` directory.


smithy-typescript-ssdk-libs/server-apigateway/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aws-smithy/server-apigateway",
-  "version": "1.0.0-alpha.10",
```
pose, line 3: Could we keep the `-alpha` label across the packages? Thanks!

pose (changes requested):

pose (approved):

PR #749, opened 2023-04-27
yarn pack before releasing

srchase:
This PR adds a `stage-release` script which is run as part of the `release` script.

Before `yarn changeset publish` can be invoked by the release script, yarn needs to handle converting the `workspace:^` entries in each package's package.json into a real version number. The `stage-release` script invokes `yarn pack` which performs this conversion while packing the contents into a `package.tgz` file. The contents of the packed package are then unpacked into a `.release` directory, which is used for publishing. Turbo will assure that `yarn build` is successfully run before attempting to stage the release contents to `.release`.

### DIFF
--- a/packages/protocol-http/package.json
+++ b/packages/protocol-http/package.json
@@ -7,6 +7,7 @@
     "build:es": "tsc -p tsconfig.es.json",
     "build:types": "tsc -p tsconfig.types.json",
     "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
+    "stage-release": "rimraf ./_release && yarn pack && mkdir ./_release && tar zxvf ./package.tgz --directory ./_release && rm ./package.tgz",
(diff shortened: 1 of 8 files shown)


packages/protocol-http/package.json
```diff
@@ -54,5 +55,8 @@
   },
   "typedoc": {
     "entryPoint": "src/index.ts"
+  },
+  "publishConfig": {
+    "directory": "_release/package"
```
hpmellema, line 60: nit: Should a private directory be `.release/` rather than `_`?

hpmellema (approved):

packages/protocol-http/package.json
srchase: Changed to .release in [fe6a59d](https://github.com/awslabs/smithy-typescript/pull/749/commits/fe6a59dfbe9286b7d38f3df5461921d70ac01ddd)

PR #757, opened 2023-05-01
docs: response structure and base service exception

MYoung25:
*Description of changes:*
Add response structure and base service exception to commands

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -180,7 +180,10 @@ final class CommandGenerator implements Runnable {
                                 model.getShape(operation.getInputShape()).get(), model))
                 + String.format("const command = new %s(input);%n", commandName)
                 + "const response = await client.send(command);\n"
-                + "```\n"
+                + String.format("/**%n%s%n",
+                        StructureExampleGenerator.generateStructuralHintDocumentation(
+                                model.getShape(operation.getOutputShape()).get(), model))
+                + "\n```\n"
                 + "\n"
                 + String.format("@param %s - {@link %s}%n", commandInput, commandInput)
                 + String.format("@returns {@link %s}%n", commandOutput)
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -180,7 +180,10 @@ private String getCommandExample(String serviceName, String configName, String c
                                 model.getShape(operation.getInputShape()).get(), model))
                 + String.format("const command = new %s(input);%n", commandName)
                 + "const response = await client.send(command);\n"
-                + "```\n"
+                + String.format("/**%n%s%n",
```
kuhe, line 183: This generates a comment block opening indicator `/**`, right? Does it need to be escaped, since it's an example block inside what is already a comment block?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -180,7 +180,10 @@ private String getCommandExample(String serviceName, String configName, String c
...
+                        StructureExampleGenerator.generateStructuralHintDocumentation(
+                                model.getShape(operation.getOutputShape()).get(), model))
+                + "\n```\n"
```
kuhe, line 186: here, where the comment within an example within a comment ends, don't we need an escaped comment block ending indicator? 

Or, rather than doing that, how about making the output shape an assigned variable?
```ts
const responseStructure = { ... };
```

This gives the output syntax highlighting as opposed to
```ts
/**
{ ... }
*/
```

which will be in comment shade.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
MYoung25: The opening indicator doesn't need to be escaped, but the closing indicator would, which is why I left it out.  TSDoc's extractor reads this implementation correctly as the example is closed by "```"

kuhe (approved):

kuhe (approved):

srchase (approved):

PR #763, opened 2023-05-12
add packageDocumentation and fix i/o documentation inconsistencies

MYoung25:
*Description of changes:*
Add `@packageDocumentation` trait to client root.
Change command `@params` and `@returns` to more accurately and consistently reflect the implementation

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -185,8 +185,12 @@ final class CommandGenerator implements Runnable {
                                 model.getShape(operation.getOutputShape()).get(), model, true))
                 + "\n```\n"
                 + "\n"
-                + String.format("@param %s - {@link %s}%n", commandInput, commandInput)
-                + String.format("@returns {@link %s}%n", commandOutput)
+                + String.format("@param %s - {@link %s}%n",
+                    model.getShape(operation.getInputShape()).get().getId().getName(),
+                    model.getShape(operation.getInputShape()).get().getId().getName())
+                + String.format(
+                    "@returns {@link %s}%n",
+                    model.getShape(operation.getOutputShape()).get().getId().getName())
                 + String.format("@see {@link %s} for command's `input` shape.%n", commandInput)
                 + String.format("@see {@link %s} for command's `response` shape.%n", commandOutput)
                 + String.format("@see {@link %s | config} for %s's `config` shape.%n", configName, serviceName);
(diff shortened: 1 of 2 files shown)


kuhe (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/IndexGenerator.java
```diff
@@ -45,6 +47,16 @@ static void writeIndex(
         ProtocolGenerator protocolGenerator,
         TypeScriptWriter writer
     ) {
+        ServiceShape service = settings.getService(model);
+        Optional<String> docs = service.getTrait(DocumentationTrait.class)
+                .map(DocumentationTrait::getValue);
+        writer.write("/* eslint-disable */");
+        docs.ifPresent(documentationTrait -> writer.writeDocs(
```
srchase, line 54: nit: since the service and its docs aren't used again, I might do this instead:
```
writer.write("/* eslint-disable */");
settings.getService(model).getTrait(DocumentationTrait.class).ifPresent(trait ->
    writer.writeDocs(trait.getValue() + "\n\n" + "@packageDocumentation"));
```
And then you can leave out the `Optional` import.

srchase (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -185,8 +185,12 @@ private String getCommandExample(String serviceName, String configName, String c
                                 model.getShape(operation.getOutputShape()).get(), model, true))
                 + "\n```\n"
                 + "\n"
-                + String.format("@param %s - {@link %s}%n", commandInput, commandInput)
-                + String.format("@returns {@link %s}%n", commandOutput)
+                + String.format("@param %s - {@link %s}%n",
+                    model.getShape(operation.getInputShape()).get().getId().getName(),
```
kuhe, line 189: The name of the input is `input` in the constructor and `args` in the aggregated client methods. However, no need to change this since in both cases there is only 1 positional param.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -185,8 +185,12 @@ private String getCommandExample(String serviceName, String configName, String c
...
+                    model.getShape(operation.getInputShape()).get().getId().getName())
+                + String.format(
+                    "@returns {@link %s}%n",
+                    model.getShape(operation.getOutputShape()).get().getId().getName())
```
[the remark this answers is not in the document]
kuhe: can you make this e.g.
```
@returns {@link ApplyArchiveRuleCommandOutput} when sent with {@link AccessAnalyzerClient#send}.
```

Do method links with unimported symbols (the Client) work in the docs app? I'm guessing at the method link syntax, but it does exist in jsdoc/tsdoc.

PR #777, opened 2023-06-13
widen blob payload input types, add blob payload output converter

kuhe:
https://github.com/aws/aws-sdk-js-v3/issues/4786
https://github.com/aws/aws-sdk-js-v3/issues/4265

Blob types enhancements

- on the client input side, operation input members with type blob and trait HttpPayload have been widened from `Uint8Array` to `BlobTypes`, which contains common and runtime specific types.
 ```ts
export type BlobTypes = string | ArrayBuffer | ArrayBufferView | Uint8Array | RuntimeBlobTypes;

// example Node.js runtime additions
export type RuntimeBlobTypes = Readable | Buffer;
// browser additions
export type RuntimeBlobTypes = Blob | ReadableStream;
```

- on the client output side, operation output members with type blob and trait HttpPayload have been wrapped in `Uint8ArrayBlobAdapter`, which extends the `Uint8Array` native class and adds a convenient platform-specific way to convert to string. Example in the linked JSv3 PR https://github.com/aws/aws-sdk-js-v3/pull/4836.

----

This applies in general, but the user story is difficulty/inconvenience around non-streaming blob i/o, e.g. in Lambda invoke. Non-streaming blobs were treated uniformly as Uint8Array in input and output, but this does not cover the realistic use cases of _payload_ blobs.

```js
const lambda = new Lambda({
  region: "us-west-2",
});

const invoke = lambda.invoke({
  FunctionName: "echo",
  Payload: JSON.stringify({
    hello: "world",
  }), // without input widening, this would be a type error since string is not Uint8Array.
});

const payload = (await invoke).Payload;

console.log(JSON.parse(payload?.transformToString() ?? "{}")); // adapter adds the string conversion method, preserves the type extending Uint8Array.
```

- [x] JSv3 will write an e2e integration test suite for this.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenUtils.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenUtils.java
@@ -29,6 +29,7 @@ import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.traits.HttpPayloadTrait;
 import software.amazon.smithy.model.traits.StreamingTrait;
 import software.amazon.smithy.typescript.codegen.integration.AddSdkStreamMixinDependency;
 import software.amazon.smithy.utils.SmithyUnstableApi;
@@ -135,10 +136,11 @@ public final class CodegenUtils {
      * Refer here for more rationales: https://github.com/aws/aws-sdk-js-v3/issues/843
      */
     static void writeClientCommandStreamingInputType(
-            TypeScriptWriter writer,
-            Symbol containerSymbol,
-            String typeName,
-            MemberShape streamingMember
+        TypeScriptWriter writer,
+        Symbol containerSymbol,
+        String typeName,
+        MemberShape streamingMember,
+        String commandName
     ) {
         String memberName = streamingMember.getMemberName();
         String optionalSuffix = streamingMember.isRequired() ? "" : "?";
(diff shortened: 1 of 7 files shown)


trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenUtils.java
```diff
@@ -180,6 +182,73 @@ static void writeClientCommandStreamingOutputType(
...
+
+        writer.writeDocs("@public");
+        writer.openBlock("export type $LType = Omit<$T, $S> & {", "};",
+            typeName,
+            containerSymbol,
+            memberName,
+            () -> {
+                writer.write("$1L$2L: BlobTypes;", memberName, optionalSuffix);
+            }
+        );
```
milesziemer, line 216: Nit: Use text block

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenUtils.java
```diff
@@ -180,6 +182,73 @@ static void writeClientCommandStreamingOutputType(
...
+
+        writer.writeDocs("@public");
+        writer.openBlock("export type $LType = Omit<$T, $S> & {", "};",
+            typeName,
+            containerSymbol,
+            memberName,
+            () -> {
+                writer.write("$1L$2L: Uint8ArrayBlobAdapter;", memberName, optionalSuffix);
+            }
+        );
```
milesziemer, line 243: Same as https://github.com/awslabs/smithy-typescript/pull/777/files#r1235363220

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -248,25 +248,14 @@ static void generateMetadataDeserializer(GenerationContext context, SymbolRefere
     }
 
     /**
-     * Writes a response body stream collector. This function converts the low-level response body stream to
+     * Imports a response body stream collector. This function converts the low-level response body stream to
      * Uint8Array binary data.
      *
      * @param context The generation context.
      */
     static void generateCollectBody(GenerationContext context) {
```
milesziemer, line 256: Do we need this whole method if it just imports something? Or maybe change the name?

kuhe: I rolled it into the generateCollectBodyString method, which is called in all places that generateCollectBody was called.

milesziemer (approved):

PR #791, opened 2023-06-22
Update codegen for package migration

srchase:
This PR updates codegen to use the `@smithy` NPM packages that were added in https://github.com/awslabs/smithy-typescript/pull/773 and https://github.com/awslabs/smithy-typescript/pull/790.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
@@ -39,82 +39,82 @@ import software.amazon.smithy.utils.SmithyUnstableApi;
 @SmithyUnstableApi
 public enum TypeScriptDependency implements SymbolDependencyContainer {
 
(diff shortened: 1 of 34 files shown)


kuhe (approved):

AndrewFossAWS (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptDependency.java
milesziemer, line 1: Some of these are still named AWS, even though they're presumably generic client dependencies (ie. smithy scope). Can we change them or is a reason to keep it like that?

milesziemer (commented):
Some of the `writer.addImport`s are updated to use the `TypeScriptDependency` param instead of package name, where others that still are using the package name have just changed the name to use the smithy scope [for example](https://github.com/awslabs/smithy-typescript/pull/791/files#diff-8626d38d400d3c5b8a1f1d5becda65553b9e5f81daf36b26c8ae7100167db01fR239). Can we update those too?

PR #792, opened 2023-06-23
feat(external-interfaces): add external package control

kuhe:
This creates a package called `@smithy/external-interfaces` designed to be used with the `^` version range by generated SDKs. 

This allows the deployment of an external dependency patch, e.g. fast-xml-parser 4.2.4 to 4.2.5 without publishing new versions of the SDKs.


packages/external-interfaces/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@smithy/external-interfaces",
+  "version": "1.0.0",
```
kuhe, line 3: ⚠️ this package must be published immediately upon merge of this PR so downstream generated SDKs can start consuming it.

kuhe: PR could also be split into 

1 - the package to be published
2 - the codegen change

packages/external-interfaces/README.md
```diff
@@ -0,0 +1,10 @@
+# @smithy/external-interfaces
+
+[![NPM version](https://img.shields.io/npm/v/@smithy/external-interfaces/latest.svg)](https://www.npmjs.com/package/@smithy/external-interfaces)
+[![NPM downloads](https://img.shields.io/npm/dm/@smithy/external-interfaces.svg)](https://www.npmjs.com/package/@smithy/external-interfaces)
+
+This package provides a control for external interfaces.
+This package is installed to generated clients with the `^` version range, allowing updates
+of this package to deploy patches of external packages without publishing incremental versions of those packages.
+
+It is a hot patch-in-place mechanism.
```
kuhe, line 10: - [ ] add a remark here that any breaking api change to how external APIs within this package are used must involve a major version bump of `@smithy/external-interfaces` because it is generated as `^1.0.0` to the clients

PR #806, opened 2023-07-03
fix(codegen): replace mismatch version for server-common

eduardomourar:
*Issue #, if available:* N/A

*Description of changes:*

Replace non-existent version with `1.0.0-alpha.10` for `@aws-smithy/server-common` package.


gosar (approved):
Thanks @eduardomourar for catching this.

PR #812, opened 2023-07-06
Bump @smithy deps for downlevel types

srchase:
Updates @smithy dependencies after patch changes from https://github.com/awslabs/smithy-typescript/pull/810 and https://github.com/awslabs/smithy-typescript/pull/811


trivikr:
This PR is not required, as:
* All existing installations without lockfile will use latest 1.x version of smithy packages.
* Any new installations will use latest version 1.x version of smithy packages.
* Any existing installations with lockfiles will update to latest 1.x versions anyway, when lockfile is updated.

This PR is required, only when there's a major version bump for any smithy packages.

kuhe (approved):

PR #815, opened 2023-07-06
exposing provider errors, since they should be a part of our public API

RanVaknin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/4907
https://github.com/aws/aws-sdk-js-v3/issues/4918

*Description of changes:*
Currently, if an error occurs within a provider and it throws an exception that is marked "internal", customers might not have a clear and stable way to identify and handle these exceptions. This can lead to challenges in ensuring that their applications behave correctly when encountering errors during the credential providing process.

Because of this, provider errors should be moved out of the "internal" state, and offered as part of our public API.

---

### DIFF
--- a/packages/property-provider/src/CredentialsProviderError.ts
+++ b/packages/property-provider/src/CredentialsProviderError.ts
@@ -1,7 +1,6 @@
 import { ProviderError } from "./ProviderError";
 
 /**
- * @internal
  *
  * An error representing a failure of an individual credential provider.
  *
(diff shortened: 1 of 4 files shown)


srchase (approved):

packages/property-provider/src/CredentialsProviderError.ts
```diff
@@ -1,7 +1,6 @@
 import { ProviderError } from "./ProviderError";
 
 /**
- * @internal
```
kuhe, line 4: this should say `@public`

PR #825, opened 2023-07-12
Add SSDK codegen test

srchase:
This PR adds testing of the `typescript-ssdk-codegen` plugin.

It adds the non-published `software.amazon.smithy.typescript.ssdk.codegen.test.utils` package, which includes an empty `FakeProtocolGenerator` which enables SSDK codegen to succeed since a protocol is required for an SSDK.

Currently, SSDK validation for `@streaming` union shapes is not supported, so the `GetCityAnnouncements` is removed from the generated SSDK since it uses the `@streaming` union `Announcements` shape in its output. Once support for this is added, the operation can be added back to the model.

---

### DIFF
--- a/smithy-typescript-ssdk-codegen-test-utils/build.gradle.kts
+++ b/smithy-typescript-ssdk-codegen-test-utils/build.gradle.kts
@@ -1,0 +1,32 @@
+extra["displayName"] = "Smithy :: Typescript :: SSDK :: Codegen :: Test :: Utils"
+extra["moduleName"] = "software.amazon.smithy.typescript.ssdk.codegen.test.utils"
+
+val smithyVersion: String by project
+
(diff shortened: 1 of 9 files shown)


smithy-typescript-ssdk-codegen-test-utils/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+dependencies {
+    implementation(project(":smithy-typescript-codegen"))
+    implementation("software.amazon.smithy:smithy-model:$smithyVersion")
+}
```
hpmellema, line 32: nit: add newlines

hpmellema (approved):

Issue #848, opened 2023-07-26
InvalidSignatureException occurs with some special characters in query string with multiple value parameters


ealain:
### Version

`@smithy/signature-4: 2.0.1`

### Current behavior

When signing a request that has a query parameter with two values such that the sort order is different before and after encoding, then the resulting signature is incorrect.

For example, `["a", "à"].sort()` is `["a", "à"]` while `["a", "à"].map(encodeURIComponent).sort()` is `["%C3%A0", "a"]`.

### How to reproduce

Sign a request with the following query string: `?filter=a&filter=à`.

### Expected behavior

A request with the aforementioned query string is correctly signed.

ealain:
Hello,

I have noticed this issue and I am considering opening a PR that fixes it.

I have also noticed that the recently deprecated `@aws-sdk/signature-v4` package has the same issue. Not sure if the fix should be applied there as well.

PR #849, opened 2023-07-26
fix: sort query parameter values after encoding

ealain:
This solves the issue #848.

As stated in the [AWS documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html#create-canonical-request) on signing API requests (emphasis in mine),
> `CanonicalQueryString` – The URL-encoded query string parameters, separated by ampersands (&). Percent-encode reserved characters, including the space character. Encode names and values separately. If there are empty parameters, append the equals sign to the parameter name before encoding. **After encoding, sort the parameters alphabetically** by key name. If there is no query string, use an empty string ("").

*Description of changes:*

- sort the parameter values after encoding in `getCanonicalQuery`,
- test the associated edge case.

---

### DIFF
--- a/packages/signature-v4/src/getCanonicalQuery.spec.ts
+++ b/packages/signature-v4/src/getCanonicalQuery.spec.ts
@@ -96,6 +96,17 @@ describe("getCanonicalQuery", () => {
     ).toBe("%F0%9F%90%8E=%F0%9F%92%A9&%F0%9F%90%8E=%F0%9F%A6%84");
   });
 
+  it("should sort URI-encode members of query param arrays", () => {
+    expect(
+      getCanonicalQuery(
+        new HttpRequest({
+          ...httpRequestOptions,
+          query: { "p": ["a", "à"] },
+        })
+      )
+    ).toBe("p=%C3%A0&p=a");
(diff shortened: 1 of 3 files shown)


trivikr (approved):

srchase:
Thanks raising the issue and making this contribution to fix it! Mind running `yarn format` on it?

packages/signature-v4/src/getCanonicalQuery.spec.ts
```diff
@@ -96,6 +96,17 @@ describe("getCanonicalQuery", () => {
     ).toBe("%F0%9F%90%8E=%F0%9F%92%A9&%F0%9F%90%8E=%F0%9F%A6%84");
   });
 
+  it("should sort URI-encode members of query param arrays", () => {
+    expect(
+      getCanonicalQuery(
+        new HttpRequest({
+          ...httpRequestOptions,
+          query: { "p": ["a", "à"] },
```
srchase, line 104: `yarn format` is only complaining about this:
```suggestion
          query: { p: ["a", "à"] },
```

PR #851, opened 2023-07-27
Integration tests

srchase:
This PR ports the integration tests from https://github.com/aws/aws-sdk-js-v3/tree/main/private/aws-middleware-test/src. The tests for middleware-endpoint are not included in this change, but will be added in a follow-up PR.

The private, `@smithy/util-test package` is added to provide the mocking test utility of `requireRequestsFrom`.
---


package.json
```diff
@@ -8,6 +8,7 @@
     "clean": "turbo run clean --force --parallel",
     "build": "turbo run build",
     "test": "turbo run test",
+    "test:integration": "turbo run test:integration",
```
srchase, line 11: The test Weather client needs generated and built before integration tests can be run. Should this be `./gradlew clean build && turbo run test:integration` instead? If someone is working on integration tests, they may or may not need to run codegen, depending on their changes. However, if you're working on a single dependency package's integration tests, you may opt to run `yarn test-integration` from the local package instead, to avoid running tests across all of the packages.

packages/middleware-apply-body-checksum/src/middleware-apply-body-checksum.integ.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { requireRequestsFrom } from "@smithy/util-test";
+import { Weather } from "weather";
```
srchase, line 2: nb: The test client is imported from the generated weather package, but it isn't listed in the dependency package's package.json as a devDependency since that would result in a circular dependency.

packages/middleware-apply-body-checksum/package.json
```diff
@@ -47,6 +48,7 @@
     "directory": "packages/middleware-apply-body-checksum"
   },
   "devDependencies": {
+    "@smithy/util-test": "workspace:^",
```
srchase, line 51: The private, `@smithy/util-test` package is added here so it will be built before tests are run.

Will this be confusing when these packages are published?

AndrewFossAWS (approved):

trivikr (approved):

PR #852, opened 2023-07-27
Add extensions to client runtime config

AndrewFossAWS:
*Description of changes:*
This PR adds `extensions: RuntimeExtension[]` to client runtime config.

```
interface S3ClientConfiguration extends DefaultClientConfiguration,  AwsClientConfiguration, ... {
}

interface RuntimeExtension {
    configureClient(clientConfiguration: S3ClientConfiguration): void;
}
```

The service client can be configured by implementing the `RuntimeExtension` interface. `*ClientConfiguration` is an interface that includes settings that can be modified on the service client. Because services have different configuration, each service has a code-generated ClientConfiguration interface implementation. Service owner can compose multiple client configurations by implementing `TypeScriptIntegration.getClientConfigurationInterfaces`

Using S3 as an example:

``` 
export class CustomChecksumExtension implements RuntimeExtension {
    configureClient(clientConfiguration: S3ClientConfiguration): void {
       clientConfiguration.addChecksumAlgorithm(...);
    }
}

S3Client({
  retryStrategy: new AdaptiveRetryStrategy(params),
  sha256: Hash.bind(null, "sha256")
  ...
  extensions: [CustomChecksumExtension()]
})
```

---

### DIFF
--- a/packages/types/src/extensions/checksum.ts
+++ b/packages/types/src/extensions/checksum.ts
@@ -1,0 +1,58 @@
+import {ChecksumConstructor} from "../checksum";
+import {HashConstructor} from "../crypto";
+import {DefaultClientConfiguration} from "./defaultClientConfiguration";
+
+/**
+ * @internal
+ */
+export interface ChecksumAlgorithm {
+    algorithmId(): string;
+    checksumConstructor(): ChecksumConstructor | HashConstructor;
+}
+
+/**
+ * @internal
+ */
+export const getChecksumClientConfiguration = (runtimeConfig: any) => {
+    const checksumAlgorithms: ChecksumAlgorithm[] = [];
+
+    if (runtimeConfig.sha256 !== undefined) {
+        checksumAlgorithms.push({
+            algorithmId: () => "sha256",
+            checksumConstructor: runtimeConfig.sha256
+        })
+    }
+
+    if (runtimeConfig.md5 !== undefined) {
+        checksumAlgorithms.push({
+            algorithmId: () => "md5",
+            checksumConstructor: runtimeConfig.md5
+        })
+    }
+
+    const clientConfiguration = {
+        _checksumAlgorithms: checksumAlgorithms,
+        addChecksumAlgorithm: function(algo: ChecksumAlgorithm): void {
+            this._checksumAlgorithms.push(algo);
+        },
+        checksumAlgorithms: function(): ChecksumAlgorithm[] {
+            return this._checksumAlgorithms;
+        }
+    }
+
+    return clientConfiguration;
+}
+
+/**
+ * @internal
+ */
+export const resolveChecksumRuntimeConfig = (clientConfig: DefaultClientConfiguration) => {
+    const runtimeConfig: any = {
+    };
+
+    clientConfig.checksumAlgorithms().forEach(checksumAlgorithm => {
+        runtimeConfig[checksumAlgorithm.algorithmId()] = checksumAlgorithm.checksumConstructor();
+    });
+
+    return runtimeConfig;
+}
--- a/packages/types/src/extensions/checksum.ts
+++ b/packages/types/src/extensions/checksum.ts
@@ -1,58 +1,86 @@
-import {ChecksumConstructor} from "../checksum";
-import {HashConstructor} from "../crypto";
-import {DefaultClientConfiguration} from "./defaultClientConfiguration";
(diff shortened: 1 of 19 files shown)


srchase (commented):
I'm not sure we need to do a MV bump for the @smithy/types change.

If we do a MV bump, we need to do all the packages together.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/Dependency.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
srchase, line 2: nit: short-form header without the year.

packages/types/src/extensions/checksum.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+export interface ChecksumAlgorithm {
+    algorithmId(): string;
+    checksumConstructor(): ChecksumConstructor | HashConstructor;
+}
+
+/**
+ * @internal
+ */
+export const getChecksumClientConfiguration = (runtimeConfig: any) => {
```
kuhe, line 16: this `any` looks like it should be 
```ts
Partial<{
  sha256: ChecksumConstructor | HashConstructor,
  md5: ChecksumConstructor | HashConstructor
}>
```

and the output is `ChecksumAlgorithm`?

packages/types/src/extensions/checksum.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * @internal
+ */
+export const resolveChecksumRuntimeConfig = (clientConfig: DefaultClientConfiguration) => {
+    const runtimeConfig: any = {
+    };
+
+    clientConfig.checksumAlgorithms().forEach(checksumAlgorithm => {
+        runtimeConfig[checksumAlgorithm.algorithmId()] = checksumAlgorithm.checksumConstructor();
```
kuhe, line 54: is the set of algorithmIds known ahead of time?

packages/types/src/extensions/checksum.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    return clientConfiguration;
+}
+
+/**
+ * @internal
+ */
+export const resolveChecksumRuntimeConfig = (clientConfig: DefaultClientConfiguration) => {
```
kuhe, line 49: this function should have a return type

packages/types/src/extensions/checksum.ts
```diff
@@ -0,0 +1,58 @@
...
+            checksumConstructor: runtimeConfig.md5
+        })
+    }
+
+    const clientConfiguration = {
+        _checksumAlgorithms: checksumAlgorithms,
+        addChecksumAlgorithm: function(algo: ChecksumAlgorithm): void {
+            this._checksumAlgorithms.push(algo);
+        },
+        checksumAlgorithms: function(): ChecksumAlgorithm[] {
```
kuhe, line 38: a function on an object should be in a method-like syntax

```ts
{
  addChecksumAlgorithm(): void { ... }
  checksumAlgorithms(): ChecksumAlgorithm[] { ... }
}
```

packages/types/src/extensions/checksum.ts
AndrewFossAWS: Yes, I have added an enum to predefined a set of algorithms

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeExtensionsGenerator.java
```diff
@@ -0,0 +1,96 @@
...
+        this.service = service;
+        this.symbolProvider = symbolProvider;
+        this.delegator = delegator;
+        this.integrations = integrations;
+    }
+
+    void generate() {
+        String clientName = symbolProvider.toSymbol(service).getName();
+
+        String template1Contents = TypeScriptUtils.loadResourceAsString(TEMPLATE_1)
```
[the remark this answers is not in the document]
AndrewFossAWS: Personally I find it easier to read the content in a seperate template file instead of text blocks. Keeping them as templates for now.

kuhe (approved):

srchase (approved):

PR #861, opened 2023-08-09
fix: pagination return type

kuhe:
fixes https://github.com/aws/aws-sdk-js-v3/issues/2610
the paginators generate with an undefined return type, not `T`.

supercedes https://github.com/aws/aws-sdk-js-v3/pull/2611
supercedes https://github.com/awslabs/smithy-typescript/pull/389

### DIFF
--- a/packages/types/src/pagination.ts
+++ b/packages/types/src/pagination.ts
@@ -5,7 +5,7 @@ import { Client } from "./client";
  *
  * Expected type definition of a paginator.
  */
-export type Paginator<T> = AsyncGenerator<T, T, unknown>;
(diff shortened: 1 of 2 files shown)


packages/types/src/pagination.ts
```diff
@@ -5,7 +5,7 @@ import { Client } from "./client";
  *
  * Expected type definition of a paginator.
  */
-export type Paginator<T> = AsyncGenerator<T, T, unknown>;
+export type Paginator<T> = AsyncGenerator<T, undefined, unknown>;
```
kuhe, line 8: this appears to be a breaking change, but this is a fix that corrects the type declaration

siddsriv (approved):

PR #867, opened 2023-08-11
reduce buffer copies

kuhe:
supercedes external contrib https://github.com/aws/aws-sdk-js-v3/pull/2612

### DIFF
--- a/packages/node-http-handler/src/write-request-body.ts
+++ b/packages/node-http-handler/src/write-request-body.ts
@@ -58,9 +58,30 @@ function writeBody(
   if (body instanceof Readable) {
     // pipe automatically handles end
     body.pipe(httpRequest);
-  } else if (body) {
-    httpRequest.end(Buffer.from(body as Parameters<typeof Buffer.from>[0]));
-  } else {
-    httpRequest.end();
+    return;
   }
+
+  if (body) {
+    if (Buffer.isBuffer(body) || typeof body === "string") {
+      httpRequest.end(body);
+      return;
+    }
+
+    const uint8 = body as Uint8Array;
+    if (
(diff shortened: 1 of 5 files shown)


packages/util-body-length-node/src/calculateBodyLength.ts
```diff
@@ -8,7 +8,7 @@ export const calculateBodyLength = (body: any): number | undefined => {
     return 0;
   }
   if (typeof body === "string") {
-    return Buffer.from(body).length;
+    return Buffer.byteLength(body);
```
kuhe, line 11: Confirmed this works in Node.js 14+

packages/node-http-handler/src/write-request-body.ts
```diff
@@ -59,7 +59,14 @@ function writeBody(
     // pipe automatically handles end
     body.pipe(httpRequest);
   } else if (body) {
-    httpRequest.end(Buffer.from(body as Parameters<typeof Buffer.from>[0]));
+    if (Buffer.isBuffer(body) || typeof body === "string") {
+      httpRequest.end(body);
```
kuhe, line 63: ClientRequest (`node:http(s)`) and ClientHttp2Stream (`node:http2`) both accept `.end(string)` in addition to `.end(Buffer)`.

packages/util-body-length-node/src/calculateBodyLength.ts
kuhe:
```
> Buffer.byteLength('🚀a')
5
> Buffer.from('🚀a').length
5
```

trivikr (approved):

syall (approved):

PR #875, opened 2023-08-15
chore: update yarn lock

kuhe:
I ran `yarn` on the main branch and there is a diff.


kuhe:
this is due to the weather client existing locally

PR #881, opened 2023-08-17
feat(experimentalIdentityAndAuth): add codegen and TS integration points for `config`

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

**NOTE: All of these are under the `experimentalIdentityAndAuth` feature flag**

Start implementing core code generation and TypeScript package for the client `config`. This change introduces:

- Adding config properties `httpAuthSchemes` and `httpAuthSchemeProvider` and corresponding runtime config defaults
  ```typescript
  /**
   * experimentalIdentityAndAuth: Configuration of HttpAuthSchemes for a client which provides default identity providers and signers per auth scheme.
   * @internal
   */
  httpAuthSchemes?: IdentityProviderConfiguration;

  /**
   * experimentalIdentityAndAuth: Configuration of an HttpAuthSchemeProvider for a client which resolves which HttpAuthScheme to use.
   * @internal
   */
  httpAuthSchemeProvider?: WeatherHttpAuthSchemeProvider;
  ```
- Generating the model-based `HttpAuthSchemeProvider` and interfaces per client
  - Only no auth trait clients are supported in this PR, but other auth traits support will be implemented in follow-up PRs.
- The `@smithy/experimental-identity-and-auth` package, which will be reorganized into proper packages and deprecated after feature development is complete.
  - This package should stay within `0.x.x`, and never have a major version bump.
- Temporarily removing auth traits from generic client test since they will fail codegen until the auth trait is supported

*Testing:*

- No authentication trait client codegen diff: https://gist.github.com/syall/a595c7738c380176a528ff29016f1c2c#file-no-auth-trait-client-diff
- See the following PRs example auth scheme support using this feature:
  - https://github.com/awslabs/smithy-typescript/pull/883
  - https://github.com/awslabs/smithy-typescript/pull/884
  - https://github.com/awslabs/smithy-typescript/pull/885
- No codegen diff for `aws-sdk-js-v3`

TODO:

- [x] Add javadoc
- [x] Add TypeScript documentation
- [x] Add generating documentation for config properties
- [x] Test with an AuthScheme (planned `@httpApiKeyAuth`)

---

### DIFF
--- a/packages/experimental-identity-and-auth/package.json
+++ b/packages/experimental-identity-and-auth/package.json
@@ -1,0 +1,63 @@
+{
+  "name": "@smithy/experimental-identity-and-auth",
+  "version": "0.0.0",
+  "scripts": {
+    "build": "concurrently 'yarn:build:cjs' 'yarn:build:es' 'yarn:build:types && yarn build:types:downlevel'",
+    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build:es": "tsc -p tsconfig.es.json",
+    "build:types": "tsc -p tsconfig.types.json",
+    "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
+    "stage-release": "rimraf ./.release && yarn pack && mkdir ./.release && tar zxvf ./package.tgz --directory ./.release && rm ./package.tgz",
+    "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo",
+    "lint": "eslint -c ../../.eslintrc.js \"src/**/*.ts\"",
+    "format": "prettier --config ../../prettier.config.js --ignore-path ../.prettierignore --write \"**/*.{ts,md,json}\"",
+    "test": "tsc -p tsconfig.test.json"
+  },
+  "main": "./dist-cjs/index.js",
+  "module": "./dist-es/index.js",
+  "types": "./dist-types/index.d.ts",
+  "author": {
+    "name": "AWS Smithy Team",
+    "email": "",
+    "url": "https://smithy.io"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@smithy/types": "workspace:^",
+    "tslib": "^2.5.0"
+  },
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "typesVersions": {
+    "<4.0": {
+      "dist-types/*": [
+        "dist-types/ts3.4/*"
+      ]
+    }
+  },
+  "files": [
+    "dist-*/**"
+  ],
+  "homepage": "https://github.com/awslabs/smithy-typescript/tree/main/packages/experimental-identity-and-auth",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/awslabs/smithy-typescript.git",
+    "directory": "packages/experimental-identity-and-auth"
+  },
+  "devDependencies": {
+    "@tsconfig/recommended": "1.0.1",
+    "concurrently": "7.0.0",
+    "downlevel-dts": "0.10.1",
+    "jest": "28.1.1",
+    "rimraf": "3.0.2",
+    "typedoc": "0.23.23",
+    "typescript": "~4.9.5"
+  },
+  "typedoc": {
+    "entryPoint": "src/index.ts"
+  },
+  "publishConfig": {
+    "directory": ".release/package"
+  }
+}
--- a/packages/experimental-identity-and-auth/src/index.ts
+++ b/packages/experimental-identity-and-auth/src/index.ts
@@ -1,0 +1,4 @@
+export * from "./IdentityProviderConfiguration";
+export * from "./HttpAuthScheme";
+export * from "./HttpSigner";
+export * from "./noAuth";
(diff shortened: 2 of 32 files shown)


packages/experimental-identity-and-auth/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@smithy/experimental-identity-and-auth",
```
kuhe, line 2: why does the package name include the word experimental? Doesn't that complicate promoting this to non-experimental later?

syall: I was thinking when the feature is ready to move out of experimental (code is stable), this package would be deprecated.

The code from this package would then be moved to the relevant packages (e.g. `@smithy/types`, individual middleware packages, etc.) and be published as production code.

This package should NOT be used in any production code at all.

packages/experimental-identity-and-auth/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build:es": "tsc -p tsconfig.es.json",
+    "build:types": "tsc -p tsconfig.types.json",
+    "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
+    "stage-release": "rimraf ./.release && yarn pack && mkdir ./.release && tar zxvf ./package.tgz --directory ./.release && rm ./package.tgz",
+    "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo",
+    "lint": "eslint -c ../../.eslintrc.js \"src/**/*.ts\"",
+    "format": "prettier --config ../../prettier.config.js --ignore-path ../.prettierignore --write \"**/*.{ts,md,json}\"",
+    "test": "tsc -p tsconfig.test.json",
+    "extract:docs": "api-extractor run --local"
```
kuhe, line 15: remove this line? so that public API docs are not generated

packages/experimental-identity-and-auth/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+   * @param schemeId schemeId of the HttpAuthScheme
+   * @returns IdentityProvider or undefined if HttpAuthScheme is not found
+   */
+  getIdentityProvider(schemeId: string): IdentityProvider<Identity> | undefined;
+
+  /**
+   * Gets the configured HttpAuthSchemes.
+   * @returns all configured HttpAuthSchemes
+   */
+  getAuthSchemes(): Map<string, HttpAuthScheme>;
```
kuhe, line 19: are these keys known ahead of time? perhaps you could use a union that you can expand later.

packages/experimental-identity-and-auth/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export interface HttpSigner {
+  /**
+   * Signs an HttpRequest with an identity and signing properties.
+   * @param httpRequest request to sign
+   * @param identity identity to sing the request with
+   * @param signingProperties property bag for signing
+   * @returns signed request in a promise
+   */
+  sign(httpRequest: HttpRequest, identity: Identity, signingProperties: Record<string, any>): Promise<HttpRequest>;
```
kuhe, line 34: use `Record<string, unknown>` so that the value can be cast or fully runtime checked at the access point with the developer knowingly checking an arbitrary value.

packages/experimental-identity-and-auth/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Interface that defines the identity and signing properties when selecting
+ * an HttpAuthScheme.
+ * @internal
+ */
+export interface HttpAuthOption {
+  schemeId: string;
+  identityProperties?: Record<string, any>;
+  signingProperties?: Record<string, any>;
```
kuhe, line 93: use `unknown type` for values

packages/experimental-identity-and-auth/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Signer for the synthetic @smithy.api#noAuth auth scheme.
+ * @internal
+ */
+export class NoAuthSigner implements HttpSigner {
+  async sign(
+    httpRequest: HttpRequest,
+    identity: Identity,
+    signingProperties: Record<string, any>
```
kuhe, line 104: use `unknown` for `any`

kuhe:
I reviewed the TS code 👍 . For the Java code, I'd want to see the output diff to evaluate it.

packages/experimental-identity-and-auth/package.json
kuhe: and remove the api-extractor config from this pkg

packages/experimental-identity-and-auth/src/index.ts
syall: These keys are known at codegen time, so I think it'd be difficult pass that information into this TS implementation.

Fo simplicity, we can create a type alias in case we want to expand auto-complete later.

```typescript
type HttpAuthSchemeId = string;
```

AndrewFossAWS (approved):

trivikr (approved):

PR #886, opened 2023-08-18
move extensions runtime code to smithy-client

kuhe:
- moves the runtime portions of the extensions feature to `smithy-client`, out of `types`.
- exports extensions-related types from `types` pkg to be used in `smithy-client`.

### DIFF
--- a/packages/smithy-client/src/extensions/defaultClientConfiguration.ts
+++ b/packages/smithy-client/src/extensions/defaultClientConfiguration.ts
@@ -1,0 +1,30 @@
+import type { DefaultClientConfiguration } from "@smithy/types";
+
+import { getChecksumConfiguration, resolveChecksumRuntimeConfig } from "./checksum";
+
+/**
+ * @internal
+ */
+export type DefaultExtensionConfigType = Parameters<typeof getChecksumConfiguration>[0];
+
+/**
+ * @internal
+ *
+ * Helper function to resolve default client configuration from runtime config
+ */
+export const getDefaultClientConfiguration = (runtimeConfig: DefaultExtensionConfigType) => {
+  return {
+    ...getChecksumConfiguration(runtimeConfig),
+  };
+};
+
+/**
+ * @internal
+ *
+ * Helper function to resolve runtime config from default client configuration
+ */
+export const resolveDefaultRuntimeConfig = (config: DefaultClientConfiguration) => {
+  return {
+    ...resolveChecksumRuntimeConfig(config),
+  };
+};
--- a/packages/smithy-client/src/extensions/index.ts
+++ b/packages/smithy-client/src/extensions/index.ts
@@ -1,0 +1,1 @@
+export * from "./defaultClientConfiguration";
--- a/packages/types/src/extensions/defaultClientConfiguration.ts
+++ b/packages/types/src/extensions/defaultClientConfiguration.ts
@@ -7,12 +7,17 @@ import { ChecksumConfiguration, getChecksumConfiguration, resolveChecksumRuntime
  */
 export interface DefaultClientConfiguration extends ChecksumConfiguration {}
 
+/**
+ * @deprecated will be removed for implicit type.
+ */
 type GetDefaultConfigurationType = (runtimeConfig: any) => DefaultClientConfiguration;
 
 /**
+ * @deprecated moving to @smithy/smithy-client.
  * @internal
  *
  * Helper function to resolve default client configuration from runtime config
+ *
  */
 export const getDefaultClientConfiguration: GetDefaultConfigurationType = (runtimeConfig: any) => {
   return {
@@ -20,9 +25,13 @@ export const getDefaultClientConfiguration: GetDefaultConfigurationType = (runti
   };
 };
 
+/**
+ * @deprecated will be removed for implicit type.
+ */
 type ResolveDefaultRuntimeConfigType = (clientConfig: DefaultClientConfiguration) => any;
 
 /**
+ * @deprecated moving to @smithy/smithy-client.
  * @internal
  *
  * Helper function to resolve runtime config from default client configuration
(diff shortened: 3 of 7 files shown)


packages/types/src/extensions/checksum.ts
```diff
@@ -20,13 +20,22 @@ export interface ChecksumAlgorithm {
...
+ * @internal
+ */
 export interface ChecksumConfiguration {
   addChecksumAlgorithm(algo: ChecksumAlgorithm): void;
   checksumAlgorithms(): ChecksumAlgorithm[];
 
+  /**
+   * @deprecated unused.
+   */
   [other: string | number]: any;
```
kuhe, line 33: I removed this in the move. What was the reason for these unbounded set of keys?

packages/types/src/extensions/defaultClientConfiguration.ts
```diff
@@ -7,22 +7,31 @@ import { ChecksumConfiguration, getChecksumConfiguration, resolveChecksumRuntime
...
 type GetDefaultConfigurationType = (runtimeConfig: any) => DefaultClientConfiguration;
 
 /**
+ * @deprecated moving to @smithy/smithy-client.
  * @internal
  *
  * Helper function to resolve default client configuration from runtime config
+ *
  */
 export const getDefaultClientConfiguration: GetDefaultConfigurationType = (runtimeConfig: any) => {
```
kuhe, line 22: in the move I changed this `any` to a type that can be extended later as you add default extensions.

turbo.json
```diff
@@ -4,6 +4,7 @@
   "pipeline": {
     "build": {
       "dependsOn": ["^build"],
+      "inputs": ["src", "package.json"],
```
kuhe, line 7: @srchase I couldn't get `yarn build` in the root to work without this. Builds would fail because the turbo cache would output a blank dist-cjs and dist-types folder for some packages. Is it possible some of the other turbo commands are causing bad values to be stored in the turbo cache?

packages/smithy-client/src/extensions/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./defaultClientConfiguration";
```
kuhe, line 1: should this also export from checksum, or is checksum entirely private to `defaultClientConfiguration`?

packages/smithy-client/src/extensions/defaultClientConfiguration.ts
```diff
@@ -0,0 +1,30 @@
+import type { DefaultClientConfiguration } from "@smithy/types";
+
+import { getChecksumConfiguration, resolveChecksumRuntimeConfig } from "./checksum";
+
+/**
+ * @internal
+ */
+export type DefaultExtensionConfigType = Parameters<typeof getChecksumConfiguration>[0];
```
kuhe, line 8: This type replaces `any`. It can form a union later when more extensions exist. Is that consistent with the intended behavior?

kuhe:
additional notes:

is the file `resolveRuntimeExtensions1.template` used? It seems to differ in only 1 line from `runtimeExtensions1.template`.

Notes on the generated file:
```typescript
import { getDefaultClientConfiguration, resolveDefaultRuntimeConfig } from "@smithy/types";

import { S3ClientConfiguration } from "./clientConfiguration";

/**
 * @public
 * This should be renamed S3RuntimeExtension, since these differ by client and are public.
 */
export interface RuntimeExtension {
  configureClient(clientConfiguration: S3ClientConfiguration): void;
}

/**
 * @public
 * This should be renamed S3RuntimeExtensionsResolvedConfig, because it is used
 * in the resolved config intersection type.
 */
export interface RuntimeExtensionsConfig {
  extensions: RuntimeExtension[];
}

/**
 * This function should be removed. Type casting can be performed by the type system.
 */
const asPartial = <T extends Partial<S3ClientConfiguration>>(t: T) => t;

/**
 * @internal
 * This should not accept nor return the any type.
 */
export const resolveRuntimeExtensions = (runtimeConfig: any, extensions: RuntimeExtension[]) => {
  const clientConfiguration: S3ClientConfiguration = {
    ...asPartial(getDefaultClientConfiguration(runtimeConfig)),
  };

  extensions.forEach((extension) => extension.configureClient(clientConfiguration));

  return {
    ...runtimeConfig,
    ...resolveDefaultRuntimeConfig(clientConfiguration),
  };
};
```

proposed result:
```typescript
import { getDefaultClientConfiguration, resolveDefaultRuntimeConfig } from "@smithy/types";

import { S3ClientConfiguration } from "./clientConfiguration";
import { S3ClientResolvedConfigType } from "./S3Client";

/**
 * @public
 */
export interface S3RuntimeExtension {
  configureClient(clientConfiguration: S3ClientConfiguration): void;
}

/**
 * @public
 */
export interface S3RuntimeExtensionsResolvedConfig {
  extensions: S3RuntimeExtension[];
}

/**
 * @internal
 * Having the "resolved" type for runtimeConfig firmly states that this is the last config resolver.
 * However, it's not entirely true, since the resolving of the extensions field may not have occurred because
 * it is supposed to be handled by this code.
 */
export const resolveRuntimeExtensions = (
  runtimeConfig: S3ClientResolvedConfigType,
  extensions: S3RuntimeExtension[]
) => {
  const clientConfiguration: S3ClientConfiguration = {
    ...getDefaultClientConfiguration(runtimeConfig),
  };

  extensions.forEach((extension) => extension.configureClient(clientConfiguration));

  return {
    ...runtimeConfig,
    ...resolveDefaultRuntimeConfig(clientConfiguration),
  } as S3ClientResolvedConfigType;
};

```

packages/smithy-client/src/extensions/index.ts
AndrewFossAWS: checksum should be entirely private to defaultClientConfiguration

trivikr (approved):

PR #894, opened 2023-08-24
Add `matchSettings()` to `TypeScriptIntegration` and `TypeScriptCodegenPlugin`

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Filter `TypeScriptIntegration`s based on `TypeScriptSettings`.

This allows for integrations to be enabled or disabled based on settings in `smithy-build.json`. 
    
By default `matchSettings()` will filter on the new `excludedIntegrations` setting, which defines a set of integrations to exclude by name.

An example of how this new setting is used can be found here: https://github.com/awslabs/smithy-typescript/commit/6b5cc2e616de6ab781a734b181b310cd8d49fbb2

Also see: https://github.com/awslabs/smithy-typescript/pull/856

---


gosar:
Can you explain the motivation / use case that wants to add this change, so we can evaluate if this is how to solve it, or another way?

PR #898, opened 2023-08-28
Add retry to runtime extension

AndrewFossAWS:
This PR adds retry component to the runtime extensions. 

Using S3 as an example:
```
export class CustomRetryExtension implements RuntimeExtension {
    configure(extensionConfiguration: S3ExtensionConfiguration): void {
       extensionConfiguration.setRetryStrategy(...);
    }
}

S3Client({
  ...
  extensions: [CustomRetryExtension()]
})
```

---

### DIFF
--- a/packages/smithy-client/src/extensions/defaultExtensionConfiguration.ts
+++ b/packages/smithy-client/src/extensions/defaultExtensionConfiguration.ts
@@ -1,20 +1,22 @@
 import type { DefaultExtensionConfiguration } from "@smithy/types";
 
-import { getChecksumConfiguration, resolveChecksumRuntimeConfig } from "./checksum";
+import { getChecksumConfiguration, PartialChecksumRuntimeConfigType, resolveChecksumRuntimeConfig } from "./checksum";
+import { getRetryConfiguration, PartialRetryRuntimeConfigType, resolveRetryRuntimeConfig } from "./retry";
 
 /**
  * @internal
  */
-export type DefaultExtensionConfigType = Parameters<typeof getChecksumConfiguration>[0];
+export type DefaultExtensionRuntimeConfigType = PartialRetryRuntimeConfigType & PartialChecksumRuntimeConfigType;
 
 /**
  * @internal
  *
  * Helper function to resolve default extension configuration from runtime config
  */
-export const getDefaultExtensionConfiguration = (runtimeConfig: DefaultExtensionConfigType) => {
+export const getDefaultExtensionConfiguration = (runtimeConfig: DefaultExtensionRuntimeConfigType) => {
   return {
     ...getChecksumConfiguration(runtimeConfig),
+    ...getRetryConfiguration(runtimeConfig),
   };
 };
 
@@ -31,8 +33,11 @@ export const getDefaultClientConfiguration = getDefaultExtensionConfiguration;
  *
  * Helper function to resolve runtime config from default extension configuration
  */
-export const resolveDefaultRuntimeConfig = (config: DefaultExtensionConfiguration) => {
+export const resolveDefaultRuntimeConfig = (
+  config: DefaultExtensionConfiguration
+): DefaultExtensionRuntimeConfigType => {
   return {
     ...resolveChecksumRuntimeConfig(config),
+    ...resolveRetryRuntimeConfig(config),
   };
 };
--- a/packages/smithy-client/src/extensions/retry.ts
+++ b/packages/smithy-client/src/extensions/retry.ts
@@ -1,0 +1,29 @@
+import { Provider, RetryStrategy, RetryStrategyConfiguration, RetryStrategyV2 } from "@smithy/types";
+
+export type PartialRetryRuntimeConfigType = Partial<{ retryStrategy: Provider<RetryStrategyV2 | RetryStrategy> }>;
+
+/**
+ * @internal
+ */
+export const getRetryConfiguration = (runtimeConfig: PartialRetryRuntimeConfigType) => {
+  let _retryStrategy = runtimeConfig.retryStrategy!;
+  return {
+    setRetryStrategy(retryStrategy: Provider<RetryStrategyV2 | RetryStrategy>): void {
+      _retryStrategy = retryStrategy;
+    },
+    retryStrategy(): Provider<RetryStrategyV2 | RetryStrategy> {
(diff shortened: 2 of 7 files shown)


packages/smithy-client/src/extensions/retry.ts
```diff
@@ -0,0 +1,27 @@
+import { Provider, RetryStrategy, RetryStrategyConfiguration, RetryStrategyV2 } from "@smithy/types";
+
+/**
+ * @internal
+ */
+export const getRetryConfiguration = (
+  runtimeConfig: Partial<{ retryStrategy: Provider<RetryStrategyV2 | RetryStrategy> }>
+) => {
+  return {
+    _retryStrategy: runtimeConfig.retryStrategy!,
```
kuhe, line 10: this value looks private and can just be in the closure

```ts
(...) => {
    let _retryStrategy = runtimeConfig.retryStrategy!;
    setRetryStrategy(retryStrategy: Provider<RetryStrategyV2 | RetryStrategy>): void {
      _retryStrategy = retryStrategy;
    },
    retryStrategy(): Provider<RetryStrategyV2 | RetryStrategy> {
      return _retryStrategy;
    },
}
```

packages/smithy-client/src/extensions/retry.ts
```diff
@@ -0,0 +1,27 @@
...
+    setRetryStrategy(retryStrategy: Provider<RetryStrategyV2 | RetryStrategy>): void {
+      this._retryStrategy = retryStrategy;
```
kuhe, line 12: what is the scope of this setter? is it supposed to mutate the retryStrategy of a client for all future requests?

packages/types/src/extensions/retry.ts
```diff
@@ -0,0 +1,10 @@
+import { RetryStrategyV2 } from "../retry";
+import { Provider, RetryStrategy } from "../util";
+
+/**
+ * @internal
+ */
+export interface RetryStrategyConfiguration {
+  setRetryStrategy(retryStrategy: Provider<RetryStrategyV2 | RetryStrategy>): void;
```
kuhe, line 8: add a jsdoc to this method: what situation involves calling this, and what is the expected consequence?

packages/smithy-client/src/extensions/defaultExtensionConfiguration.ts
```diff
@@ -1,11 +1,13 @@
...
 
 import { getChecksumConfiguration, resolveChecksumRuntimeConfig } from "./checksum";
+import { getRetryConfiguration, resolveRetryRuntimeConfig } from "./retry";
 
 /**
  * @internal
  */
-export type DefaultExtensionConfigType = Parameters<typeof getChecksumConfiguration>[0];
+export type DefaultExtensionConfigType = Parameters<typeof getChecksumConfiguration>[0] &
+  Parameters<typeof getRetryConfiguration>[0];
```
syall, line 10: Is there a way to make this more readable? I'm not sure what this means.

packages/smithy-client/src/extensions/retry.ts
```diff
@@ -0,0 +1,27 @@
+import { Provider, RetryStrategy, RetryStrategyConfiguration, RetryStrategyV2 } from "@smithy/types";
+
+/**
+ * @internal
+ */
+export const getRetryConfiguration = (
+  runtimeConfig: Partial<{ retryStrategy: Provider<RetryStrategyV2 | RetryStrategy> }>
```
syall, line 7: Is there a reason why the type signature here is different than `resolveRetryRuntimeConfig()`?

I think we should extract the `runtimeConfig` type to it's own interface to keep it consistent.

```typescript
/**
 * @internal
 */
export interface RetryConfigurationRuntimeConfig 
  extends Partial<{ retryStrategy: Provider<RetryStrategyV2 | RetryStrategy> }> {}
```

Although not in scope of this PR, the checksum configuration should also be changed to match this pattern.

kuhe (approved):

packages/types/src/extensions/defaultExtensionConfiguration.ts
```diff
@@ -1,10 +1,9 @@
...
+import { RetryStrategyConfiguration } from "./retry";
 
 /**
  * @internal
  *
  * Default extension configuration consisting various configurations for modifying a service client
  */
-export interface DefaultExtensionConfiguration extends ChecksumConfiguration {}
-
-type GetDefaultConfigurationType = (runtimeConfig: any) => DefaultExtensionConfiguration;
```
syall, line 10: Any risk of `GetDefaultConfigurationType` being removed?

packages/types/src/extensions/retry.ts
```diff
@@ -0,0 +1,20 @@
...
+   * Set retry strategy used for all http requests
+   * @param retryStrategy
+   */
+  setRetryStrategy(retryStrategy: Provider<RetryStrategyV2 | RetryStrategy>): void;
+
+  /**
+   * Get retry strategy used for all http requests
+   * @param retryStrategy
+   */
+  retryStrategy(): Provider<RetryStrategyV2 | RetryStrategy>;
```
syall, line 19: nit, but since `Provider<RetryStrategyV2 | RetryStrategy>` is repeated so much, maybe make a type alias for it?

syall (changes requested):

packages/types/src/extensions/defaultExtensionConfiguration.ts
AndrewFossAWS: No because it's not used

packages/smithy-client/src/extensions/defaultExtensionConfiguration.ts
```diff
@@ -34,5 +36,6 @@ export const getDefaultClientConfiguration = getDefaultExtensionConfiguration;
 export const resolveDefaultRuntimeConfig = (config: DefaultExtensionConfiguration) => {
   return {
     ...resolveChecksumRuntimeConfig(config),
-  };
+    ...resolveRetryRuntimeConfig(config),
+  } as RuntimeConfigType;
```
syall, line 40: nit: would prefer the return type to be in the function signature rather than using `as`.

syall (changes requested):

syall (approved):

PR #899, opened 2023-08-28
Add http client component to runtime extension

AndrewFossAWS:
This PR adds http client component to the runtime extensions. 

Using S3 as an example:
```
export class CustomHttpExtension implements RuntimeExtension {
    configure(extensionConfiguration: S3ExtensionConfiguration): void {
       extensionConfiguration.setHttpHandler(...);
       extensionConfiguration.updateHttpClientConfig("requestTimeout", requestTimeout);
    }
}

S3Client({
  ...
  extensions: [CustomHttpExtension()]
})
```

---

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -19,9 +19,9 @@ export interface FetchHttpHandlerOptions {
 
 type FetchHttpHandlerConfig = FetchHttpHandlerOptions;
 
-export class FetchHttpHandler implements HttpHandler {
+export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
   private config?: FetchHttpHandlerConfig;
-  private readonly configProvider: Promise<FetchHttpHandlerConfig>;
+  private configProvider: Promise<FetchHttpHandlerConfig>;
 
   constructor(options?: FetchHttpHandlerOptions | Provider<FetchHttpHandlerOptions | undefined>) {
     if (typeof options === "function") {
@@ -129,5 +129,17 @@ export class FetchHttpHandler implements HttpHandler {
       );
     }
     return Promise.race(raceOfPromises);
+  }
+
+  updateHttpClientConfig(key: keyof FetchHttpHandlerConfig, value: FetchHttpHandlerConfig[typeof key]): void {
+    this.config = undefined;
+    this.configProvider = this.configProvider.then((config) => {
+      config[key] = value;
+      return config;
+    });
+  }
+
+  httpHandlerConfigs(): FetchHttpHandlerConfig {
+    return this.config ?? {};
   }
 }
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -50,9 +50,9 @@ interface ResolvedNodeHttpHandlerConfig {
 
 export const DEFAULT_REQUEST_TIMEOUT = 0;
 
-export class NodeHttpHandler implements HttpHandler {
+export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
   private config?: ResolvedNodeHttpHandlerConfig;
-  private readonly configProvider: Promise<ResolvedNodeHttpHandlerConfig>;
+  private configProvider: Promise<ResolvedNodeHttpHandlerConfig>;
 
   // Node http handler is hard-coded to http/1.1: https://github.com/nodejs/node/blob/ff5664b83b89c55e4ab5d5f60068fb457f1f5872/lib/_http_server.js#L286
   public readonly metadata = { handlerProtocol: "http/1.1" };
@@ -191,5 +191,19 @@ export class NodeHttpHandler implements HttpHandler {
 
       writeRequestBodyPromise = writeRequestBody(req, request, this.config.requestTimeout).catch(_reject);
     });
+  }
+
+  updateHttpClientConfig(key: keyof NodeHttpHandlerOptions, value: NodeHttpHandlerOptions[typeof key]): void {
+    this.config = undefined;
+    this.configProvider = this.configProvider.then((config) => {
+      return {
+        ...config,
+        [key]: value,
+      };
+    });
+  }
+
+  httpHandlerConfigs(): NodeHttpHandlerOptions {
+    return this.config ?? {};
   }
 }
--- a/packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
+++ b/packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
@@ -1,0 +1,56 @@
+import { HttpHandler } from "../httpHandler";
+
+/**
+ * @internal
+ */
+export interface HttpHandlerExtensionConfiguration<HandlerConfig extends object = Record<string, unknown>> {
+  setHttpHandler(handler: HttpHandler<HandlerConfig>): void;
+  httpHandler(): HttpHandler<HandlerConfig>;
+  updateHttpClientConfig(key: keyof HandlerConfig, value: HandlerConfig[typeof key]): void;
+  httpHandlerConfigs(): HandlerConfig;
+}
+
+/**
+ * @internal
+ */
+export type HttpHandlerExtensionConfigType<HandlerConfig extends object = Record<string, unknown>> = Partial<{
+  httpHandler: HttpHandler<HandlerConfig>;
+}>;
+
+/**
+ * @internal
+ *
+ * Helper function to resolve default extension configuration from runtime config
+ */
+export const getHttpHandlerExtensionConfiguration = <HandlerConfig extends object = Record<string, unknown>>(
+  runtimeConfig: HttpHandlerExtensionConfigType<HandlerConfig>
+) => {
+  let httpHandler = runtimeConfig.httpHandler!;
+  return {
+    setHttpHandler(handler: HttpHandler<HandlerConfig>): void {
+      httpHandler = handler;
+    },
+    httpHandler(): HttpHandler<HandlerConfig> {
+      return httpHandler;
+    },
+    updateHttpClientConfig(key: keyof HandlerConfig, value: HandlerConfig[typeof key]): void {
+      httpHandler.updateHttpClientConfig(key, value);
+    },
+    httpHandlerConfigs(): HandlerConfig {
+      return httpHandler.httpHandlerConfigs();
+    },
+  };
+};
+
+/**
+ * @internal
+ *
+ * Helper function to resolve runtime config from default extension configuration
+ */
+export const resolveHttpHandlerRuntimeConfig = <HandlerConfig extends object = Record<string, unknown>>(
+  httpHandlerExtensionConfiguration: HttpHandlerExtensionConfiguration<HandlerConfig>
+): HttpHandlerExtensionConfigType<HandlerConfig> => {
+  return {
+    httpHandler: httpHandlerExtensionConfiguration.httpHandler(),
+  };
+};
--- a/packages/protocol-http/src/httpHandler.ts
+++ b/packages/protocol-http/src/httpHandler.ts
@@ -3,4 +3,19 @@ import { HttpHandlerOptions, RequestHandler } from "@smithy/types";
 import { HttpRequest } from "./httpRequest";
 import { HttpResponse } from "./httpResponse";
 
-export type HttpHandler = RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions>;
+/**
+ * @internal
+ */
+export type HttpHandler<HttpHandlerConfig> = RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions> & {
+  /**
+   * @internal
+   * @param key
+   * @param value
+   */
+  updateHttpClientConfig(key: keyof HttpHandlerConfig, value: HttpHandlerConfig[typeof key]): void;
+
+  /**
+   * @internal
+   */
+  httpHandlerConfigs(): HttpHandlerConfig;
+};
--- a/.changeset/shy-kids-call.md
+++ b/.changeset/shy-kids-call.md
@@ -1,0 +1,8 @@
+---
+"@smithy/protocol-http": major
+"@smithy/fetch-http-handler": minor
+"@smithy/node-http-handler": minor
+"@smithy/util-test": minor
+---
+
+Add http client component to runtime extension
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "test": "turbo run test",
     "test:integration": "yarn build-test-packages && turbo run test:integration",
     "lint": "turbo run lint",
+    "lint-fix": "turbo run lint -- --fix",
     "format": "turbo run format --parallel",
     "stage-release": "turbo run stage-release",
     "extract:docs": "mkdir -p api-extractor-packages && turbo run extract:docs",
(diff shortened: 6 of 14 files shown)


packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -130,4 +130,18 @@ export class FetchHttpHandler implements HttpHandler {
     }
     return Promise.race(raceOfPromises);
   }
+
+  updateHttpClientConfig(key: string, value: any): void {
```
kuhe, line 134: `any` should not be used in most cases. This interfaces appears to be

```ts
updateHttpClientConfig<
  ConfigKey extends keyof FetchHttpHandlerConfig
>(key: ConfigKey, value: FetchHttpHandlerConfig[ConfigKey]): void
```

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -130,4 +130,18 @@ export class FetchHttpHandler implements HttpHandler {
...
+    this.config = undefined;
+    this.configProvider = this.configProvider.then((resolveConfig) => {
+      return {
+        ...resolveConfig,
+        [key]: value,
+      };
+    });
```
kuhe, line 141: is this copy necessary?

```ts
this.configProvider = this.configProvider.then(config => {
  config[key] = value;
  return config;
});
```

should suffice?

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -130,4 +130,18 @@ export class FetchHttpHandler implements HttpHandler {
...
+  }
+
+  httpHandlerConfigs(): Record<string, any> {
+    return this.config || {};
```
kuhe, line 145: this return type appears to be `FetchHttpHandlerConfig`.

packages/protocol-http/src/httpHandler.ts
```diff
@@ -3,4 +3,19 @@ import { HttpHandlerOptions, RequestHandler } from "@smithy/types";
...
+   * @internal
+   * @param key
+   * @param value
+   */
+  updateHttpClientConfig: (key: string, value: any) => void;
+
+  /**
+   * @internal
+   */
+  httpHandlerConfigs: () => Record<string, any>;
```
kuhe, line 20: this should be generic on the config type instead of `any`

kuhe (changes requested):
`any` should be replaced by types

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,39 @@
...
+  };
+};
+
+/**
+ * @internal
+ */
+export const resolveHttpHandlerRuntimeConfig = (
+  httpHandlerExtensionConfiguration: HttpHandlerExtensionConfiguration
+) => {
+  const runtimeConfig: Partial<Record<string, HttpHandler>> = {};
```
kuhe, line 32: this doesn't appear to be a Record

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,39 @@
...
+  runtimeConfig.requestHandler = httpHandlerExtensionConfiguration.httpHandler();
+  const record: Record<string, any> = httpHandlerExtensionConfiguration.httpHandlerConfigs();
```
kuhe, line 34: needs typings instead of Record and `any`

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
```diff
@@ -0,0 +1,42 @@
...
+} from "./http";
+
+/**
+ * @internal
+ */
+export interface HttpHandlerExtensionConfiguration {
+  setHttpHandler(handler: HttpHandler): void;
+  httpHandler(): HttpHandler;
+  updateHttpClientConfig(key: string, value: any): void;
+  httpHandlerConfigs(): Record<string, any>;
```
kuhe, line 14: use a type param instead of Record with `any`

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -192,4 +192,18 @@ export class NodeHttpHandler implements HttpHandler {
...
     });
   }
+
+  updateHttpClientConfig(key: keyof NodeHttpHandlerOptions, value: NodeHttpHandlerOptions[typeof key]): void {
+    this.config = undefined;
+    this.configProvider = this.configProvider.then((resolveConfig) => {
+      return {
+        ...resolveConfig,
+        [key]: value,
+      };
```
kuhe, line 202: this can be the same as the fetch http one with no copy.

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,42 @@
...
+  runtimeConfig: Partial<{ httpHandler: HttpHandler<HandlerConfig> }>
+) => {
+  let httpHandler = runtimeConfig.httpHandler!;
+  const httpHandlerConfigs: HandlerConfig = httpHandler.httpHandlerConfigs();
+  return {
+    setHttpHandler(handler: HttpHandler<HandlerConfig>): void {
+      httpHandler = handler;
+    },
+    httpHandler(): HttpHandler<HandlerConfig> {
+      return httpHandler;
```
kuhe, line 17: when setHttpHandler is called, `httpHandler()` will start to return its new value. However, the config's httphandler will not change. Is that intended?

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+    updateHttpClientConfig(key: keyof HandlerConfig, value: HandlerConfig[typeof key]): void {
+      httpHandlerConfigs[key] = value;
+    },
+    httpHandlerConfigs(): HandlerConfig {
+      return httpHandlerConfigs;
```
kuhe, line 23: this doesn't look right, the value won't change since it is a const ref. 

should it be?:
```
return this.httpHandler().httpHandlerConfigs()
```

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,42 @@
...
+  const httpHandlerConfigs: HandlerConfig = httpHandler.httpHandlerConfigs();
+  return {
+    setHttpHandler(handler: HttpHandler<HandlerConfig>): void {
+      httpHandler = handler;
+    },
+    httpHandler(): HttpHandler<HandlerConfig> {
+      return httpHandler;
+    },
+    updateHttpClientConfig(key: keyof HandlerConfig, value: HandlerConfig[typeof key]): void {
+      httpHandlerConfigs[key] = value;
```
kuhe, line 20: should this be:
```
this.httpHandler().httpHandlerConfigs()[key] = value
```
?

This reference looks like it will become stale after a call to setHttpHandler

packages/protocol-http/src/extensions/http.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @internal
+ */
+export const resolveHttpHandlerRuntimeConfig = <HandlerConfig extends object = Record<string, unknown>>(
+  httpHandlerExtensionConfiguration: HttpHandlerExtensionConfiguration<HandlerConfig>
+) => {
+  const requestHandler = httpHandlerExtensionConfiguration.httpHandler();
+  const handlerConfig = httpHandlerExtensionConfiguration.httpHandlerConfigs();
+  for (const key in handlerConfig) {
+    requestHandler.updateHttpClientConfig(key, handlerConfig[key]);
+  }
```
kuhe, line 38: why is this code in this position?

it looks like it just tells a request handler to set its config to values provided by its own config, redundant?

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
```diff
@@ -0,0 +1,42 @@
...
+    ...getHttpHandlerConfiguration(runtimeConfig),
+  };
+};
+
+/**
+ * @internal
+ *
+ * Helper function to resolve runtime config from default extension configuration
+ */
+export const resolveHttpHandlerRuntimeConfig = (config: HttpHandlerExtensionConfiguration) => {
```
kuhe, line 38: why is this function needed? it only performs an object copy?

packages/node-http-handler/src/node-http-handler.ts
AndrewFossAWS: I tried and it throws `Type 'number | boolean' is not assignable to type 'never'.   Type 'number' is not assignable to type 'never'` error

packages/protocol-http/src/httpHandler.ts
```diff
@@ -3,4 +3,19 @@ import { HttpHandlerOptions, RequestHandler } from "@smithy/types";
...
+   * @internal
+   * @param key
+   * @param value
+   */
+  updateHttpClientConfig(key: keyof HttpHandlerConfig, value: HttpHandlerConfig[typeof key]): void;
+
+  /**
+   * @internal
+   */
+  httpHandlerConfigs: () => HttpHandlerConfig;
```
kuhe, line 20: the actual implementations are not arrow functions, right? 

this should be `httpHandlerConfigs(): HttpHandlerConfig`

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -130,4 +130,16 @@ export class FetchHttpHandler implements HttpHandler {
     }
     return Promise.race(raceOfPromises);
   }
+
+  updateHttpClientConfig(key: keyof FetchHttpHandlerConfig, value: FetchHttpHandlerConfig[typeof key]): void {
+    this.config = undefined;
```
syall, line 135: Is it safe to just "discard" `this.config`'s value?

packages/node-http-handler/src/node-http-handler.ts
kuhe: what about with 

```ts
(config: NodeHttpHandlerOptions) => {
    config[key] = value;
    return config;
}
```

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -130,4 +130,16 @@ export class FetchHttpHandler implements HttpHandler {
...
+    this.configProvider = this.configProvider.then((resolveConfig) => {
```
kuhe, line 136: the resolved value of configProvider should be named `config`, because `resolveConfig` sounds like an action (method).

syall (changes requested):
Overall, I also agree with @kuhe's comments.

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
AndrewFossAWS: Initially there was plan to add authentication related logic inside resolveHttpHandlerRuntimeConfig. I will remove the duplicated function for now.

packages/node-http-handler/src/node-http-handler.ts
AndrewFossAWS: same error

packages/protocol-http/src/httpHandler.ts
```diff
@@ -3,4 +3,19 @@ import { HttpHandlerOptions, RequestHandler } from "@smithy/types";
...
+/**
+ * @internal
+ */
+export type HttpHandler<HttpHandlerConfig> = RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions> & {
+  /**
+   * @internal
+   * @param key
+   * @param value
+   */
+  updateHttpClientConfig(key: keyof HttpHandlerConfig, value: HttpHandlerConfig[typeof key]): void;
```
syall, line 15: I think there is too much type magic going on here.
Can we not just assume it can be any generic key / value pair and remove the generic?

packages/protocol-http/src/httpHandler.ts
```diff
@@ -3,4 +3,19 @@ import { HttpHandlerOptions, RequestHandler } from "@smithy/types";
 import { HttpRequest } from "./httpRequest";
 import { HttpResponse } from "./httpResponse";
 
-export type HttpHandler = RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions>;
+/**
+ * @internal
+ */
+export type HttpHandler<HttpHandlerConfig> = RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions> & {
```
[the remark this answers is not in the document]
syall: Also, existing implementations not implementing the new methods may be broken.

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
```diff
@@ -0,0 +1,42 @@
+import { HttpHandler } from "../httpHandler";
+import {
+  getHttpHandlerConfiguration,
+  resolveHttpHandlerRuntimeConfig as _resolveHttpHandlerRuntimeConfig,
+} from "./http";
+
+/**
+ * @internal
+ */
+export interface HttpHandlerExtensionConfiguration<HandlerConfig extends object = Record<string, unknown>> {
```
syall, line 10: I think we can remove the generic here.

packages/protocol-http/src/httpHandler.ts
kuhe: I don't think this is magic? It is saying that if you call updateConfig, it should be for a key and value type that is known to exist on the config's type.

If it feels restrictive, a compromise would be to intersect `HttpHandlerConfig` with `{ [key: string]: unknown }` for unbounded additional values. But, I don't think that should be necessary. 

For real type shenanigans, you should see what I committed to https://github.com/awslabs/smithy-typescript/blob/main/packages/types/src/transform/type-transform.ts

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
kuhe: I was discussing it with @AndrewFossAWS and we found a part of the spec that says the config "bag" should be treated in a type-safe manner if possible, so I believe that supports using this generic here. It has a fallback default that makes it easier to use if the config type is not known.

packages/fetch-http-handler/src/fetch-http-handler.ts
AndrewFossAWS: Not sure what you mean by "discard". `this.config` needs to set to `undefined` here because `handle` method call will populate `this.config` (with the updated config property) if it is undefined.

See https://github.com/awslabs/smithy-typescript/blob/main/packages/fetch-http-handler/src/fetch-http-handler.ts#L40-L42

packages/protocol-http/src/httpHandler.ts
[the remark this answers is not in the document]
AndrewFossAWS: Confirmed with @kuhe that `HttpHandler` isn't being used by much yet so it should be safe to changed without introducing a new interface e.g. `HttpHandlerV2`.

kuhe (approved):

packages/protocol-http/src/extensions/httpExtensionConfiguration.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @internal
+ *
+ * Helper function to resolve runtime config from default extension configuration
+ */
+export const resolveHttpHandlerRuntimeConfig = <HandlerConfig extends object = Record<string, unknown>>(
+  httpHandlerExtensionConfiguration: HttpHandlerExtensionConfiguration<HandlerConfig>
+) => {
+  return {
+    requestHandler: httpHandlerExtensionConfiguration.httpHandler(),
+  } as HttpHandlerExtensionConfigType<HandlerConfig>;
```
syall, line 55: nit: would prefer the return type to be in the function signature rather than using `as`.

.changeset/three-nails-press.md
```diff
@@ -0,0 +1,8 @@
+---
+"@smithy/fetch-http-handler": minor
+"@smithy/node-http-handler": minor
+"@smithy/protocol-http": minor
```
syall, line 4: I think this needs a major version bump due to breaking changes in interfaces, e.g. generics.

syall (changes requested):

syall (approved):

PR #907, opened 2023-09-02
feat(experimentalIdentityAndAuth): update `HttpAuthOption` and `HttpAuthScheme` codegen

syall:
### Issue
Issue number, if available, prefixed with "#"

N/A.

### Description
What does this implement/fix? Explain your changes.

After upgrading to Smithy 1.37.0
(https://github.com/awslabs/smithy-typescript/pull/906), upgrade to
`getEffectiveAuthSchemes` using `NO_AUTH_AWARE` and remove redundant
code.

For `HttpAuthOption`, allow any auth scheme to generate the
corresponding `HttpAuthOption` function regardless of registering it in
codegen.

For `HttpAuthScheme`, reduce code duplication by looking at "inherited"
`LanguageTarget` platforms (e.g. `REACT_NATIVE` inherits from
`BROWSER`). `httpAuthSchemes` is only written in a `runtimeConfig` if
the `IdentityProvider` and `Signer` are different between platforms, or
if the registered `HttpAuthScheme`s are different. `httpAuthSchemes` is
always written for `SHARED` as a default.

For each of the generic auth (`@httpApiKeyAuth`, `@httpBearerAuth`, and
`@aws.auth#sigv4`), make the identity providers and signers the same and
available on all platforms.

For the `httpAuthSchemes` property on the client config interface,
update it to `HttpAuthScheme[]` rather than
`IdentityProviderConfiguration`.

Also, add `@customAuth` auth scheme to generic tests,

### Testing
How was this change tested?

Everything is gated behind the `experimentalIdentityAndAuth` flag.

Observed codegen diff in generic client test: TODO

### Additional context
Add any other context about the PR here.

N/A.

---


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeConfigGenerator.java
```diff
@@ -269,6 +269,105 @@ private void generateHttpAuthSchemeConfig(
...
+            HttpAuthSchemeTarget o = (HttpAuthSchemeTarget) other;
+            return httpAuthScheme.equals(o.httpAuthScheme)
+                && identityProvider.equals(o.identityProvider)
+                && signer.equals(o.signer);
+        }
+
+        @Override
+        public int hashCode() {
+            return super.hashCode();
+        }
```
kuhe, line 301: is this the same if you leave this override out?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
```diff
@@ -5,25 +5,41 @@
...
-    public static final ShapeId NO_AUTH_ID = ShapeId.from("smithy.api#noAuth");
-
     private AuthUtils() {}
+
+    public static Map<ShapeId, HttpAuthScheme> getAllEffectiveNoAuthAwareAuthSchemes(
+        ServiceShape serviceShape,
+        ServiceIndex serviceIndex,
+        SupportedHttpAuthSchemesIndex authIndex
+    ) {
+        Map<ShapeId, HttpAuthScheme> effectiveAuthSchemes = new TreeMap<>();
```
kuhe, line 30: can you document why this map's order is important?

kuhe (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeConfigGenerator.java
syall: Yes, but `Checkstyle` will report a error violation:

```
Definition of 'equals()' without corresponding definition of 'hashCode()'.
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
syall: For this method, the map order is actually not important. The `TreeMap` implementation was arbitrarily chosen.

packages/experimental-identity-and-auth/src/IdentityProviderConfig.ts
```diff
@@ -1,32 +1,26 @@
...
  * Default implementation of IdentityProviderConfiguration
  * @internal
  */
-export class DefaultIdentityProviderConfiguration implements IdentityProviderConfiguration {
-  private authSchemes: Map<HttpAuthSchemeId, HttpAuthScheme> = new Map();
+export class DefaultIdentityProviderConfig implements IdentityProviderConfig {
+  private authSchemes: Map<HttpAuthSchemeId, IdentityProvider<Identity>> = new Map();
 
   /**
    * Creates an IdentityProviderConfiguration with a list of HttpAuthSchemes.
```
milesziemer, line 26: This doc and a few others need to be updated

milesziemer (changes requested):

packages/experimental-identity-and-auth/src/IdentityProviderConfig.ts
syall: Updated the relevant TS docs in the newest revision.

milesziemer (approved):

PR #916, opened 2023-09-07
Explicitly check for process.geteuid from global scope

richarddd:
*Description of changes:*

Windows doesn't have `geteuid` so import statement will fail running in ESM. We can assume that targets that have `geteuid` also have process in global scope.

---

### DIFF
--- a/.changeset/many-plums-agree.md
+++ b/.changeset/many-plums-agree.md
@@ -1,0 +1,5 @@
+---
+"@smithy/shared-ini-file-loader": minor
+---
+
+Fix invalid import on Windows hosts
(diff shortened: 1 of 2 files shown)


.changeset/many-plums-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/shared-ini-file-loader": minor
+---
+
+Fix invalid import on Windows hosts
```
trivikr, line 5:
```suggestion
Explicitly check for process.geteuid from global scope
```

.changeset/many-plums-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/shared-ini-file-loader": minor
```
trivikr, line 2:
```suggestion
"@smithy/shared-ini-file-loader": patch
```

trivikr (approved):

syall (approved):

AndrewFossAWS (approved):

PR #918, opened 2023-09-07
Fix default HttpHandlerConfig type for HttpHandler

AndrewFossAWS:
---

### DIFF
--- a/packages/protocol-http/src/httpHandler.ts
+++ b/packages/protocol-http/src/httpHandler.ts
@@ -6,7 +6,7 @@ import { HttpResponse } from "./httpResponse";
 /**
  * @internal
  */
-export type HttpHandler<HttpHandlerConfig extends object = Record<string, unknown>> = RequestHandler<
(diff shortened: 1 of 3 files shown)


packages/protocol-http/src/httpHandler.ts
```diff
@@ -6,7 +6,7 @@ import { HttpResponse } from "./httpResponse";
 /**
  * @internal
  */
-export type HttpHandler<HttpHandlerConfig extends object = Record<string, unknown>> = RequestHandler<
+export type HttpHandler<HttpHandlerConfig extends object = object> = RequestHandler<
```
kuhe, line 9: use `X extends object = {}`

kuhe (approved):

syall (approved):

packages/protocol-http/src/httpHandler.ts
syall: Is using `object` the same as `{}`? Isn't `{}` a literal type?

PR #930, opened 2023-09-13
Add `httpSigningMiddleware` to authorize and sign requests

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Add `httpSigningMiddleware` to authorize and sign requests.

Dependent on: https://github.com/awslabs/smithy-typescript/pull/927, https://github.com/awslabs/smithy-typescript/pull/928

---

### DIFF
--- a/packages/experimental-identity-and-auth/src/middleware-http-signing/httpSigningMiddleware.ts
+++ b/packages/experimental-identity-and-auth/src/middleware-http-signing/httpSigningMiddleware.ts
@@ -1,0 +1,57 @@
+import { HttpRequest } from "@smithy/protocol-http";
+import {
+  FinalizeHandler,
+  FinalizeHandlerArguments,
+  FinalizeHandlerOutput,
+  FinalizeRequestMiddleware,
+  HandlerExecutionContext,
+  SMITHY_CONTEXT_KEY,
+} from "@smithy/types";
+import { getSmithyContext } from "@smithy/util-middleware";
+
+import { SelectedHttpAuthScheme } from "../HttpAuthScheme";
+
+/**
+ * @internal
+ */
+interface HttpSigningMiddlewareSmithyContext extends Record<string, unknown> {
+  selectedHttpAuthScheme?: SelectedHttpAuthScheme;
+}
+
+/**
+ * @internal
+ */
+interface HttpSigningMiddlewareHandlerExecutionContext extends HandlerExecutionContext {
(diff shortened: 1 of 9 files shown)


packages/experimental-identity-and-auth/src/middleware-http-signing/httpSigningMiddleware.ts
```diff
@@ -0,0 +1,57 @@
...
+  args: FinalizeHandlerArguments<Input>
+): Promise<FinalizeHandlerOutput<Output>> => {
+  if (!HttpRequest.isInstance(args.request)) {
+    return next(args);
+  }
+
+  const smithyContext: HttpSigningMiddlewareSmithyContext = getSmithyContext(context);
+  const scheme = smithyContext.selectedHttpAuthScheme;
+  if (!scheme) {
+    throw new Error(`No HttpAuthScheme was selected: unable to sign request`);
```
kuhe, line 46: what should someone encountering this error actually do to fix it?

syall: This error is more as a precaution if the middleware was somehow used without `httpAuthSchemeMiddleware`.

`httpAuthSchemeMiddleware` (https://github.com/awslabs/smithy-typescript/pull/929) should always be code-generated along-side with `httpSigningMiddleware`. `httpAuthSchemeMiddleware` throws an error if a `HttpAuthScheme` could not be selected.

---

There should always be at least 1 valid option if codegen-supported auth schemes in `smithy-typescript` are used, e.g. `@httpBearerAuth`, `@httpApiKeyAuth`, generic `@aws.auth#sigv4`. Other custom codegen-supported auth scheme implementations can also be provided (currently through the internal `HttpAuthTypeScriptIntegration` interface, which is subject to change).

If an auth scheme was not selected, it's likely:

- At codegen time, an codegen-unsupported auth scheme (e.g. a custom auth trait) was used, or
- The client was configured with `httpAuthSchemes` or `httpAuthSchemeProvider` values that didn't provide a valid selected `HttpAuthScheme`.

These cases should be very rare as they both require customization from either the customer codegen'ing the clients, or customizing client configs different from the default values.

Customers should probably not add custom auth schemes or customize the client config values for `httpAuthSchemes` or `httpAuthSchemeProvider` without thorough testing.

AndrewFossAWS (approved):

PR #932, opened 2023-09-13
ignore __type field in unions

kuhe:
Ignore `__type` key in unions for JSON protocols. 

- [ ] TODO: only for JSON protocols


kuhe:
this should be done in JSv3, being an AWS quirk

PR #936, opened 2023-09-15
move runtime packages into deps from devDeps

kuhe:
moves devDeps starting with `@smithy` into deps. Even if imported for types, they should not be in devDeps since they are imported in the runtime code.


syall (approved):

trivikr (approved):

kuhe:
fixes https://github.com/aws/aws-sdk-js-v3/issues/4482

PR #937, opened 2023-09-15
enforce import type statements with linter

kuhe:
this PR auto-fixes import statements to use import-type when the imported elements are types.

This is quite optional. There are some edge case benefits explained here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

- [x] verified downlevel types does not use `import type`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     /** Turn off strict enforcement */
     "@typescript-eslint/ban-types": "off",
     "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/consistent-type-imports": "error",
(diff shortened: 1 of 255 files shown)


.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     /** Turn off strict enforcement */
     "@typescript-eslint/ban-types": "off",
     "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/consistent-type-imports": "error",
```
kuhe, line 16: if you want to implement this later, this line is the only required change.

Then run `yarn lint-fix` and `yarn format` and `npx changeset`.

PR #961, opened 2023-09-25
Support `aliases` for `MiddlewareStack`

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Support `aliases` for `MiddlewareStack`

`aliases` is introduced to enable refactoring of the middleware stack of
existing clients.

Instead of only considering `name` of middlewares, all entries in
`alias` are also considered for actions like adding, removing.

*Testing:*

- [x] Add tests for every relevant middleware stack command using `aliases`.

---

### DIFF
--- a/.changeset/soft-bikes-change.md
+++ b/.changeset/soft-bikes-change.md
@@ -1,0 +1,6 @@
+---
+"@smithy/middleware-stack": patch
+"@smithy/types": patch
+---
+
+Support `aliases` for `MiddlewareStack`
--- a/packages/types/src/middleware.ts
+++ b/packages/types/src/middleware.ts
@@ -284,6 +284,14 @@ export interface HandlerOptions {
   name?: string;
 
   /**
+   * @internal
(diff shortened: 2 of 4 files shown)


.changeset/soft-bikes-change.md
```diff
@@ -0,0 +1,6 @@
+---
+"@smithy/middleware-stack": patch
+"@smithy/types": patch
```
syall, line 3: Should these be `minor` bumps?

packages/types/src/middleware.ts
```diff
@@ -283,6 +283,12 @@ export interface HandlerOptions {
    */
   name?: string;
 
+  /**
+   * @internal
+   * Aliases for the middleware
```
kuhe, line 288: expand the explanation: 

aliases allow a middleware to be found by multiple names, used for example in auth refactor to allow older references to the existing middleware names to continue working.

kuhe (approved):

AndrewFossAWS (approved):

PR #975, opened 2023-09-28
fix unit test endpoints model

kuhe:
- removes the `isSet` check on a required parameter
- imports `smithy-aws-endpoints` because the test uses the AWS region builtin


kstich:
Closing in favor of #976

PR #986, opened 2023-10-05
Read values from main section when parsing INI files

trivikr:
*Issue #, if available:*
Refs: https://github.com/awslabs/smithy-typescript/issues/985#issuecomment-1749399264

*Description of changes:*
Reads values from main settings when parsing INI files

---

### DIFF
--- a/packages/shared-ini-file-loader/src/parseIni.ts
+++ b/packages/shared-ini-file-loader/src/parseIni.ts
@@ -18,22 +18,13 @@ export const parseIni = (iniData: string): ParsedIniData => {
         throw new Error(`Found invalid profile name "${currentSection}"`);
       }
     } else if (currentSection) {
-      const [key, value] = line.split("=").map((item) => item.trim());
+      const [name, value] = line.split("=").map((item) => item.trim());
       if (value === "") {
-        currentSubSection = key;
+        currentSubSection = name;
       }
       if (currentSubSection === undefined) {
         // ToDo: populate subsection in future PR, when IniSection is updated to support subsections.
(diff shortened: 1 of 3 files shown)


packages/shared-ini-file-loader/src/parseIni.ts
```diff
@@ -4,29 +4,32 @@ const profileNameBlockList = ["__proto__", "profile __proto__"];
...
-      const isAssignment: boolean =
-        indexOfEqualsSign !== -1 && indexOfEqualsSign !== start && indexOfEqualsSign !== end;
-      if (isAssignment) {
-        const [name, value]: [string, string] = [
-          line.substring(0, indexOfEqualsSign).trim(),
-          line.substring(indexOfEqualsSign + 1).trim(),
-        ];
-        map[currentSection] = map[currentSection] || {};
-        map[currentSection][name] = value;
+      const [name, value] = line.split("=").map((item) => item.trim());
```
kuhe, line 21: can't some tokens contain `=`?

should we keep the split on the first `=`?

AndrewFossAWS (approved):

packages/shared-ini-file-loader/src/parseIni.ts
trivikr: I'm not sure of examples where values contain `=`.

But since it's currently handled, I fixed it and added a test case [`3d5b7c6` (#986)](https://github.com/awslabs/smithy-typescript/pull/986/commits/3d5b7c69f0e76cdeac4376c3dfa1c8dca845054c)

kuhe (approved):

PR #988, opened 2023-10-05
Populate subsections when parsing INI files

trivikr:
*Issue #, if available:*
Fixes: https://github.com/awslabs/smithy-typescript/issues/985

*Description of changes:*
Populates subsections when parsing INI files

---

### DIFF
--- a/.changeset/fast-cows-dream.md
+++ b/.changeset/fast-cows-dream.md
@@ -1,0 +1,5 @@
+---
+"@smithy/types": minor
+---
+
+Allow subsections in IniSection type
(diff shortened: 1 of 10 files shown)


trivikr:
This would be a breaking change if someone has written internal components to process config, and haven't checked type to be string. It would require doing a major version bump for types when IniSection key is updated to include `Record<string, string>`

Instead, I'll use dot separator to populate subsections.

For the following file:
```console
[profile dev]
s3 =
  max_concurrent_requests = 20
```

The processed Ini file will be:
```json
{
  "dev": {
    "s3.max_concurrent_requests": 20
  }
}
```

The dot separator is already used by AWS CLI when populating subsections.
Reference: https://docs.aws.amazon.com/cli/latest/topic/s3-config.html

PR #1014, opened 2023-10-09
Read service specific endpoints from env/config

trivikr:
*Issue #, if available:*
Internal JS-4281

*Description of changes:*
Reads service specific endpoints from environment or configuration

---

### DIFF
--- a/packages/middleware-endpoint/package.json
+++ b/packages/middleware-endpoint/package.json
@@ -24,6 +24,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@smithy/middleware-serde": "workspace:^",
+    "@smithy/node-config-provider": "workspace:^",
     "@smithy/types": "workspace:^",
     "@smithy/url-parser": "workspace:^",
     "@smithy/util-middleware": "workspace:^",
--- a/packages/middleware-endpoint/src/adaptors/getEndpointUrlConfig.spec.ts
+++ b/packages/middleware-endpoint/src/adaptors/getEndpointUrlConfig.spec.ts
@@ -1,0 +1,63 @@
+import { getEndpointUrlConfig } from "./getEndpointUrlConfig";
+
+const ENV_ENDPOINT_URL = "AWS_ENDPOINT_URL";
+const CONFIG_ENDPOINT_URL = "endpoint_url";
+
+describe(getEndpointUrlConfig.name, () => {
+  const serviceId = "foo";
+  const endpointUrlConfig = getEndpointUrlConfig(serviceId);
+
+  const mockEndpoint = "https://mock-endpoint.com";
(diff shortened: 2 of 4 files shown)


kuhe (approved):

packages/middleware-endpoint/src/adaptors/getEndpointFromConfig.ts
```diff
@@ -0,0 +1,5 @@
+import { loadConfig } from "@smithy/node-config-provider";
+
+import { getEndpointUrlConfig } from "./getEndpointUrlConfig";
+
+export const getEndpointFromConfig = async (serviceId: string) => loadConfig(getEndpointUrlConfig(serviceId))();
```
syall, line 5: I think `loadConfig` needs an `await` since it returns a `Provider`?

Or if nothing needs `await`, this could be a regular function.

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
```diff
@@ -36,6 +38,13 @@ export const getEndpointFromInstructions = async <
   clientConfig: Partial<EndpointResolvedConfig<T>> & Config,
   context?: HandlerExecutionContext
 ): Promise<EndpointV2> => {
+  if (!clientConfig.endpoint) {
+    const endpointFromConfig = await getEndpointFromConfig(clientConfig.serviceId || "");
+    if (endpointFromConfig) {
+      clientConfig.endpoint = () => Promise.resolve(toEndpointV1(endpointFromConfig));
+    }
+  }
```
syall, line 46: Would checking the config files on each operation invocation be expensive for I/O?

Maybe this should be in a `resolve*Config` function since it shouldn't change between calls.

syall (changes requested):
Would adding `@internal` to every export make sense? I don't think customers should use these directly.

packages/middleware-endpoint/src/adaptors/getEndpointFromConfig.ts
trivikr: The `loadConfig` does not need await. The callee `getEndpointFromInstructions` awaits it.

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
trivikr: No. We cache the contents of the config file in https://github.com/awslabs/smithy-typescript/blob/main/packages/shared-ini-file-loader/src/slurpFile.spec.ts, and read operation is called only once per file.

trivikr:
> Would adding @internal to every export make sense?

None of the utilities added in this PR are exported at top level. We don't need to add `@internal`.

PR #1028, opened 2023-10-12
feat: remove "| string" and "| number" from enum targeted members

kuhe:
This PR removes the open-ended union of `| string` and `| number` from shape members that targeted enums. 

This improves the dev experience with the type system by 
- flagging enum value typos
- potentially improving auto-complete in certain code editors
- enabling IDE-generation or manual writing of exhaustive switch statements

User guide:
- if using only JavaScript, you would not be affected by this change.
- If using existing declared enums, no change is expected.
- If using enums *not* declared in the generated model, such as when using an older version of a generated SDK, use the following overrides:
  - on inputs, use `'NEW_VALUE' as any`.
  - on outputs, use e.g. `if (output.enum as any === 'NEW_VALUE') {`
  - in either case, you can also use a compiler comment to ignore any unenumerated fields
- you should still write a default branch, if only to throw an error, in any switch statement on an enum value.


kuhe:
> @github-actions
ci / Ensure TypeScript is formatted (pull_request) 

there are no typescript changes in this PR

srchase (approved):

syall (approved):

siddsriv (approved):

PR #1034, opened 2023-10-13
fix(experimentalIdentityAndAuth): check effective auth schemes equality including order

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Check effective auth schemes equality including order.

---

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
@@ -18,6 +18,7 @@ import software.amazon.smithy.model.knowledge.ServiceIndex;
 import software.amazon.smithy.model.knowledge.ServiceIndex.AuthSchemeMode;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.typescript.codegen.CodegenUtils;
 import software.amazon.smithy.typescript.codegen.ConfigField;
 import software.amazon.smithy.typescript.codegen.Dependency;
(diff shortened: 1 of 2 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
```diff
@@ -149,4 +150,23 @@ public static Map<String, HttpAuthSchemeParameter> collectHttpAuthSchemeParamete
...
+        if (httpAuthSchemes1.size() != httpAuthSchemes2.size()) {
+            return false;
+        }
+        var iter1 = httpAuthSchemes1.entrySet().iterator();
+        var iter2 = httpAuthSchemes2.entrySet().iterator();
+        while (iter1.hasNext() && iter2.hasNext()) {
+            var entry1 = iter1.next();
+            var entry2 = iter2.next();
+            if (!entry1.equals(entry2)) {
+                return false;
```
kuhe, line 167: if order is important the method interfaces involved or leading up to the point of comparison should indicate a return type of the actual ordered map implementation instead of `Map`. 

For example

```java
Map<ShapeId, Trait> getEffectiveAuthSchemes
```

should become

```java
LinkedHashMap<ShapeId, Trait> getEffectiveAuthSchemes
```

kuhe (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/auth/AuthUtils.java
syall: The `Map`s returned in Smithy are not more specifically typed so we can't assume a more specific type.

Typically, `Map` equality in Java only checks if all of the items in 1 map is in the other and vice versa, and not order, so this explicit iterator check (albeit kind of a black box) is necessary.

PR #1037, opened 2023-10-13
feat(enums): generate enum Record keys when target is enum

kuhe:
due to the closure of enums in https://github.com/awslabs/smithy-typescript/pull/1028,

Map deserializers for Maps where the key is an enum need an update to be compatible. 

```ts
prop: Record<Enum | string, Value>
```

becomes

```ts
prop: Record<Enum, Value>
```


kuhe:
linked: https://github.com/aws/aws-sdk-js-v3/pull/5356

trivikr (approved):

srchase (approved):

PR #1046, opened 2023-10-18
transform inputs for platform specific type helpers

kuhe:
there is a type helper that narrows payload types, currently only for outputs

```ts
import { NodeJsClient } from "@smithy/types";

const s3 = new S3({}) as NodeJsClient<S3>;

await s3.putObject({
    Bucket: '',
    Key: '',
    Body: new Blob() // helper makes this a type error, must use Node.js http(s) payload types like Buffer/Uint8Array/string
})

const get = await s3.getObject({
    Bucket: '',
    Key: ''
});

const body = get.Body; // helper narrows this to SdkStream<IncomingMessage>
```

this PR adds type narrowing to inputs as well.

### DIFF
--- a/.changeset/polite-flies-explode.md
+++ b/.changeset/polite-flies-explode.md
@@ -1,0 +1,5 @@
+---
+"@smithy/types": minor
+---
+
+transform inputs for env specific type helpers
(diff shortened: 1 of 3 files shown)


kuhe:
closes https://github.com/aws/aws-sdk-js-v3/issues/5361

srchase (approved):

siddsriv (approved):

PR #1054, opened 2023-10-24
allow TypeScriptIntegration to write prior to the config object literal

kuhe:
allow a typescript integration to write content outside of the runtimeConfig object literal as part of the config setup.


kuhe:
- [x] there are no TS package changes and `yarn test` passes locally

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeConfigGenerator.java
```diff
@@ -169,11 +169,17 @@ void generate(LanguageTarget target) {
...
+                .replace("$${customizations}", "${L@customizations}")
+                .replace("$${prepareCustomizations}", "${L@prepareCustomizations}");
 
         delegator.useFileWriter(target.getTargetFilename(), writer -> {
             // Inject customizations into the ~template.
+            writer.onSection("prepareCustomizations", original -> {
+                for (TypeScriptIntegration integration : integrations) {
+                    integration.prepareCustomizations(writer, target, settings, model);
+                }
+            });
```
syall, line 182: Not in the scope of this PR, but I feel we should move away from using templates and instead migrate to writers and `CodeSection`s.

By using `DirectedCodegen`, `CodeInterceptor`s can be used. This would allow state to be accessible to the interceptors (e.g. `target`) and avoid adding a new integration point in `TypeScriptIntegration`, e.g.

```java
public record PrepareRuntimeConfigCodeSection(LanguageTarget target) implements CodeSection {}

public final class ExamplePrepareRuntimeConfigCodeInterceptorIntegration implements TypeScriptIntegration {
    @Override
    public List<? extends CodeInterceptor<? extends CodeSection, TypeScriptWriter>> interceptors(
        TypeScriptCodegenContext codegenContext) {
        return List.of(CodeInterceptor.appender(PrepareRuntimeConfigCodeSection.class, (w, s) -> {
            // ...
        }));
    }
}
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/TypeScriptIntegration.java
```diff
@@ -253,4 +253,16 @@ default Map<String, Consumer<TypeScriptWriter>> getRuntimeConfigWriters(
     default List<ExtensionConfigurationInterface> getExtensionConfigurationInterfaces() {
         return Collections.emptyList();
     }
+
+    /**
+     * Allows the customization to write arbitrary preparatory code prior to the returned config object.
+     */
+    default void prepareCustomizations(
```
syall, line 260: I think annotating this method with `@SmithyInternalApi` for now would be good.

syall (changes requested):

syall (approved):

trivikr (approved):

PR #1059, opened 2023-11-02
feat(credential-provider-imds): add flag for blocking imds v1 fallback behavior

kuhe:
this is an AWS-specific feature but the code package is housed here in Smithy-TS at least for the time being.

this PR adds 3 locations from which to shut off the IMDS v1 fallback behavior:
- runtime code initialization of the IMDS or upstream chained credential providers
- AWS config file profile setting
- process environment variable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ wrapper/
 
 # local scripting
 Makefile
+workspace
 
 # Visual Studio Code
 .vscode/*
--- a/packages/credential-provider-imds/src/fromInstanceMetadata.ts
+++ b/packages/credential-provider-imds/src/fromInstanceMetadata.ts
@@ -116,9 +116,6 @@ const getInstanceImdsProvider = (init: RemoteProviderInit) => {
       let token: string;
       try {
         token = (await getMetadataToken({ ...endpoint, timeout })).toString();
-        throw {
-          statusCode: 404,
-        };
       } catch (error) {
         if (error?.statusCode === 400) {
           throw Object.assign(error, {
--- a/packages/credential-provider-imds/src/fromInstanceMetadata.ts
+++ b/packages/credential-provider-imds/src/fromInstanceMetadata.ts
@@ -1,5 +1,5 @@
+import { loadConfig } from "@smithy/node-config-provider";
 import { CredentialsProviderError } from "@smithy/property-provider";
-import { parseKnownFiles } from "@smithy/shared-ini-file-loader";
 import { AwsCredentialIdentity, Provider } from "@smithy/types";
 import { RequestOptions } from "http";
 
@@ -15,7 +15,7 @@ import { staticStabilityProvider } from "./utils/staticStabilityProvider";
 const IMDS_PATH = "/latest/meta-data/iam/security-credentials/";
 const IMDS_TOKEN_PATH = "/latest/api/token";
 const AWS_EC2_METADATA_V1_DISABLED = "AWS_EC2_METADATA_V1_DISABLED";
-const AWS_PROFILE = "AWS_PROFILE";
+const PROFILE_AWS_EC2_METADATA_V1_DISABLED = "ec2_metadata_v1_disabled";
 
 /**
  * @internal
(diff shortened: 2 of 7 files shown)


packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -25,10 +29,35 @@ export const fromInstanceMetadata = (init: RemoteProviderInit = {}): Provider<In
...
-    const profile = (
+    const isImdsV1Fallback = disableFetchToken || options.headers?.["x-aws-ec2-metadata-token"] == null;
+
+    if (isImdsV1Fallback) {
+      const configFiles = await parseKnownFiles({});
+      const fallbackBlockedFromProfile =
+        configFiles[profile]?.ec2_metadata_v1_disabled &&
+        String(configFiles[profile]?.ec2_metadata_v1_disabled).toLowerCase() !== "false";
+      const fallbackBlockedFromProcessEnv =
+        process.env[AWS_EC2_METADATA_V1_DISABLED] && process.env[AWS_EC2_METADATA_V1_DISABLED] !== "false";
```
trivikr, line 44: Use configLoader from `node-config-provider`?
It memoizes the result. We just need to pass selectors.

https://github.com/awslabs/smithy-typescript/blob/39219d7d2377ed43f4fa9d827b9346e8a36a3a57/packages/node-config-provider/src/configLoader.ts

trivikr (approved):

.gitignore
```diff
@@ -25,6 +25,7 @@ wrapper/
 
 # local scripting
 Makefile
+workspace
```
syall, line 28: nit: what does `workspace` refer to?

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -25,10 +29,56 @@ export const fromInstanceMetadata = (init: RemoteProviderInit = {}): Provider<In
 const getInstanceImdsProvider = (init: RemoteProviderInit) => {
   // when set to true, metadata service will not fetch token
   let disableFetchToken = false;
+  const { logger, profile } = init;
   const { timeout, maxRetries } = providerConfigFromInit(init);
 
   const getCredentials = async (maxRetries: number, options: RequestOptions) => {
-    const profile = (
+    const isImdsV1Fallback = disableFetchToken || options.headers?.["x-aws-ec2-metadata-token"] == null;
```
syall, line 36: nit: move the header `x-aws-ec2-metadata-token` to a constant

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -25,10 +29,56 @@ export const fromInstanceMetadata = (init: RemoteProviderInit = {}): Provider<In
...
+          profile,
+        }
+      )();
+
+      if (init.ec2MetadataV1Disabled || configValue) {
+        const causes: string[] = [];
+        if (init.ec2MetadataV1Disabled)
+          causes.push("credential provider initialization (runtime option ec2MetadataV1Disabled)");
+        if (fallbackBlockedFromProfile) causes.push("config file profile (ec2_metadata_v1_disabled)");
+        if (fallbackBlockedFromProcessEnv) causes.push("process environment variable (AWS_EC2_METADATA_V1_DISABLED)");
```
syall, line 71: nit: interpolate using the constants defined

packages/credential-provider-imds/src/utils/staticStabilityProvider.ts
```diff
@@ -32,7 +33,7 @@ export const staticStabilityProvider = (
         credentials = getExtendedInstanceMetadataCredentials(credentials, logger);
       }
     } catch (e) {
-      if (pastCredentials) {
+      if (pastCredentials && !(e instanceof InstanceMetadataV1FallbackError)) {
```
syall, line 36: Clarifying question: so if the error is `InstanceMetadataV1FallbackError`, it shouldn't try to extend the credentials?

syall (approved):

.gitignore
kuhe: For ad hoc testing scripts. Scripts run from the context of the monorepo workspace automatically import the packages in the repo, so it's for convenience.

packages/credential-provider-imds/src/utils/staticStabilityProvider.ts
kuhe: yeah, it should still extend, so I'll remove this condition

.gitignore
siddsriv: we should be able to remove this bit I think, it usually refers to the dir where we do some local testing/experiments

PR #1066, opened 2023-11-08
add script for enforcing pkg json standards

kuhe:
This adds a pre-commit hook to audit the pkg json bundler directives.

Specifically, we don't know if a bundler is going to use the CJS or ES entrypoint, and thus the file replacement directives should account for either case.

### DIFF
--- a/packages/hash-blob-browser/package.json
+++ b/packages/hash-blob-browser/package.json
@@ -55,6 +55,9 @@
   "react-native": {
     "@smithy/chunked-blob-reader": "@smithy/chunked-blob-reader-native"
   },
+  "browser": {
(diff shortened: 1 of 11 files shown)


syall (approved):

packages/hash-blob-browser/package.json
```diff
@@ -55,6 +55,9 @@
   "react-native": {
     "@smithy/chunked-blob-reader": "@smithy/chunked-blob-reader-native"
   },
+  "browser": {
+    "@smithy/chunked-blob-reader": "@smithy/chunked-blob-reader"
```
kuhe, line 59: this is redundant, but much easier to validate

trivikr (approved):

PR #1080, opened 2023-11-21
feat(experimentalIdentityAndAuth): add internal error and success handlers to `HttpSigner` and QOL changes

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Add internal error and success handlers to `HttpSigner` for `experimentalIdentityAndAuth`.

Also add some quality of life improvements:

- Log error message per test client for build failures
- Copy `getSmithyContext()` and `normalizeProvider()` to `@smithy/core`

---

### DIFF
--- a/packages/types/src/auth/HttpSigner.ts
+++ b/packages/types/src/auth/HttpSigner.ts
@@ -1,5 +1,19 @@
-import { HttpRequest } from "../http";
+import { HttpRequest, HttpResponse } from "../http";
 import { Identity } from "../identity/identity";
+
+/**
+ * @internal
+ */
+export interface ErrorHandler {
+  (signingProperties: Record<string, unknown>): <E extends Error>(error: E) => never;
+}
+
+/**
+ * @internal
+ */
+export interface SuccessHandler {
+  (httpResponse: HttpResponse | unknown, signingProperties: Record<string, unknown>): void;
+}
 
 /**
  * Interface to sign identity and signing properties.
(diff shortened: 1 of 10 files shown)


packages/types/src/auth/HttpSigner.ts
```diff
@@ -1,6 +1,20 @@
...
+ */
+export interface ErrorHandler {
+  (signingProperties: Record<string, unknown>): <E extends Error>(error: E) => never;
+}
+
+/**
+ * @internal
+ */
+export interface SuccessHandler {
+  (httpResponse: HttpResponse | unknown, signingProperties: Record<string, unknown>): void;
```
kuhe, line 15: signingProperties should have a type. It can be a record too (`&`), but it should also indicate that it tends to have properties like `name`, `in`, etc.

packages/types/src/auth/HttpSigner.ts
```diff
@@ -1,6 +1,20 @@
-import { HttpRequest } from "../http";
+import { HttpRequest, HttpResponse } from "../http";
 import { Identity } from "../identity/identity";
 
+/**
+ * @internal
+ */
+export interface ErrorHandler {
+  (signingProperties: Record<string, unknown>): <E extends Error>(error: E) => never;
```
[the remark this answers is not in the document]
syall: The error handler is expected to throw the error it receives, so it would be `never`.

I can expand it to `never | void`, but I think for now it's better to constrain the behavior to expect throwing the error.

I can add a line about it in the documentation.

AndrewFossAWS (approved):

packages/types/src/auth/HttpSigner.ts
syall: This is the generic `HttpSigner` which has no context for the properties of any specific `HttpAuthScheme`.

I think concrete implementations of `HttpSigner` can provide more type hints, but I can add that in a follow-up PR.

kuhe (approved):

PR #1089, opened 2023-11-30
feat: add requestHandler ctor pass through type

kuhe:
Currently, to override any options in the default request handler, the requestHandler class must be instantiated.

This adds a type representing default request handler ctor params, 
intended to allow the following syntactic sugar:
```js
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
import https from "https";    

var agent = new https.Agent({
  maxSockets: 25
});

var dynamodbClient = new DynamoDBClient({
  requestHandler: new NodeHttpHandler({
    httpsAgent: agent
 })
});
```

the PR's ultimate goal is to allow the following:

```js
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
import https from "https";    

var agent = new https.Agent({
  maxSockets: 25
});

var dynamodbClient = new DynamoDBClient({
  requestHandler: {
    httpsAgent: agent
  }
});
```

This may appear to be a very mild simplification, but it avoids user errors like accidentally installing a locked version of `@smithy/node-http-handler` or needing to differentiate between the previous `@aws-sdk/node-http-handler` and its migrated name.


packages/protocol-http/package.json
```diff
@@ -16,6 +16,13 @@
   "main": "./dist-cjs/index.js",
   "module": "./dist-es/index.js",
   "types": "./dist-types/index.d.ts",
+  "browser": {
+    "./dist-es/httpHandlerInitialization": "./dist-es/httpHandlerInitialization.browser"
+  },
```
syall, line 21: does this need a `dist-cjs` entry point?

packages/protocol-http/src/httpHandlerInitialization.browser.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Alternate compilation for browser excluding http Agent types.
+ *
+ * @internal
+ */
+export type DefaultRequestHandlerInitializationPassThroughParameters = FetchHttpHandlerOptions;
```
syall, line 6: should these types and interfaces live in `@smithy/types`, and the `@smithy/*-http-handler` packages import them so there is no duplication?

packages/protocol-http/src/httpHandlerInitialization.ts
```diff
@@ -0,0 +1,35 @@
...
+ * This type represents an alternate client constructor option for the entry
+ * "requestHandler". Instead of providing an instance of a requestHandler, the user
+ * may provide the requestHandler's constructor options for either the
+ * NodeHttpHandler or FetchHttpHandler.
+ *
+ * For other RequestHandlers, constructor parameter passthrough is not available.
+ *
+ * @public
+ */
+export type RequestHandlerParams = NodeHttpHandlerOptions | FetchHttpHandlerOptions;
```
syall, line 15: should these types and interfaces live in `@smithy/types`, and the `@smithy/*-http-handler` packages import them so there is no duplication?

packages/protocol-http/src/httpHandlerInitialization.browser.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Alternate compilation for browser excluding http Agent types.
+ *
+ * @internal
```
syall, line 4: the browser one is marked as `@internal`, but the Node.js export is marked as `@public`

syall (changes requested):

packages/protocol-http/package.json
kuhe: no, I'm going to remove this

packages/protocol-http/src/httpHandlerInitialization.ts
kuhe: yes, let's do that

packages/protocol-http/src/httpHandlerInitialization.browser.ts
kuhe: removed the extra browser file after the move to `@smithy/types`

syall (approved):

PR #1092, opened 2023-12-01
fix: warn streaming requests are not retryable

kuhe:
Current behavior when retrying a streaming payload:
- client hangs, the stream has been read for the initial request and cannot be read again for the retry

In this PR:
- a warning is logged: "An error was encountered in a non-retryable streaming request."
- the error is thrown regardless of its retryable status

----
How to reproduce streaming retry:

```js
import { Readable } from "stream";

function streamToString(stream) {
  const chunks = [];
  return new Promise((resolve, reject) => {
    stream.on("data", (chunk) => chunks.push(Buffer.from(chunk)));
    stream.on("error", (err) => reject(err));
    stream.on("end", () => resolve(Buffer.concat(chunks).toString("utf8")));
  });
}

const s3 = new S3({
  region: "us-west-2",
  requestHandler: new (class {
    constructor() {
      this.call = 0;
      this.handler = new NodeHttpHandler({});
    }

    async handle(req) {
      if (this.call++ === 0) {
        const body = await streamToString(req.body);
        console.log("mock error has read the stream body", body);
        throw {
          $metadata: {
            httpStatusCode: 500,
          },
        };
      }
      return this.handler.handle(req);
    }
  })(),
});

const Bucket = "....";

await s3.putObject({
  Bucket,
  Key: "retry-stream",
  Body: Readable.from("abcd"),
  ContentLength: 4,
});

```

This PR prevents retries for streaming payloads because we cannot restore the stream.
A warning is also logged prior to the error.

### DIFF
--- a/packages/middleware-retry/src/isStreamingPayload/isStreamingPayload.ts
+++ b/packages/middleware-retry/src/isStreamingPayload/isStreamingPayload.ts
@@ -4,6 +4,6 @@ import { Readable } from "stream";
 /**
  * @internal
  */
-export const isStreamingPayload = (request: HttpRequest): boolean => 
+export const isStreamingPayload = (request: HttpRequest): boolean =>
   request?.body instanceof Readable ||
   (typeof ReadableStream !== "undefined" && request?.body instanceof ReadableStream);


packages/middleware-retry/src/isStreamingPayload/isStreamingPayload.browser.ts
```diff
@@ -0,0 +1,8 @@
+import type { HttpRequest } from "@smithy/protocol-http";
+
+/**
+ * @internal
+ */
+export const isStreamingPayload = (request: HttpRequest): boolean => {
+  return request?.body instanceof ReadableStream;
+};
```
trivikr, line 8: nit: brackets can be removed
```suggestion
export const isStreamingPayload = (request: HttpRequest): boolean => request?.body instanceof ReadableStream;
```

packages/middleware-retry/src/isStreamingPayload/isStreamingPayload.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * @internal
+ */
+export const isStreamingPayload = (request: HttpRequest): boolean => {
+  return (
+    request?.body instanceof Readable ||
+    (typeof ReadableStream !== "undefined" && request?.body instanceof ReadableStream)
+  );
+};
```
trivikr, line 12: nit: brackets can be removed
```suggestion
export const isStreamingPayload = (request: HttpRequest): boolean => 
  request?.body instanceof Readable ||
  (typeof ReadableStream !== "undefined" && request?.body instanceof ReadableStream);
```

packages/middleware-retry/src/retryMiddleware.ts
```diff
@@ -20,6 +21,7 @@ import { INVOCATION_ID_HEADER, REQUEST_HEADER } from "@smithy/util-retry";
 import { v4 } from "uuid";
 
 import { RetryResolvedConfig } from "./configurations";
+import { isStreamingPayload } from "./isStreamingPayload/isStreamingPayload";
```
trivikr, line 24: Rename `isStreamingPayload/isStreamingPayload.ts` to `isStreamingPayload/index.ts` so that it can be directly imported?

trivikr (approved):

packages/middleware-retry/src/retryMiddleware.ts
kuhe: I usually open files by typing in their name, so naming things index makes it hard for me to find

vvo:
@kuhe We're getting these warnings for unrelated errors, for example when trying to pass an UploadId that doesn't exist in an UploadPartCommand. I wonder if always sending the stream warning on any error is what we really want here?

PR #1108, opened 2023-12-11
feat: xml serde reduction

kuhe:
This PR implements a class called `StringStore`, which allocates shortened variables for any given string.

The basic idea is a transform of something like:
```ts
const body = '<?xml version="1.0" encoding="UTF-8"?>';
...
const body = '<?xml version="1.0" encoding="UTF-8"?>';
...
const body = '<?xml version="1.0" encoding="UTF-8"?>';
// in multiple functions
```
to
```ts
const body = _ve;
...
const body = _ve;
...
const body = _ve;
...
// usages only appear in functions so it's ok to define this at the bottom of the file.
const _v2 = '<?xml version="1.0" encoding="UTF-8"?>';
```

Real example from S3:
```ts
const headers = map({}, isSerializableHeaderValue, {
    "content-type": "application/xml",
    "x-amz-acl": input.ACL,
    "content-md5": input.ContentMD5,
    "x-amz-sdk-checksum-algorithm": input.ChecksumAlgorithm,
    "x-amz-grant-full-control": input.GrantFullControl,
    "x-amz-grant-read": input.GrantRead,
    "x-amz-grant-read-acp": input.GrantReadACP,
    "x-amz-grant-write": input.GrantWrite,
    "x-amz-grant-write-acp": input.GrantWriteACP,
    "x-amz-expected-bucket-owner": input.ExpectedBucketOwner,
});
```

```ts
const headers: any = map({}, isSerializableHeaderValue, {
    "content-type": "application/xml",
    [_xaa]: input[_ACL],
    [_cm]: input[_CMD],
    [_xasca]: input[_CA],
    [_xagfc]: input[_GFC],
    [_xagr]: input[_GR],
    [_xagra]: input[_GRACP],
    [_xagw]: input[_GW],
    [_xagwa]: input[_GWACP],
    [_xaebo]: input[_EBO],
});
// variables are defined at the end
```

To answer some potential questions:

- doesn't this make the code hard to read?
  - Yes. At least, you can hover over the variables to see what they are 😢. I believe users would prefer the reduced code size over readability of the protocol files.

- is the variable allocation deterministic?
  - Yes. Variables are named with a preference for the word initials of the given string. This makes the variables slightly longer than they need to be, but reduces variable thrashing with model updates.
 
- does this affect tree-shaking?
  - no. I tested with `esbuild` and only those variables being used are included in the bundle.

- what is the effect on protocol file sizes?
  - **Note**: this is not the size of the client, it's only the protocol file
  -  S3 (plus other optimizations from https://github.com/aws/aws-sdk-js-v3/pull/5566) 412kb on disk to 328kb
  -  EC2, 2.67mb to 2.5mb (there's room for further optimization)

- what about strings that only appear once?
  - within the codegen program's current framework, it is not known when allocating a variable how many times it will be used. It would require a second pass to inline variables with only 1 usage. This change increases char-count for strings that only appear once. However, in practice (spot-checking https://github.com/aws/aws-sdk-js-v3/pull/5566) most strings appear multiple times.


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -673,7 +677,9 @@ private void generateOperationRequestSerializer(
             // Get the hostname, path, port, and scheme from client's resolved endpoint.
             // Then construct the request from them. The client's resolved endpoint can
             // be default one or supplied by users.
-            writer.write("const {hostname, protocol = $S, port, path: basePath} = await context.endpoint();", "https");
+
+            writer.addImport("requestBuilder", "rb", TypeScriptDependency.SMITHY_CORE);
```
kuhe, line 681: requestBuilder is from https://github.com/smithy-lang/smithy-typescript/pull/1107

.changeset/selfish-pens-scream.md
mtdowling, line 1: What is this?

kuhe: this is a no-op file to pass the status check that normally requires any typescript package change to have a changeset entry for publishing purposes

kuhe: generated with `npx changeset --empty`

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+        String[] sections = literal.split("[-_\\s]");
+        StringBuilder v = new StringBuilder("_");
+        Queue<Character> deconfliction = new LinkedList<>();
+        if (sections.length > 1) {
+            Arrays.stream(sections)
+                .map(s -> s.charAt(0))
+                .filter(this::isAllowedChar)
+                .forEach(v::append);
+        } else {
+            for (int i = 0; i < literal.length(); ++i) {
```
mtdowling, line 96: is this not skipping the first character because you used `++i` rather than `i++`?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Assigns a unique variable using the letters from the literal.
+     * Prefers the uppercase or word-starting letters.
+     */
+    private String allocateVariable(String literal) {
+        String[] sections = literal.split("[-_\\s]");
+        StringBuilder v = new StringBuilder("_");
+        Queue<Character> deconfliction = new LinkedList<>();
+        if (sections.length > 1) {
+            Arrays.stream(sections)
```
mtdowling, line 91: you should consider using a for loop here to simplify and this is kind of a hot path:
```
for (String s : sections) {
    char c = s.charAt(0);
    if (isAllowedChar(c)) {
        v.append(c);
    }
}
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Assigns a new variable or returns the existing variable for a given string literal.
+     */
+    private String assignKey(String literal) {
+        if (literalToVariable.containsKey(literal)) {
+            return literalToVariable.get(literal);
+        }
+        String variable = allocateVariable(literal);
+        variableToLiteral.put(variable, literal);
+        literalToVariable.put(literal, variable);
```
mtdowling, line 78: Since you'd use computeIfAbsent, remove this assignment too

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Intended for use at the
+ * {@link software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator.GenerationContext}
+ * level, this class allocates and tracks variables assigned to string literals, allowing a
+ * form of compression on long protocol serde files.
+ */
+public class StringStore {
+    private final ConcurrentHashMap<String, String> literalToVariable = new ConcurrentHashMap<>();
```
mtdowling, line 24: since this class isn't actually thread safe at the moment, you could just use a TreeMap here directly, and a HashMap below.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Intended for use at the
+ * {@link software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator.GenerationContext}
+ * level, this class allocates and tracks variables assigned to string literals, allowing a
+ * form of compression on long protocol serde files.
+ */
+public class StringStore {
```
mtdowling, line 23: Maybe mark as internal

mtdowling (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
kuhe: the 3rd part of the for is not evaluated until the iteration completes, so it shouldn't make a difference. 

this is an old habit from hearing that pre-increment is (imperceptibly) faster than post-increment 😅

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,131 @@
...
+        writelog.addAll(incrementalKeys);
+
+        return sourceCode.toString();
+    }
+
+    /**
+     * Assigns a new variable or returns the existing variable for a given string literal.
+     */
+    private String assignKey(String literal) {
+        if (literalToVariable.containsKey(literal)) {
```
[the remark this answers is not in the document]
kuhe: Can I throw here? I want to be sure to avoid accidental allocation of two variables to the same literal.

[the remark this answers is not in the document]
mtdowling: No need. It should only ever be called if the mapping doesn't already exist.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
mtdowling: if it doesn't matter either way, I'd recommend ++i to not trip up people in the future reading this code. I'd assume it's important to start at 1 here, but it doesn't seem to be the case.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/util/StringStore.java
```diff
@@ -0,0 +1,130 @@
...
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Intended for use at the
+ * {@link software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator.GenerationContext}
+ * level, this class allocates and tracks variables assigned to string literals, allowing a
+ * form of compression on long protocol serde files.
+ */
+@SmithyInternalApi
+public class StringStore {
```
mtdowling, line 26:
```suggestion
public final class StringStore {
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -879,36 +877,41 @@ private void writeRequestQueryParam(
...
-            "input." + memberName + memberAssertionComponent,
+            "input[" + context.getStringStore().var(memberName) + "]" + memberAssertionComponent,
             binding.getMember(),
             target
         );
 
         writer.addImport("expectNonNull", "__expectNonNull", TypeScriptDependency.AWS_SMITHY_CLIENT);
 
-        if (Objects.equals("input." + memberName + memberAssertionComponent, queryValue)) {
+        if (
```
mtdowling, line 887: nit - this if statement is getting kind of hard to read. Maybe break this up by storing the predicate string as a variable first.

mtdowling (changes requested):

mtdowling (approved):

syall (approved):

trivikr (approved):

PR #1110, opened 2023-12-12
fix(json-rpc): merge initial-response event into operation response

kuhe:
This adds an eventStream serde helper for checking the first event of an event stream. 

- if the type is `initial-response`, then it is merged into the operation output.
- if the type is not `initial-response`, then it is re-emplaced onto the iterator given to the user.

Additionally, for eventStream deser functions, deserialization of the implicit `initial-response` event is added to the union to avoid ignoring that unmodeled union member.

Note: this behavior is specific to AWS. It should have no effect on JSON eventstreams that do not implement `initial-response`, but there are also no hooks available to override this behavior in AWS SDK TypeScript codegen at the moment.

If non-AWS users want to have members other than the streaming member for the output, this could also be of use to them.

### DIFF
--- a/.changeset/shaggy-owls-happen.md
+++ b/.changeset/shaggy-owls-happen.md
@@ -1,0 +1,6 @@
+---
+"@smithy/eventstream-serde-universal": patch
+"@smithy/eventstream-codec": patch
+---
+
+merge initial-response event into operation output
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ smithy-typescript-integ-tests/yarn.lock
 
 # Issue https://github.com/awslabs/smithy-typescript/issues/425
 smithy-typescript-codegen/bin/
+smithy-typescript-codegen-test/bin/
 smithy-typescript-ssdk-codegen-test-utils/bin/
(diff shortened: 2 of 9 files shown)


kuhe:
diff example for JSv3

modified code for command deser (streaming)
```ts
/**
 * deserializeAws_json1_1XYZCommand
 */
export const de_XYZCommand = async (
  output: __HttpResponse,
  context: __SerdeContext & __EventStreamSerdeContext
): Promise<XYZCommandOutput> => {
  if (output.statusCode >= 300) {
    return de_XYZCommandError(output, context);
  }
  const contents = await bufferInitialResponse("someFieldName", de_XYZResponseStream, output, context);
  const response: XYZCommandOutput = {
    $metadata: deserializeMetadata(output),
    ...contents,
  };
  return response;
};
```

additional code in streaming member deser (`de_XYZResponseStream`)
```ts
    if (event["initial-response"] != null) {
      const ir = await parseBody(event["initial-response"].body, context);
      return {
        "initial-response": map({
          someNewField: [, ir.someNewField],
        }),
      } as any;
    }
```

PR #1113, opened 2023-12-14
feat: inline cjs index

kuhe:
This PR adds a script that can be run for any package. 
It is a variation of the standard `dist-cjs` or Node.js build.

The transform inlines the contents of the package into the package's `dist-cjs/index.js` file using `esbuild`.
This inlining does not include external packages, only the source code of the package itself. The purpose of this is to reduce file resolution time in application initialization for Node.js. 

Additionally, any files which have package metadata or filename metadata indicating they are variant files used by React Native are preserved as external to the inlined index.


syall (approved):

kuhe:
additional revisions:
- fix cases where react-native entry point is string
- fix for cyclical transitive variant files
- fix for multiple require path replacement of variant files

PR #1114, opened 2023-12-15
feat: paginator factory applied to codegen

kuhe:
this PR requires that https://github.com/smithy-lang/smithy-typescript/pull/1115 be merged and published


syall:
CI expected to fail until #1115 is merged, but approving in the meantime.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/PaginationGenerator.java
```diff
@@ -167,57 +160,41 @@ private void writePager() {
...
+            """
+            export const paginate$L: (config: $LPaginationConfiguration, input: $L, ...rest: any[]) => Paginator<$L> =
+            """,
+            operationName,
+            aggregatedClientName,
+            inputTypeName,
+            outputTypeName
+        );
+        writer
             .openBlock(
```
mtdowling, line 176: This is a template where it would help to make a Map to contain the variables by name, use a text block without openBlock, and then refer to variables by name in the template. See https://smithy.io/2.0/guides/building-codegen/generating-code.html#named-parameters

mtdowling (approved):

PR #1122, opened 2023-12-20
feat: add credential scope field

kuhe:
- Adds the `credentialScope` field to AWS credentials type
- Modifies the endpoints ruleset resolver to pull credentialScope from config.credentials if requested

### DIFF
--- a/packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
+++ b/packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
@@ -24,6 +24,13 @@ export const createConfigValueProvider = <Config extends Record<string, unknown>
     }
     return configValue;
   };
+  if (configKey === "credentialScope" || canonicalEndpointParamKey === "CredentialScope") {
+    return async () => {
+      const credentials = typeof config.credentials === "function" ? await config.credentials() : config.credentials;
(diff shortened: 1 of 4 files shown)


packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
```diff
@@ -24,6 +24,13 @@ export const createConfigValueProvider = <Config extends Record<string, unknown>
     }
     return configValue;
   };
+  if (configKey === "credentialScope" || canonicalEndpointParamKey === "CredentialScope") {
+    return async () => {
+      const credentials = typeof config.credentials === "function" ? await config.credentials() : config.credentials;
```
syall, line 29: For `credentials`, I think we may need to add an undefined check since `credentials` could possibly be undefined.

Though, I am not sure where the `createConfigValueProvider()` function is used in the request lifecycle.

syall (approved):

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
syall: That was a miss from my side, LGTM.

siddsriv (approved):

PR #1130, opened 2024-01-03
Add 2024 to CheckStyle RegexpHeader

trivikr:
*Issue #, if available:*
N/A

*Description of changes:*

Adds 2024 to CheckStyle RegexpHeader

---


syall:
Instead of continuously updating this regex per year, using the short-form header without a copyright year moving forward makes more sense.

trivikr:
We decided to stop using year headers in new source code from 2024 onwards.
We'll retain headers from previous code, to not change the git history of files.

PR #1133, opened 2024-01-04
Make CompressionInputConfig properties optional

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/pull/5643#issuecomment-1877397637

*Description of changes:*
Make CompressionInputConfig properties optional
* The properties need to be optional, as they don't need to be passed during client creation.
* The properties need to be required during resolution, since they're resolved by runtimeConfig.

---

### DIFF
--- a/packages/middleware-compression/src/resolveCompressionConfig.ts
+++ b/packages/middleware-compression/src/resolveCompressionConfig.ts
@@ -5,7 +5,9 @@ import { CompressionInputConfig, CompressionResolvedConfig } from "./configurati
 /**
  * @internal
  */
-export const resolveCompressionConfig = <T>(input: T & CompressionInputConfig): T & CompressionResolvedConfig => ({
+export const resolveCompressionConfig = <T>(
+  input: T & Required<CompressionInputConfig>
+): T & CompressionResolvedConfig => ({
   ...input,
   disableRequestCompression: normalizeProvider(input.disableRequestCompression),
   requestMinCompressionSizeBytes: async () => {
--- a/.changeset/nice-pumas-collect.md
+++ b/.changeset/nice-pumas-collect.md
@@ -1,0 +1,5 @@
+---
+"@smithy/middleware-compression": patch
+---
+
+Make CompressionInputConfig properties optional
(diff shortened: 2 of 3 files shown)


siddsriv (approved):

packages/middleware-compression/src/resolveCompressionConfig.ts
```diff
@@ -5,7 +5,9 @@ import { CompressionInputConfig, CompressionResolvedConfig } from "./configurati
 /**
  * @internal
  */
-export const resolveCompressionConfig = <T>(input: T & CompressionInputConfig): T & CompressionResolvedConfig => ({
+export const resolveCompressionConfig = <T>(
+  input: T & Required<CompressionInputConfig>
+): T & CompressionResolvedConfig => ({
   ...input,
   disableRequestCompression: normalizeProvider(input.disableRequestCompression),
   requestMinCompressionSizeBytes: async () => {
```
syall, line 13: nit: is it safer to guard against `undefined` values for `disableRequestCompression` and `requestMinCompressionSizeBytes`, and not make `CompressionInputConfig` required?

syall (approved):

packages/middleware-compression/src/resolveCompressionConfig.ts
trivikr: There won't be undefined values when calling `resolveCompressionConfig` as runtimeConfig resolution ensure that it's defined. If it doesn't, the client compilation should fail like it's happening in protocol tests.

Also, the [Required](https://www.typescriptlang.org/docs/handbook/utility-types.html) doesn't allow undefined.
[Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApQPYAcDOyDeAUMsnAPwBcyIArgLYBG0A3Mcg5cjmFKAOasAvsUIIMIbsgwMAVlXTY8AXgKkqAVmSDWyUeMnSZAJioAlCAEcawKBAAmAHgW4AfMhX41ydQBp2VGhA7CBhQey1mIA)

PR #1136, opened 2024-01-09
chore: only unit test changed packages

kuhe:
the md5 unit test keeps crashing (OOM) the test runner 🤕, let's just run unit tests in changed packages

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -5,10 +5,10 @@
     "build": {
       "dependsOn": ["^build"],
       "inputs": ["src/**/*", "package.json"],
-      "outputs": ["dist-types/**", "dist-cjs/**", "dist-es/**"]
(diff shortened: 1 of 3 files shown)


turbo.json
```diff
@@ -5,10 +5,10 @@
     "build": {
       "dependsOn": ["^build"],
       "inputs": ["src/**/*", "package.json"],
-      "outputs": ["dist-types/**", "dist-cjs/**", "dist-es/**"]
+      "outputs": ["dist-types/**/*", "dist-cjs/**/*", "dist-es/**/*"]
     },
     "test": {
-      "dependsOn": ["build"],
+      "dependsOn": [],
```
kuhe, line 11: Both karma and jest don't need the package in question to be built to run unit tests. 

jest (via babel) and karma (via typescript) compile the unit test sources during the test run.

trivikr (approved):

PR #1146, opened 2024-01-16
feat: use inline cjs build

kuhe:
This PR switches the `build:cjs` operation of all packages to use the inline transform. 

See https://github.com/smithy-lang/smithy-typescript/pull/1113

JS SDK tests:
- [x] integration
- [x] e2e
- [x] e2e legacy
- [x] test the metro bundler in react-native configuration


syall (approved):
LGTM given the tests in the PR description pass.

PR #1150, opened 2024-01-31
feat: generate unified error dispatcher

kuhe:
Currently, output deserializers generate one "Error Dispatcher" per operation output. This may mean hundreds of these dispatcher functions (given that many commands) when the set of modeled errors tends to only be 5 to 10 in number.

This PR merges all error dispatchers into one function per service. 

Comparison Metrics:
- Web: bundle size of S3 is ~400 bytes larger when fully gzipped/minified as in a production environment. Total of ~160kb -> ~161kb.

- Node.js: install size of dist-cjs 
  - S3, 1020kb -> 976kb
  - Lambda, 784kb -> 700kb
  - EC2, 6.5mb -> 5.8mb
  - SageMaker, 3.1mb -> 2.6mb

- What are the outlier AWS services with the highest count of distinct modeled exceptions? and what is the average count of modeled exceptions per service?
  - cloudtrail, 83
  - codecommit, 188 (600 line de_CommandError fn)
      - regardless of that, codecommit dist-cjs also gets smaller 1.1mb -> 896kb
  - codedeploy, 110
  - config-server, 53
  
  - a few services have between 20 and 50, (about 20 services)
  - the vast majority have less than 20 distinct exceptions (300+ services)


kuhe:
it is further possible to convert the syntax of the dispatcher from

```js
switch (code) {
  case 'A':
  case 'B':
      throw await de_ExceptionRes(output, context);
  case 'C':
  case 'D':
      throw await de_Exception2Res(output, context);
  default:
    throwDefaultError(...)
}
```

to

```js
const exceptions = {
  A: 1, // look up the index of the deser fn
  B: 1,
  C: 2,
  D: 2,
};
const deser = [
  de_ExceptionRes,
  de_Exception2Res
];
if (exceptions[code]) {
  throw await deser[exceptions[code]](output, context);
}
throwDefaultError(...)
```

But this is a very incremental reduction

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -298,51 +298,49 @@ public static void writeRetryableTrait(TypeScriptWriter writer, StructureShape e
...
 
         writer.writeDocs(errorMethodLongName);
-                writer.openBlock("const $L = async(\n"
-                               + "  output: $T,\n"
-                               + "  context: __SerdeContext,\n"
-                               + "): Promise<$T> => {", "}", errorMethodName, responseType, outputType, () -> {
+        writer.openBlock("const $L = async(\n"
+            + "  output: $T,\n"
+            + "  context: __SerdeContext,\n"
+            + "): Promise<unknown> => {", "}", errorMethodName, responseType, () -> {
```
syall, line 335: Clarifying question: is the change to `unknown` a breaking change, or does it not matter?

syall (approved):
LGTM.

Clarifying question: as an edge case, is there a possibility of a tree-shaked artifact to become larger with this change? E.g., only importing 1 command, which pre-PR would only import a smaller command-specific error deserializer, compared to post-PR which will now import a larger unified error deserializer?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -2108,7 +2121,7 @@ private void generateOperationResponseDeserializer(
             // status code that's not the modeled code (300 or higher). This allows for
             // returning other 2XX codes that don't match the defined value.
             writer.openBlock("if (output.statusCode !== $L && output.statusCode >= 300) {", "}", trait.getCode(),
-                    () -> writer.write("return $L(output, context);", errorMethodName));
+                    () -> writer.write("return $L(output, context) as any;", errorMethodName));
```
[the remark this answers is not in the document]
kuhe: The return type of the deser method is the operation output, so in order to return anything, it must be of that type.

However, this method throws 100% of the time. It should really be called without a return statement, but I didn't want to risk the method not throwing accidentally and then proceeding after this line.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
kuhe: it does not matter, it should be `Promise<never>` since this function is supposed to throw all the time

syall (approved):

PR #1155, opened 2024-02-05
fix: return throwDefaultError as never

kuhe:
Now that the container function returns `Promise<never>`, no reachable statement that isn't also `never` is allowed


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -357,7 +357,7 @@ static Set<StructureShape> generateUnifiedErrorDispatcher(
 
                 // Get the protocol specific error location for retrieving contents.
                 String errorLocation = bodyErrorLocationModifier.apply(context, "parsedBody");
-                writer.openBlock("return throwDefaultError({", "})", () -> {
+                writer.openBlock("return throwDefaultError({", "}) as never", () -> {
```
kuhe, line 360: this is easier than releasing a change to the return type of `throwDefaultError`, which may cause instability in existing clients

syall (approved):

PR #1158, opened 2024-02-06
feat: endpoint caching

kuhe:
https://github.com/aws/aws-sdk-js-v3/issues/5723

this PR adds a cache for the `resolveEndpoint` function keyed on two levels, the ruleset object as a first key and the endpoint parameters as the second key.  

This makes resolving endpoints repeatedly, such as when generating multiple signed URLs, faster. 

The expected effect on endpoint resolution with a cache hit is 1-5ms (depending on hardware) to <0.5ms.

### DIFF
--- a/packages/util-endpoints/src/EndpointResolverCache.ts
+++ b/packages/util-endpoints/src/EndpointResolverCache.ts
@@ -1,0 +1,90 @@
+import type { EndpointParams, EndpointV2, RuleSetObject } from "@smithy/types";
+
+/**
+ * @internal
+ *
+ * Used by the resolveEndpoint function.
+ */
+export class EndpointResolverCache {
+  private static rulesetMap = new Map<RuleSetObject, EndpointResolverCache>();
+  private store = new Map<string, EndpointV2>();
+
+  public constructor(
+    /**
+     * Desired maximum number of entries in the cache.
+     */
+    public capacity: number = 100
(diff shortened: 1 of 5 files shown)


kuhe:
- [ ] suggestion: the lifecycle of the cache should follow a client so as not to prevent garbage collection

kuhe:
- [ ] benchmark simple ruleset client
- [ ] benchmark 100% cache miss diff
- [ ] benchmark 50% cache miss diff

packages/util-endpoints/src/EndpointResolverCache.ts
```diff
@@ -0,0 +1,90 @@
...
+  public get(key: string): EndpointV2 | undefined {
+    return this.store.get(key);
+  }
+
+  /**
+   * Write to the cache.
+   */
+  public set(key: string, endpoint: EndpointV2): void {
+    this.store.set(key, endpoint);
+    this.trim();
```
trivikr, line 65: Use `LRUCache` from `mnemonist/lru-cache`?

We currently use it in `endpoint-cache` for flexible endpoints https://github.com/aws/aws-sdk-js-v3/blob/5afac5e175e7b0a92cd612c5a4730a4269c4fa52/packages/endpoint-cache/src/EndpointCache.ts#L1C23-L1C42

PR #1164, opened 2024-02-15
feat: add checked socket exhaustion warning when throughput is slow

kuhe:
Addresses https://github.com/aws/aws-sdk-js-v3/issues/3560

This PR adds a warning:
```
@smithy/node-http-handler:WARN socket usage at capacity=<X> and <Y> additional 
requests are enqueued. 
See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html
```

Where `X` is the configured maxSockets for the Node http(s) agent, and `Y` is the number of open requests. 

This warning is emitted in the following situation:

- a request is taking more than `socketAcquisitionWarningTimeout ?? connectTimeout + requestTimeout` to complete.
  - indicates delay in acquiring a socket
- no similar warning has been emitted in the last 15 seconds.
  - prevent log spam
- the number of active sockets is equal to `maxSockets`
  - indicates socket saturation
- the number of enqueued requests is greater than or equal to 2x `maxSockets`
  - indicates socket saturation

Notes:
- it's easy to reach the maxSocket limit even in normal usage. This is why we can't immediately emit the warning on reaching the limit. 
- this is why the PR uses a timeout function, and only when there is a backed up queue of requests both in timing and in request count is a warning emitted.

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -741,7 +741,8 @@ describe("NodeHttpHandler", () => {
       expect(console.warn).toHaveBeenCalledWith(
         "@smithy/node-http-handler:WARN",
         "socket usage at capacity=2 and 4 additional requests are enqueued.",
-        "See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html"
+        "See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html",
+        "or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config."
       );
     });
   });
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -720,10 +720,18 @@ describe("NodeHttpHandler", () => {
         {
           maxSockets: 2,
           sockets: {
-            addr: [null, null],
+            addr: [],
+            addr2: [null],
+            addr3: [null, null],
+            // this is not checked because an earlier addr causes the warning to be emitted.
+            addr4: Array.from({ length: 400 }),
           },
           requests: {
-            addr: [null, null, null, null],
+            addr: Array.from({ length: 0 }),
+            addr2: Array.from({ length: 3 }),
+            addr3: Array.from({ length: 4 }),
+            // this is not checked because an earlier addr causes the warning to be emitted.
+            addr4: Array.from({ length: 800 }),
           },
         } as any,
         lastTimestamp
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -704,4 +704,37 @@ describe("NodeHttpHandler", () => {
       expect(nodeHttpHandler.httpHandlerConfigs()).toEqual({});
     });
   });
+
+  describe("checkSocketUsage", () => {
+    beforeEach(() => {
+      jest.spyOn(console, "warn").mockImplementation(jest.fn());
+    });
+
+    afterEach(() => {
+      jest.resetAllMocks();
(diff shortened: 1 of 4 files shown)


trivikr (approved):

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -17,6 +17,7 @@ export { NodeHttpHandlerOptions };
 interface ResolvedNodeHttpHandlerConfig {
   requestTimeout?: number;
   connectionTimeout?: number;
+  socketAcquisitionWarningTimeout?: number;
```
syall, line 20: Can we mark this as `@internal` for now?

This is more for debugging rather than something customers should actually configure.

packages/types/src/http/httpHandlerInitialization.ts
```diff
@@ -34,6 +34,15 @@ export interface NodeHttpHandlerOptions {
...
   requestTimeout?: number;
 
+  /**
+   * Delay before the NodeHttpHandler checks for socket exhaustion,
+   * and emits a warning if the active sockets and enqueued request count is greater than
+   * 2x the maxSockets count.
+   *
+   * Defaults to connectionTimeout + requestTimeout or 3000ms if those are not set.
+   */
+  socketAcquisitionWarningTimeout?: number;
```
syall, line 44: Same comment as before, can we mark this with `@internal`?

packages/node-http-handler/src/node-http-handler.spec.ts
```diff
@@ -704,4 +704,37 @@ describe("NodeHttpHandler", () => {
...
+      const warningTimestamp = NodeHttpHandler.checkSocketUsage(
+        {
+          maxSockets: 2,
+          sockets: {
+            addr: [null, null],
+          },
+          requests: {
+            addr: [null, null, null, null],
+          },
+        } as any,
```
syall, line 728: Can we add multiple addresses in the test case?

syall (changes requested):

packages/node-http-handler/src/node-http-handler.ts
kuhe: This is intended to be configured by the customer. Because this is a console warning, there should be a way to shut it off. Setting a high timeout here would do it.

We don't use this number in an "internal" way. It goes straight from this config to the timeout function.

packages/node-http-handler/src/node-http-handler.spec.ts
kuhe: added

packages/types/src/http/httpHandlerInitialization.ts
kuhe: Alternatives are to not let this timeout be configurable, or have a different opt-in to the warning. 

Since users are dealing with a mysterious process exit, having the log output without opt-in I think would be better for their debugging experience.

syall (approved):

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -106,6 +160,14 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
 
       // determine which http(s) client to use
       const isSSL = request.protocol === "https:";
+      const agent = isSSL ? this.config.httpsAgent : this.config.httpAgent;
+
+      // If the request is taking a long time, check socket usage and potentially warn.
+      // This warning will be cancelled if the request resolves.
+      this.socketCheckTimeoutId = setTimeout(() => {
```
kuhe, line 167: - [ ] on second thought, this needs to be set on a map and cleared from that map. We can't use a single ID at a time because requests are concurrent.

kuhe: or rather than a map, the ID of the timeout needs to be localized to this promise so as not to be overwritten, not the handler instance.

syall (approved):

siebrand:
This causes https://github.com/generalui/s3p/issues/89

PR #1167, opened 2024-02-21
feat: enable ctor arg passthrough for requestHandler

kuhe:
This implements step 3 of the following progression mentioned earlier in https://github.com/smithy-lang/smithy-typescript/pull/1165.

#### Simplifying provision of custom settings to the requestHandler.

Previously:
```ts
import https from "node:https";
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
import { NodeHttpHandler } from "@smithy/node-http-handler";

const client = new DynamoDBClient({
  requestHandler: new NodeHttpHandler({
    requestTimeout: 3_000,
    httpsAgent: new https.Agent({
      maxSockets: 25
    }),
  }),
});
```

Step 2 (implemented):
```ts
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
import { NodeHttpHandler } from "@smithy/node-http-handler";

const client = new DynamoDBClient({
  requestHandler: new NodeHttpHandler({
    requestTimeout: 3_000,
    httpsAgent: { maxSockets: 25 },
  }),
});
```

Step 3 (this PR):
```ts
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";

const client = new DynamoDBClient({
  requestHandler: {
    requestTimeout: 3_000,
    httpsAgent: { maxSockets: 25 },
  },
});

// the longer forms will still be valid
```

AWS SDK sample diff in https://github.com/aws/aws-sdk-js-v3/pull/5820

PR #1168, opened 2024-02-21
feat: enable ctor arg passthrough for requestHandler (Java codegen)

kuhe:
Java codegen changes for https://github.com/smithy-lang/smithy-typescript/pull/1167

Sample SDK diff in https://github.com/aws/aws-sdk-js-v3/pull/5820


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeConfigGenerator.java
```diff
@@ -197,11 +197,20 @@ void generate(LanguageTarget target) {
...
+                    writer
+                        .indent(indentation)
+                        .disableNewlines()
+                        .openBlock(
+                            "$1L: " + defaultPrefix, ",\n", key,
+                            () -> {
+                                value.accept(writer);
+                            }
+                        )
+                        .dedent(indentation);
```
syall, line 213: This type of change probably means that the existing `TypeScriptIntegration` hook is not extensible enough, as special casing `requestHandler` doesn't seem right.

kuhe: The current type of `Map<String, Consumer<TypeScriptWriter>>` doesn't have any room to indicate whether the default `config?.someProperty ??` should be prefixed or not. 

What do you suggest?

kuhe: make a `Set<String>` static collection to indicate which fields should not have a default prefix of the value defined on config?

kuhe: going to add a method to TypeScriptIntegration

syall (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/RuntimeConfigGenerator.java
syall: Decided on keeping a static map of default prefixes rather than exposing a new hook in `TypeScriptIntegration`.

PR #1170, opened 2024-02-26
fix(core): retry after clock skew correction

kuhe:
PR adds metadata if a clock skew correction occurred after a request error. This metadata, when present, is considered a transient and retryable error. `maxAttempts` must be `>=2` to allow clock skew corrections to be retried.

Test code, using DynamoDB, which returns a date header that can be used to correct clock skew for sigv4.

```js
// run using Node.js as a *.js file.
const { DynamoDB } = require("@aws-sdk/client-dynamodb");
const pkgJson = require("@aws-sdk/client-dynamodb/package.json");

const dynamodb = new DynamoDB({
  region: "us-west-2",
  systemClockOffset: -30 * 60 * 1000, // start with a bad offset so your initial request is rejected.
});

(async () => {
  console.log(pkgJson.name, "@", pkgJson.version);
  console.log("system clock offset is", dynamodb.config.systemClockOffset);
  const listTables = await dynamodb.listTables({});
  // When functioning correctly, this operation retries and the system
  // clock offset is printed again with its corrected value.
  // No error is thrown by the SDK.
  console.log("system clock offset is", dynamodb.config.systemClockOffset);
  console.log("listTables", listTables.$metadata.httpStatusCode);
})().catch(async (e) => {
  console.log("error", e.name, e.$metadata.httpStatusCode);
  console.log("error", e.$response.headers);
  console.log("system clock offset is", dynamodb.config.systemClockOffset);
  // when malfunctioning (currently), the system clock offset is corrected in the error handler
  // but no retry is made.
  console.log("retrying manually");
  const listTables = await dynamodb.listTables({});
  console.log("system clock offset is", dynamodb.config.systemClockOffset);
  console.log("listTables after manual retry", listTables.$metadata.httpStatusCode);
});
```

Context:

As far as I can tell, a clock skew correction does not trigger a retry anymore, since AWS SDK v3.222.0, which corresponds to the retry SRA. 

https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.222.0

Versions tested:
- `"@aws-sdk/client-dynamodb": "<=3.521.0"` no retry after clock skew correction
- `"@aws-sdk/client-dynamodb": "<=3.450.0"` no retry after clock skew correction
- `"@aws-sdk/client-dynamodb": "<=3.350.0"` no retry after clock skew correction
- `"@aws-sdk/client-dynamodb": "<=3.222.0"` no retry after clock skew correction
- `"@aws-sdk/client-dynamodb": "<=3.221.0"` retry after clock skew correction
- `"@aws-sdk/client-dynamodb": "<=3.100.0"` retry after clock skew correction

### DIFF
--- a/packages/middleware-retry/src/retryMiddleware.ts
+++ b/packages/middleware-retry/src/retryMiddleware.ts
@@ -56,14 +56,14 @@ export const retryMiddleware = (options: RetryResolvedConfig & Partial<Previousl
         if (isRequest) {
           request.headers[REQUEST_HEADER] = `attempt=${attempts + 1}; max=${maxAttempts}`;
         }
-        initialSystemClockOffset = options.systemClockOffset | 0;
+        initialSystemClockOffset = options.systemClockOffset ?? 0 | 0;
         const { response, output } = await next(args);
         retryStrategy.recordSuccess(retryToken);
         output.$metadata.attempts = attempts + 1;
         output.$metadata.totalRetryDelay = totalRetryDelay;
         return { response, output };
       } catch (e: unknown) {
-        const latestSystemClockOffset = options.systemClockOffset | 0;
+        const latestSystemClockOffset = options.systemClockOffset ?? 0 | 0;
         const clockSkewCorrected = initialSystemClockOffset !== latestSystemClockOffset;
 
         const sdkError = e as SkdErrorWithClockSkewMetadata;
--- a/packages/middleware-retry/src/retryMiddleware.ts
+++ b/packages/middleware-retry/src/retryMiddleware.ts
@@ -16,18 +16,15 @@ import {
   RetryStrategyV2,
   RetryToken,
   SdkError,
-  SkdErrorWithClockSkewMetadata,
 } from "@smithy/types";
 import { INVOCATION_ID_HEADER, REQUEST_HEADER } from "@smithy/util-retry";
 import { v4 } from "uuid";
 
-import { PreviouslyResolved, RetryResolvedConfig } from "./configurations";
+import { RetryResolvedConfig } from "./configurations";
 import { isStreamingPayload } from "./isStreamingPayload/isStreamingPayload";
 import { asSdkError } from "./util";
 
-export const retryMiddleware = (options: RetryResolvedConfig & Partial<PreviouslyResolved>) => <
-  Output extends MetadataBearer = MetadataBearer
->(
+export const retryMiddleware = (options: RetryResolvedConfig) => <Output extends MetadataBearer = MetadataBearer>(
   next: FinalizeHandler<any, Output>,
   context: HandlerExecutionContext
 ): FinalizeHandler<any, Output> => async (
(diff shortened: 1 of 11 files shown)


syall:
It looks like the pre-SRA retry uses [the deprecated `StandardRetryStrategy` in `@smithy/middleware-retry`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/middleware-retry/src/StandardRetryStrategy.ts#L34), which checks if something is retryable using the [`defaultRetryDecider`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/middleware-retry/src/retryDecider.ts#L9). The `defaultRetryDecider` would [check and retry on clock skew errors](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/middleware-retry/src/retryDecider.ts#L14).

However, [the newer `StandardRetryStrategy` in `@smithy/util-retry`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/util-retry/src/StandardRetryStrategy.ts#L18) checks if [error are retryable only if they are transient or throttling](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/util-retry/src/StandardRetryStrategy.ts#L102-L104), which does NOT include clock skew errors.

To fix this regression, we can either:

1. (Recommended) Put clock skew error as `TRANSIENT` (`THROTTLING` doesn't seem right) by using [`isClockSkewError()`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/service-error-classification/src/index.ts#L13) in [`getErrorRetryType()`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/middleware-retry/src/retryMiddleware.ts#L109)
    - Although not perfect semantically, I think this is a good compromise for fixing the regression with minimal changes.
2. (Not recommended) Add [`isClockSkewError()`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/service-error-classification/src/index.ts#L13) as a condition in [`isTransientError()`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/service-error-classification/src/index.ts#L26)
3. (Not recommended) Expand [`RetryErrorInfo`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/types/src/retry.ts#L4) to include a clock skew error classification, and expand [`StandardRetryStrategy::isRetryableError()`](https://github.com/smithy-lang/smithy-typescript/blob/3b3774052b1985a680c3db7c244c59d03357b73c/packages/util-retry/src/StandardRetryStrategy.ts#L102-L104) to include the clock skew error classification.

packages/types/src/retry.ts
```diff
@@ -33,14 +35,19 @@ export type RetryErrorType =
  * @public
  */
 export interface RetryErrorInfo {
+  /**
+   * The error thrown during the initial request, if available.
+   */
+  error?: SdkError;
```
kuhe, line 41: Users have requested that the thrown exception is available in the retry strategy implementation. Otherwise, they cannot even check (for example) the http status code of an error when deciding whether to retry.

kuhe: Without this, they are forced to write a verbose retry wrapper instead of using a RetryStrategy.

kuhe:
Updated PR to use option 1. E2E test will be added in AWS SDK.

packages/middleware-retry/src/retryMiddleware.ts
```diff
@@ -55,7 +56,7 @@ export const retryMiddleware = (options: RetryResolvedConfig) => <Output extends
         output.$metadata.attempts = attempts + 1;
         output.$metadata.totalRetryDelay = totalRetryDelay;
         return { response, output };
-      } catch (e) {
+      } catch (e: any) {
```
syall, line 59: nit: do we need to cast this as `any`?

packages/middleware-retry/src/retryMiddleware.ts
```diff
@@ -45,6 +45,7 @@ export const retryMiddleware = (options: RetryResolvedConfig) => <Output extends
     if (isRequest) {
       request.headers[INVOCATION_ID_HEADER] = v4();
     }
+
```
syall, line 48: nit: extra newline

syall (approved):

packages/service-error-classification/src/index.ts
```diff
@@ -23,7 +32,8 @@ export const isThrottlingError = (error: SdkError) =>
  * cause where the NodeHttpHandler does not decorate the Error with
  * the name "TimeoutError" to be checked by the TRANSIENT_ERROR_CODES condition.
  */
-export const isTransientError = (error: SdkError) =>
+export const isTransientError = (error: SdkError | SkdErrorWithClockSkewMetadata) =>
+  isClockSkewCorrectedError(error) ||
```
syall, line 36: nit: I think it'd still be nice to move this check to the `getRetryErrorType()` in `retryMiddleware.ts`.

syall (approved):

packages/middleware-retry/src/retryMiddleware.ts
kuhe: without the type annotation this was `any` already, I made it explicit.

Alternative is to set `unknown` and cast below to SdkError when referencing `e`.

syall (approved):

PR #1171, opened 2024-02-27
fix(middleware-retry): retry on clock skew errors

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

Retry on clock skew errors.

---

### DIFF
--- a/packages/middleware-retry/src/retryMiddleware.ts
+++ b/packages/middleware-retry/src/retryMiddleware.ts
@@ -1,5 +1,5 @@
 import { HttpRequest, HttpResponse } from "@smithy/protocol-http";
-import { isServerError, isThrottlingError, isTransientError } from "@smithy/service-error-classification";
+import { isClockSkewError, isServerError, isThrottlingError, isTransientError } from "@smithy/service-error-classification";
 import { NoOpLogger } from "@smithy/smithy-client";
 import {
   AbsoluteLocation,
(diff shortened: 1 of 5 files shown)


packages/middleware-retry/src/retryMiddleware.ts
```diff
@@ -109,6 +110,7 @@ const getRetryErrorInfo = (error: SdkError): RetryErrorInfo => {
 const getRetryErrorType = (error: SdkError): RetryErrorType => {
   if (isThrottlingError(error)) return "THROTTLING";
   if (isTransientError(error)) return "TRANSIENT";
+  if (isClockSkewError(error)) return "TRANSIENT"; // V969229003
```
kuhe, line 113: I was told we should not reference internal identifiers.

packages/middleware-retry/src/retryMiddleware.ts
```diff
@@ -1,5 +1,5 @@
 import { HttpRequest, HttpResponse } from "@smithy/protocol-http";
-import { isServerError, isThrottlingError, isTransientError } from "@smithy/service-error-classification";
+import { isClockSkewError, isServerError, isThrottlingError, isTransientError } from "@smithy/service-error-classification";
```
kuhe, line 2: `isClockSkewError` is deprecated, and I think we should not use it because it will retry repeatedly for errors like `InvalidSignatureException`

PR #1176, opened 2024-02-29
feat: encoders accept strings

kuhe:
Clients currently have two internal fields of type `Encoder`, currently `Uint8Array => string`.

These have been updated to be `Uint8Array | string => string`, to allow direct conversion of utf8 strings to base64, rather than requiring a `Uint8Array` input.

The `Encoder` type has been loosened, since it is public.

related issues:
https://github.com/aws/aws-sdk-js-v3/issues/5745
https://github.com/aws/aws-sdk-js-v3/issues/5240

### DIFF
--- a/.changeset/fluffy-deers-doubt.md
+++ b/.changeset/fluffy-deers-doubt.md
@@ -1,0 +1,8 @@
+---
+"@smithy/middleware-serde": minor
+"@smithy/util-base64": minor
+"@smithy/util-utf8": minor
+"@smithy/types": minor
+---
+
+encoders allow string inputs
--- a/packages/middleware-serde/src/deserializerMiddleware.spec.ts
+++ b/packages/middleware-serde/src/deserializerMiddleware.spec.ts
@@ -1,3 +1,5 @@
+import { EndpointBearer, SerdeFunctions } from "@smithy/types";
+
 import { deserializerMiddleware } from "./deserializerMiddleware";
 
 describe("deserializerMiddleware", () => {
@@ -11,7 +13,7 @@ describe("deserializerMiddleware", () => {
         hostname: "hostname",
         path: "path",
       }),
-  };
+  } as EndpointBearer & SerdeFunctions;
 
   const mockArgs = {
     input: {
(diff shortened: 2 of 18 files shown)


kuhe:
This will allow users to send strings as input for fields that otherwise require `Uint8Array`, which can be cumbersome to initialize.

Though it is as easy as writing `Buffer.from('abcd')`, discovering this is somewhat obscure because it's not obvious that Buffer is a Uint8Array.

kuhe:
I verified this using SES

```ts
import * as sesClientModule from "@aws-sdk/client-ses";
import nodemailer from "nodemailer";

export const sendEmailWithAttachments = (from = "@email", to = "@email") => {
  const ses = new sesClientModule.SESClient({
    region: "us-west-2",
  });

  ses.middlewareStack.add(
    (next, context) => async (args: any) => {
      console.log("AWS SDK context", context.clientName, context.commandName);

      args.input.RawMessage.Data = args.input.RawMessage.Data.toString();
      console.log("AWS SDK request input", args.input);

      console.log("AWS SDK raw request", args.request);
      const result = await next(args);
    //   console.log("AWS SDK raw response", result.response);
      console.log("AWS SDK response metadata:", result.output.$metadata);
      // console.log("AWS SDK request output:", result.output);
      return result;
    },
    {
      name: "MyMiddleware",
      step: "build",
      override: true,
    }
  );

  const transporter = nodemailer.createTransport({
    SES: { ses, aws: sesClientModule },
  });

  return new Promise((resolve, reject) => {
    transporter.sendMail(
      {
        from,
        to,
        subject: "using string input for RawMessage",
        text: "using string input for RawMessage",
        attachments: [{ content: "Hello World!", filename: "hello.txt" }],
      },
      (err, info) => {
        if (err) {
          reject(err);
        } else {
          resolve(info);
        }
      }
    );
  });
};

sendEmailWithAttachments();

```

syall (changes requested):
For inputs that take `Uint8Array | string`, can we do a runtime check if it's not one of these types to restrict the types more?

I see `typeof xxx === "string"` is used, but can we use `instanceof Uint8Array` checks?

syall (approved):

Issue #1121, opened 2023-12-20
Invalid S3 arn error thrown for valid format


FelixMarcus:
### What?

Smithy middleware-endpoint is unable to properly recognise an S3 bucket arn following the format:

`arn:aws:s3:::bucket_name`

...despite it being the [the expected standard](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html).

Instead, [it will throw an error](https://github.com/smithy-lang/smithy-typescript/blob/f0ff3b7cdb93ce607c3873bab9a553566e0aec0b/packages/middleware-endpoint/src/service-customizations/s3.ts#L64) of `Invalid ARN: ${bucketName} was an invalid ARN.`

### Expected Behaviour

S3 ARN values without an account id should be valid and thus not throw an error.

### Where?

Tested on:

node@v18.19.0

Using:
- @aws-sdk/client-s3@3.472.0
- @aws-sdk/middleware-sdk-s3@3.470.0
- @smithy/middleware-endpoint@2.2.3

### Example

N.B. `s3BucketExampleArn` is passed in via cloudformation as an environment variable, and shows up correctly in the logs - matching an existing bucket in the accounts.

```
import { PutObjectCommand, S3Client } from '@aws-sdk/client-s3';

const input = {
      Body: JSON.stringify(payloads),
      Bucket: s3BucketExampleArn,
      Key: `example-key`,
    };
const command = new PutObjectCommand(input);
await this.s3Client.send(command);
```

```
/usr/src/app/node_modules/@smithy/middleware-endpoint/dist-cjs/service-customizations/s3.js:39
        throw new Error(`Invalid ARN: ${bucketName} was an invalid ARN.`);
              ^
Error: Invalid ARN: arn:aws:s3:::s3BucketExampleArn was an invalid ARN.
    at isArnBucketName (/usr/src/app/node_modules/@smithy/middleware-endpoint/dist-cjs/service-customizations/s3.js:39:15)
    at resolveParamsForS3 (/usr/src/app/node_modules/@smithy/middleware-endpoint/dist-cjs/service-customizations/s3.js:9:37)
    at resolveParams (/usr/src/app/node_modules/@smithy/middleware-endpoint/dist-cjs/adaptors/getEndpointFromInstructions.js:47:63)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #1227, opened 2024-04-03
fix(middleware-endpoint): check for s3 arn parts

bitIO:
As stated in this issue https://github.com/smithy-lang/smithy-typescript/issues/1121 arn check should not check s3 arns contain nor account nor region since they are not part of s3 arns

*Issue #, if available:*

#1121

*Description of changes:*

Check S3 arn is valid by not expecting account nor region to be in the arn (fail if any of those two are present)

---

### DIFF
--- a/packages/middleware-endpoint/src/service-customizations/s3.ts
+++ b/packages/middleware-endpoint/src/service-customizations/s3.ts
@@ -59,12 +59,15 @@ export const isDnsCompatibleBucketName = (bucketName: string): boolean =>
  * @internal
  */
 export const isArnBucketName = (bucketName: string): boolean => {
-  const [arn, partition, service, region, account, typeOrId] = bucketName.split(":");
+  // region and account are unused between service and bucket.
+  const [arn, partition, service, , , bucket] = bucketName.split(":");
+
   const isArn = arn === "arn" && bucketName.split(":").length >= 6;
-  const isValidArn =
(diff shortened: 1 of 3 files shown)


kuhe (approved):

packages/middleware-endpoint/src/service-customizations/s3.ts
```diff
@@ -61,9 +61,10 @@ export const isDnsCompatibleBucketName = (bucketName: string): boolean =>
 export const isArnBucketName = (bucketName: string): boolean => {
   const [arn, partition, service, region, account, typeOrId] = bucketName.split(":");
   const isArn = arn === "arn" && bucketName.split(":").length >= 6;
-  const isValidArn = [arn, partition, service, account, typeOrId].filter(Boolean).length === 5;
+  const isValidArn =
+    [arn, partition, service, typeOrId].filter(Boolean).length === 4 && [region, account].filter(Boolean).length == 0;
```
kuhe: We cannot require that region and account are definitively not set. 

An access point MRAP ARN works with and without the account id present, for example with `listObjects({ Bucket: "arn:aws:s3::12345:accesspoint/xyz.mrap" })`

kuhe:
I amended the PR to only make account id segment of the ARN optional, not to require that it is not present.

AndrewFossAWS (approved):

Issue #1159, opened 2024-02-07
SignatureV4 module loads unneeded dependencies


dreamorosi:
As per title, the `@smithy/signature-v4` package loads some dependencies (`@aws-crypto/*`) even though they are not needed / not used in the context of signing a request. This happens because of the package importing a utility function from the `@smithy/eventstream-codec`, and in doing so, it also loads the full content of the package and its dependencies.

This results in a bundle size that is 47% larger than it could be. Below an in depth exploration with reproduction steps.

## Use Case

In a Node.js environment (AWS Lambda) I am using the `@smithy/signature-v4` package to sign requests. The use case is a Lambda@Edge attached to a CloudFront distribution with a [Lambda function URL](https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html) as origin. Given that Lambda function URLs support only IAM as authentication method, I'm using the Lambda@Edge to manipulate the request & sign it (among other validations), so that by the time it reaches the origin it's authenticated. 

Below a high level diagram of what I just described above:

![image](https://github.com/smithy-lang/smithy-typescript/assets/7353869/7a430957-4b9c-4976-9059-615b0d44f0cb)

## Reproduction steps

I have prepared a bare bones repository with the function that uses the `@smithy/signature-v4` package that you can use to reproduce the steps below, you can find it [here](https://github.com/dreamorosi/smithy-sigv4) and follow along.

### 1. Clone the repository & install dependencies

Run `git clone git@github.com:dreamorosi/smithy-sigv4.git` and then `npm ci` to setup the repo.

### 2. Bundle code 

Run `npm run bundle` in the project's root.

This step uses `esbuild` (which is default in both AWS CDK and AWS SAM) to create a bundle of the function (`index.ts`) and all its dependencies. As you can see from the command I use to bundle, I am using ESM and enabling tree shaking:

```sh
esbuild --bundle index.ts --format=esm --platform=node --main-fields=module,main --tree-shaking=true --outfile=out/index.mjs --metafile=out/meta.json
```

The command generates two output files, one is the function code including dependencies (`out/index.mjs`) and the other is a meta file (`out/meta.json`) which we'll use in the next step. The main output file is not minified so you can optionally inspect the content to verify the claims of this issue.

### 3. Analyze the bundle

In your browser, open https://esbuild.github.io/analyze/ and drag (or select) the meta file (`out/meta.json`) generated at the previous step. The resulting chart should look like this:

[...]

PR #1233, opened 2024-04-04
feat: give SigV4 its own header formatter to avoid import of entire eventstream-codec package

kuhe:
*Issue #, if available:*
fixes https://github.com/smithy-lang/smithy-typescript/issues/1159

*Description of changes:*
temporarily duplicates the header formatting functionality from evenstream-codec to signature-v4 so that sigv4 does not have to import the entire eventstream-codec package and transitively the large-ish aws-crypto packages.

when Node.js 14 support is dropped, our typescript config will support submodules (package.json:exports) and then the headerMarshaller functionality can be reunified as a single submodule of e.g. `@smithy/core`.

### DIFF
--- a/.changeset/cuddly-ties-cover.md
+++ b/.changeset/cuddly-ties-cover.md
@@ -1,0 +1,5 @@
+---
+"@smithy/signature-v4": patch
+---
+
+internalize header format function from eventstream-codec into signature-v4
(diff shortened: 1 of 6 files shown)


AndrewFossAWS (approved):

dreamorosi:
Thank you for this PR, appreciate it!

PR #1239, opened 2024-04-11
fix: check dependencies when adding imports

kuhe:
when calling `addImport`, the TSWriter will either add the import or throw an error if it determines the import is a non-native package and has not been added via addDependency.

this is to prevent calls to `addImport` that forget to also call `addDependency`.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
@@ -16,13 +16,11 @@
 package software.amazon.smithy.typescript.codegen;
 
 import java.nio.file.Path;
-import java.util.List;
 import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.UnaryOperator;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
-import software.amazon.smithy.codegen.core.SymbolDependency;
 import software.amazon.smithy.codegen.core.SymbolReference;
 import software.amazon.smithy.codegen.core.SymbolWriter;
 import software.amazon.smithy.model.Model;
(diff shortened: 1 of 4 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -114,19 +142,54 @@ public TypeScriptWriter addIgnoredDefaultImport(String name, String from, String
      */
     @Deprecated
     public TypeScriptWriter addImport(String name, String as, String from) {
+        boolean isNodePackage = from.startsWith("node:");
```
kuhe, line 145: should a class wrap the String input `from` to provide answers like `boolean isNamespaced`, `boolean isNodeJSNativePackage`, `String getPackageName` etc.?

syall: That probably should go in `Dependency` in the long run.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -114,19 +142,54 @@ public TypeScriptWriter addIgnoredDefaultImport(String name, String from, String
...
      *
      * @param name Type to import.
      * @param as Alias to refer to the type as.
      * @param from PackageContainer to import the type from.
      * @return Returns the writer.
      */
     public TypeScriptWriter addImport(String name, String as, PackageContainer from) {
+        if (from instanceof Dependency) {
+            addDependency((Dependency) from);
+        }
```
syall, line 192: In the long run, we should move it to taking `Dependency` as the third parameter.

syall (approved):

trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/validation/ImportFrom.java
```diff
@@ -0,0 +1,93 @@
...
+
+package software.amazon.smithy.typescript.codegen.validation;
+
+import java.util.Set;
+import software.amazon.smithy.utils.SetUtils;
+
+/**
+ * Interprets the string portion of an import statement.
+ */
+public class ImportFrom {
```
syall, line 14: Add `@SmithyInternalApi` annotation

syall (approved):

syall (approved):

PR #1240, opened 2024-04-11
feat(types): add accountId to the credential identity interface

siddsriv:
*Issue #, if available:*
Internal JS-4633

*Description of changes:*
Adds `accountId` as optional to the `AwsCredentialIdentity` interface. This `accountId` may be used for endpoint routing. 

---

### DIFF
--- a/packages/types/src/identity/awsCredentialIdentity.ts
+++ b/packages/types/src/identity/awsCredentialIdentity.ts
@@ -24,6 +24,11 @@ export interface AwsCredentialIdentity extends Identity {
    * AWS credential scope for this set of credentials.
    */
   readonly credentialScope?: string;
+
+  /**
+   * AWS accountId that may be used for endpoint routing.
+   */
+  readonly accountId?: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


kuhe (approved):

syall (approved):

trivikr (approved):

packages/types/src/identity/awsCredentialIdentity.ts
```diff
@@ -24,6 +24,11 @@ export interface AwsCredentialIdentity extends Identity {
    * AWS credential scope for this set of credentials.
    */
   readonly credentialScope?: string;
+
+  /**
+   * AWS accountId that may be used for endpoint routing.
```
syall, line 29: After looking at this, maybe we should only say `AWS account ID` (similar to access key ID and secret access key), and not tie it to endpoint routing only.

```suggestion
   * AWS account ID
```

syall (changes requested):

packages/types/src/identity/awsCredentialIdentity.ts
siddsriv: yeah, i had left that in conservatively so that it's easy to know why we introduced this now. but i agree generally, let's do it as per your suggestion.

syall (approved):

PR #1252, opened 2024-04-30
feat: add support for accountId in configValueProvider

siddsriv:
*Issue #, if available:*
Internal JS-4633
Part 3

*Description of changes:*
adds support for accountId in `configValueProvider` (sourced from credentials)

*Testing*
```console
 PASS  src/adaptors/createConfigValueProvider.spec.ts (14.921 s)
  createConfigValueProvider
    ✓ should create a normalized provider for any config value (10 ms)
    ✓ should look up both the canonical Endpoint ruleset param name and any localized override (1 ms)
    ✓ uses a special lookup for CredentialScope (1 ms)
    ✓ uses a special lookup for accountId (1 ms)
    ✓ should normalize endpoint objects into URLs (1 ms)

Test Suites: 1 passed, 1 total
Tests:       5 passed, 5 total
Snapshots:   0 total
Time:        15.411 s
```

---

### DIFF
--- a/.changeset/brown-bags-leave.md
+++ b/.changeset/brown-bags-leave.md
@@ -1,0 +1,5 @@
+---
+"@smithy/middleware-endpoint": minor
+---
+
+add support for accountId in configValueProvider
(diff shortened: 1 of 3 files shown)


kuhe (approved):

milesziemer (approved):
In the future we should try to move this logic for determining aws endpoint param values like account id and credential scope into the SDK, but looking at the code it seems out of scope here.

PR #1256, opened 2024-05-01
feat: improve support for fetch and web-streams in Node.js

kuhe:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/4619

*Description of changes:*
- `streamCollector` and `sdkStreamMixin` components that are part of the client config, specifically the SerdeContext subset, have been modified to allow web streams

### DIFF
--- a/packages/fetch-http-handler/README.md
+++ b/packages/fetch-http-handler/README.md
@@ -2,3 +2,10 @@
 
 [![NPM version](https://img.shields.io/npm/v/@smithy/fetch-http-handler/latest.svg)](https://www.npmjs.com/package/@smithy/fetch-http-handler)
 [![NPM downloads](https://img.shields.io/npm/dm/@smithy/fetch-http-handler.svg)](https://www.npmjs.com/package/@smithy/fetch-http-handler)
+
+This is the default `requestHandler` used for browser applications.
+Although some versions of Node.js support the web streams API, this implementation has
+not been extensively tested in Node.js. 
+
+For the Node.js default `requestHandler` implementation, see instead
+[`@smithy/node-http-handler`](https://www.npmjs.com/package/@smithy/node-http-handler). 
--- a/packages/fetch-http-handler/README.md
+++ b/packages/fetch-http-handler/README.md
@@ -4,8 +4,8 @@
 [![NPM downloads](https://img.shields.io/npm/dm/@smithy/fetch-http-handler.svg)](https://www.npmjs.com/package/@smithy/fetch-http-handler)
 
 This is the default `requestHandler` used for browser applications.
-Although some versions of Node.js support the web streams API, this implementation has
-not been extensively tested in Node.js.
+Since Node.js introduced experimental Web Streams API in v16.5.0 and made it stable in v21.0.0,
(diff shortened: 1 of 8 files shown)


packages/fetch-http-handler/README.md
```diff
@@ -2,3 +2,10 @@
 
 [![NPM version](https://img.shields.io/npm/v/@smithy/fetch-http-handler/latest.svg)](https://www.npmjs.com/package/@smithy/fetch-http-handler)
 [![NPM downloads](https://img.shields.io/npm/dm/@smithy/fetch-http-handler.svg)](https://www.npmjs.com/package/@smithy/fetch-http-handler)
+
+This is the default `requestHandler` used for browser applications.
+Although some versions of Node.js support the web streams API, this implementation has
+not been extensively tested in Node.js.
```
trivikr, line 8: Explicitly call out version number, and mention it's not recommended instead of tested?
```suggestion
Since Node.js introduced experimental Web Streams API in v16.5.0 and made it stable in v21.0.0,
you can consider using `fetch-http-handler` in Node.js, although it's not recommended.
```

packages/node-http-handler/src/stream-collector/index.ts
```diff
@@ -3,8 +3,12 @@ import { Readable } from "stream";
 
 import { Collector } from "./collector";
 
-export const streamCollector: StreamCollector = (stream: Readable): Promise<Uint8Array> =>
-  new Promise((resolve, reject) => {
+export const streamCollector: StreamCollector = (stream: Readable | ReadableStream): Promise<Uint8Array> => {
+  if (isReadableStreamInstance(stream)) {
+    // web stream in Node.js indicates user has overridden requestHandler with FetchHttpHandler.
```
trivikr, line 8: This stream collector is specific to Node.js. Does this mean user has passed Node.js WebStream in body?

```suggestion
  if (isReadableStreamInstance(stream)) {
    // Web stream API in Node.js
```

If that's the case, the type on line 6 should be
```js
import { ReadableStream } from "stream/web";
```

trivikr (changes requested):

packages/node-http-handler/src/stream-collector/index.ts
kuhe:
```ts
import { ReadableStream } from "stream/web";
```
I don't think we can statically import this because of supporting Node.js 16.

```ts
import type ...
```
This might be possible? our TS types for Node.js 16 lib don't seem to support this right now, and I refreshed my installation today.

trivikr: > our TS types for Node.js 16 lib don't seem to support this right now

We need to bump `@types/node` for `@smithy/*` deps just like it's being done in https://github.com/aws/aws-sdk-js-v3/pull/6047. Let's do that first before merging this change?

trivikr (approved):

kuhe:
I tested this by running our dynamodb e2e suite using `fetch`. S3 get/putObject also seem to work.

syall (approved):

PR #1258, opened 2024-05-07
chore: update engines to node16

kuhe:
- drops support for node14
- major version bump for all non-experimental packages
- remove package-level `@tsconfig/recommended` dependency since it was unused
- added `lint:versions` to ensure consistent cross-package dependency versions

I confirmed that the generated code in `dist-cjs` does not change at all from this PR.


syall (approved):
Side note: we need a SSDK libs release

PR #1280, opened 2024-05-14
feat(cbor): add Smithy RPCv2 CBOR protocol generator

kuhe:
Adds a protocol generator for the trait `smithy.protocols#rpcv2Cbor`


packages/core/package.json
```diff
@@ -11,7 +11,7 @@
     "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo || exit 0",
     "lint": "npx eslint -c ../../.eslintrc.js \"src/**/*.ts\" && node ./scripts/lint",
     "format": "prettier --config ../../prettier.config.js --ignore-path ../.prettierignore --write \"**/*.{ts,md,json}\"",
-    "test": "yarn g:jest"
+    "test": "yarn g:jest --maxWorkers 1"
```
kuhe, line 14: jest has problems passing the JSON data between workers when it includes BigInt's, apparently?

packages/core/src/submodules/cbor/cbor-decode.ts
```diff
@@ -0,0 +1,477 @@
...
+      throw new Error(`unexpected major type ${major} in indefinite string.`);
+    }
+    if (minor === minorIndefinite) {
+      throw new Error("nested indefinite string.");
+    }
+
+    const bytes = decodeUnstructuredByteString(at, to);
+    const length = _offset;
+    at += length;
+    vector.push(...bytes);
```
milesziemer, line 343: Not sure how this works in JS - does spreading `bytes` create a new array?

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+        count: keys.length,
+      });
+      continue;
+    }
+
+    throw new Error(`data type ${input?.constructor?.name ?? typeof input} not compatible for encoding.`);
+  }
+}
+
+// function encodeString(input: string) {
```
milesziemer, line 222: FYI

packages/core/src/submodules/cbor/parseCborBody.ts
```diff
@@ -0,0 +1,78 @@
...
+    return sanitizeErrorCode(data["__type"]);
+  }
+
+  if (data.code !== undefined) {
+    return sanitizeErrorCode(data.code);
+  }
+};
+
+export const checkCborResponse = (response: HttpResponse): void => {
+  if (response.headers["smithy-protocol"] !== "rpc-v2-cbor") {
```
milesziemer, line 75: Is the casing of the header and/or value guaranteed?

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
```diff
@@ -0,0 +1,1227 @@
...
+  }
+
+  const data: any = await parseBody(output.body, context);
+  let contents: any = {};
+  contents = de_RpcV2CborSparseMapsInputOutput(data, context);
+  const response: RpcV2CborSparseMapsCommandOutput = {
+    $metadata: deserializeMetadata(output),
+    ...contents,
+  };
+  return response;
```
milesziemer, line 488: All these de_Command functions look basically the same, other than the function used to assign to `contents`. Is there something else that could happen here or could refactor out the common code?

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
```diff
@@ -0,0 +1,1227 @@
...
+const deserializeMetadata = (output: __HttpResponse): __ResponseMetadata => ({
+  httpStatusCode: output.statusCode,
+  requestId:
+    output.headers["x-amzn-requestid"] ?? output.headers["x-amzn-request-id"] ?? output.headers["x-amz-request-id"],
+  extendedRequestId: output.headers["x-amz-id-2"],
+  cfId: output.headers["x-amz-cf-id"],
+});
+
+const throwDefaultError = withBaseException(__BaseException);
+const buildHttpRpcRequest = async (
```
milesziemer, line 1197: Could this be part of a library or is it custom?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -153,6 +153,17 @@ public TypeScriptWriter addImport(String name, String as, PackageContainer from)
         return this.addImport(name, as, from.getPackageName());
     }
 
+    /**
+     * Same as {@link #addImport(String, String, PackageContainer)} but appends a
+     * submodule path, for example "@smithy/core/cbor".
+     */
+    public TypeScriptWriter addImportSubmodule(String name, String as, PackageContainer from, String submodule) {
+        if (from instanceof Dependency) {
```
milesziemer, line 161: Suggestion (also just excited about any java features past 8): You can use instanceof pattern matching here: https://blogs.oracle.com/javamagazine/post/pattern-matching-for-instanceof-in-java-14

trivikr (approved):
LGTM. I didn't go through CBOR encode/decode source code, as it has unit tests.

packages/core/src/submodules/cbor/cbor-decode.ts
siddsriv: if im not wrong, it spreads the elements of the `bytes` array as individual arguments to `push()`.

Makefile
```diff
@@ -0,0 +1,19 @@
+.PHONY: build sync
+
+build:
+	./gradlew clean build publishToMavenLocal
+
+sync:
+	gh repo sync $$GITHUB_USERNAME/smithy-typescript -b main
```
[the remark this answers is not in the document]
kuhe: `sync` is for local dev, those who install gh cli can opt to use it. 

`npx` works in CI, and is used in the new status check which can be seen in this PR.

packages/core/src/submodules/cbor/cbor-decode.ts
```diff
@@ -0,0 +1,477 @@
...
+  }
+  if (textDecoder) {
+    return textDecoder!.decode(bytes.subarray(at, to));
+  }
+  return toUtf8(bytes.subarray(at, to));
+}
+
+function demote(bigInteger: bigint): number {
+  const num = Number(bigInteger);
+  if (num < Number.MIN_SAFE_INTEGER || Number.MAX_SAFE_INTEGER < num) {
```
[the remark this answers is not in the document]
kuhe: This function is used to demote (cast) a `BigUint64` to `number`. 
This is not _generally_ safe, but in this context it is safe because it is only used to retrieve a string or array length.

Since `Number.MAX_SAFE_INTEGER`  is much larger than the spec-allowed maximums for JS strings and arrays, this will not cause information loss.

packages/core/src/submodules/cbor/cbor-decode.ts
kuhe: semantically, it doesn't require the creation of an intermediate temporary array, but I believe the implementation does so.

I'm switching this to an index for loop, since `...args` spread is known to crash earlier than index for-loops.

packages/core/src/submodules/cbor/parseCborBody.ts
kuhe: spec doesn't say whether to check casing, but since http headers are supposed to be case-insensitive I'll lowercase the value to compare

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
kuhe: This can be moved to the `@smithy/core/cbor` submodule.

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
kuhe: I believe shortening this may be out of scope, because they may end up identical, but have the potential to be different based on certain model parameters. It's a general issue affecting codegen of all protocols.

packages/core/src/submodules/cbor/cbor-encode.ts
kuhe: I wanted to leave this as a reference

packages/core/src/submodules/cbor/cbor-decode.ts
```diff
@@ -0,0 +1,483 @@
...
+  } else {
+    exponentComponent = Math.pow(2, exponent - 15);
+    summation = 1;
+  }
+
+  summation += fraction / 1024;
+  return scalar * (exponentComponent * summation);
+}
+
+// function bytesToFloat32(a: Uint8, b: Uint8, c: Uint8, d: Uint8): Float32 {
```
milesziemer, line 221: Same as https://github.com/smithy-lang/smithy-typescript/pull/1280#discussion_r1651555302

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
milesziemer: Looking again, `buildHttpRpcRequest` uses [util-body-length-browser](https://github.com/smithy-lang/smithy-typescript/pull/1280/files#diff-ef0fad4ac8f2e42e19d601ca95685947a35dfa2980fb6c4d2ada6b95bde4d03fR4) directly - should this be provided through __SerdeContext depending on the runtime?

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
milesziemer: Good point. Something to investigate in the future I guess.

packages/core/src/submodules/cbor/cbor-encode.ts
kuhe: This is a JS implementation of `@smithy/util-utf8` `fromUtf8 -> Uint8Array` or `Buffer.from("string")`

private/smithy-rpcv2-cbor/src/protocols/Rpcv2cbor.ts
kuhe: since the body for CBOR is always Uint8Array, undefined, or streaming, the browser implementation of calculateBodyLength is sufficient for Node.js as well.

mtdowling (changes requested):

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+      encodeHeader(input.major, input.count);
+      continue;
+    }
+
+    ensureSpace(typeof input === "string" ? input.length * 4 : 64);
+
+    if (typeof input === "string") {
+      if (USE_BUFFER) {
+        encodeHeader(majorUtf8String, Buffer.byteLength(input));
+        if ((data as BufferWithUtf8Write).utf8Write) {
```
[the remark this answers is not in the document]
kuhe: No, I'm removing this if-statement. I was using `utf8Write` previously, when referencing cbor-x, but the method is undocumented and currently appears to be no different than the `.write` method, which is being used.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+    const input = encodeStack.pop();
+    if (input?.headerEncode === headerEncode) {
+      encodeHeader(input.major, input.count);
+      continue;
+    }
+
+    ensureSpace(typeof input === "string" ? input.length * 4 : 64);
+
+    if (typeof input === "string") {
+      if (USE_BUFFER) {
```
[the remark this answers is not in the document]
kuhe: Using a function here showed decreased performance, with relative throughput decreasing from ~25% to ~20% in `list<struct> PutMetricData-like`.

I take it to mean the additional stack frame defeats the avoidance of the if-statement.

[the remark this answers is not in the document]
kuhe: There was previously an encoding function for each type of data, but it evolved to a single mostly-inlined `encode()` function in repeated iterative benchmarking, and also changed from recursive to the `encodeStack` work queue currently seen in the function.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+        count: input.length,
+      });
+      continue;
+    } else if (typeof input.byteLength === "number") {
+      ensureSpace(input.length * 2);
+      encodeHeader(majorUnstructuredByteString, input.length);
+      data.set(input, cursor);
+      cursor += input.byteLength;
+      continue;
+    } else if (typeof input === "object" && "tag" in input && "value" in input && Object.keys(input).length === 2) {
```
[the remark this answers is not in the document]
kuhe: this is for tag detection, but after some clarification it's most likely that we won't need this block, so I'll be removing it.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (remaining < bytes) {
+    if (cursor < 16_000_000) {
+      resize(Math.max(data.byteLength * 4, data.byteLength + bytes));
+    } else {
+      resize(data.byteLength + bytes + 16_000_000);
+    }
+  }
+}
+
+export function toUint8Array(): Uint8Array {
```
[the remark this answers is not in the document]
kuhe: The currently exported public API is in the `cbor.ts` file, which is an object with 3 static methods:

```ts
export const cbor = {
  deserialize(Uint8Array) -> any
  serialize(any) -> Uint8Array
  resizeEncodingBuffer(number) -> void
}
```


Exports from this file are not module level exports and are inaccessible to users.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+      });
+      continue;
+    } else if (typeof input === "object") {
+      const keys = Object.keys(input);
+      for (let i = keys.length - 1; i >= 0; --i) {
+        const key = keys[i];
+        encodeStack.push(input[key]);
+        encodeStack.push(key);
+      }
+      encodeStack.push({
```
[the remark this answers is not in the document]
kuhe: Good spot 😮. The push of a header encoding task always directly precedes the `continue` statement, so it gets popped immediately. 

Yes, these can be called on the spot instead of pushed.

packages/core/src/submodules/cbor/cbor-encode.ts
kuhe: I'll save it somewhere else in case I need it then

packages/core/src/submodules/cbor/cbor-encode.ts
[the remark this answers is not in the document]
kuhe: this changes improves relative serialization throughput from the 25-28% range to the 28-30% range for `list<struct> PutMetricData-like`.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+export function toUint8Array(): Uint8Array {
+  const out = alloc(cursor);
+  out.set(data.subarray(0, cursor), 0);
+  cursor = 0;
+  return out;
+}
+
+export function resize(size: number) {
+  const old = data;
+  data = alloc(size);
```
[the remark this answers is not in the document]
kuhe: The global encoding buffer is inaccessible to consumers, and so it cannot be passed to workers and won't be subject to concurrent operations. Workers will each have a copy of that buffer. 

Because the encoding function is synchronous and the runtime single-threaded, the buffer is in use by only one encoding process at a time. A copy is made when the result is handed over to the outgoing request. This copy is done very quickly and represents <1% of the time spent.

I believe using a pool of buffers would not improve the speed or memory usage, even if it could avoid that copy operation, due to the speed of that copy operation and the overhead of managing the buffer pool.

packages/core/src/submodules/cbor/cbor-encode.ts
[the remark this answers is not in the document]
mtdowling: 👍 Ah, I think this changed while I was writing up my review.

packages/core/src/submodules/cbor/cbor-encode.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (cursor < 16_000_000) {
+      resize(Math.max(data.byteLength * 4, data.byteLength + bytes));
+    } else {
+      resize(data.byteLength + bytes + 16_000_000);
+    }
+  }
+}
+
+export function toUint8Array(): Uint8Array {
+  const out = alloc(cursor);
```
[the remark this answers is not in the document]
kuhe: I don't know of a specific Node.js API that maps to the term "buffer pooling". If you mean using for example creating a manager class that allocates and reclaims separate buffers, I don't believe it will improve the perf profile over using a single buffer, for reasons given below in the comment thread about the global buffer.

Initial development was done with individual buffers per encoding process, and iterative development showed that allocation/reclamation was costly and would unpredictably trigger GC. Even a well managed pool shouldn't beat the simplicity of a single buffer.

packages/core/src/submodules/cbor/parseCborBody.ts
```diff
@@ -0,0 +1,114 @@
...
+    headers,
+  };
+  if (resolvedHostname !== undefined) {
+    contents.hostname = resolvedHostname;
+  }
+  if (body !== undefined) {
+    contents.body = body;
+    try {
+      contents.headers["content-length"] = String(calculateBodyLength(body));
+    } catch (e) {}
```
milesziemer, line 111: Should this error be propagated?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SmithyCoreSubmodules.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.typescript.codegen;
+
+public final class SmithyCoreSubmodules {
```
milesziemer, line 8: Suggestion: doc comment on what this class contains, the '/' was slightly confusing to me at first, until I saw the usage in `addImportSubmodule`

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/cbor/SmithyRpcV2Cbor.java
```diff
@@ -0,0 +1,334 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.typescript.codegen.protocols.cbor;
+
+import java.util.Set;
+import java.util.TreeSet;
+import software.amazon.smithy.aws.traits.protocols.AwsQueryCompatibleTrait;
```
milesziemer, line 10: Can't remember if there was some discussion on why we need this here. Is there any way around it?

packages/core/src/submodules/cbor/parseCborBody.ts
kuhe: no, because if it fails that means there is no content length as far as we know, and is a valid case.

This is equivalent to the following code, only we don't have the `hasBodyLength` function:

```ts

if (hasBodyLength(body)) {
  headers['content-length'] = calculateBodyLength(body);
}

```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/cbor/SmithyRpcV2Cbor.java
kuhe: There's some conditional generated code if the trait is present, and the spec says to handle this trait, so I don't see how we avoid this line?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/SmithyCoreSubmodules.java
kuhe: added docblock

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/cbor/SmithyRpcV2Cbor.java
kuhe: note: remove AwsQueryCompat from here, implement in extending class in JSv3

milesziemer (approved):

PR #1281, opened 2024-05-14
chore(core): enable package exports in smithy/core

kuhe:
similar to https://github.com/aws/aws-sdk-js-v3/pull/6079

Creates a submodules system in `@smithy/core`. 

This allows the creation of any number of `@smithy/core/submodule`s behaving like a separate package but existing within the core package. This is aimed at controlling/reducing the package count (currently 56) and consolidating version consistency while retaining the benefits of modularity and preserving existing Node.js init performance optimizations.


trivikr (approved):

syall (approved):
Approving, pending `cbor` removal and `util-identity-and-auth` being moved back to the root package

PR #1284, opened 2024-05-21
Export types with `export type`

libre-man:
*Description of changes:* Some types were exported with `export` instead of `export type` which is causing issues in our build system. This indicates more clearly that a type is exported, and no runtime behaviour is changed.


kuhe:
This is a problem with the build system, since type export/import statements are optional.

What is the error?

libre-man:
The problem is that we strongly prefer to consume the source code (better go-to-definition, faster builds for us), not the compiled code. Webpack then complains about it not being to find the things being exported here, as they refer to types.

What would be the downside of having these exported as types instead of normal exports?

kuhe:
You generate the client source code and merge it into a larger compiled application? That's fine.

I'll let the team know to review this PR and we will likely merge it, but the real fix is going to be in your webpack configuration. 

You can also configure a post-generation step. You may have a linting/formatting transform applied after code generation, and in that you can add https://typescript-eslint.io/rules/consistent-type-exports/ or https://biomejs.dev/linter/rules/use-export-type/.

kuhe (approved):

kuhe:
```
IndexGenerator.java:118: Line is longer than 120 characters (found 122). [LineLength]
```

libre-man:
> @libre-man It looks like the Java checkstyle is failing, can you resolve the errors?

Done @syall!

syall (approved):

Issue #1278, opened 2024-05-13
TypeScript timestamp input serialization should not remove milliseconds


rkyker:
As per the Smithy documentation [6.4. timestampFormat trait](https://smithy.io/1.0/spec/core/protocol-traits.html#timestampformat-trait):
* date-time: "...with optional millisecond precision..."
* epoch-seconds: "...with optional millisecond precision (for example, 1515531081.123). Values that are more granular than millisecond precision SHOULD be truncated to fit millisecond precision. "

Currently in: [HttpProtocolGeneratorUtils - getTimestampInputParam()](https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java#L74)

* The milliseconds are [removed from the output DATA_TIME format](https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java#L83)
* The milliseconds are [rounded up to the second for the EPOCH_SECONDS format](https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpProtocolGeneratorUtils.java#L85)

In both cases the milliseconds (if present) should be passed through so that the input timestamp value is not modified by the serialization.

In particular, we are using the date-time format and need the milliseconds passed through.

PR #1289, opened 2024-05-23
fix: serialize milliseconds for DateTime and EpochSeconds timestamp formats

kuhe:
fixes https://github.com/smithy-lang/smithy-typescript/issues/1278

### DIFF
--- a/packages/smithy-client/src/ser-utils.ts
+++ b/packages/smithy-client/src/ser-utils.ts
@@ -22,7 +22,7 @@ export const serializeFloat = (value: number): string | number => {
 };
 
 /**
- * @param date - to be serialized
+ * @param date - to be serialized.
  * @returns https://smithy.io/1.0/spec/core/protocol-traits.html#timestampformat-trait date-time format.
  */
 export const serializeDateTime = (date: Date): string => {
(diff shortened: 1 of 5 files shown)


syall (approved):

packages/smithy-client/src/ser-utils.ts
```diff
@@ -20,3 +20,11 @@ export const serializeFloat = (value: number): string | number => {
       return value;
   }
 };
+
+/**
+ * @param date - to be serialized.
+ * @returns https://smithy.io/1.0/spec/core/protocol-traits.html#timestampformat-trait date-time format.
+ */
+export const serializeDateTime = (date: Date): string => {
+  return date.toISOString();
```
kuhe, line 29: this is a bit redundant at a glance, but leaves the option to modify truncation or formatting later without a huge codegen diff

packages/smithy-client/src/ser-utils.ts
```diff
@@ -20,3 +20,9 @@ export const serializeFloat = (value: number): string | number => {
       return value;
   }
 };
+
+/**
+ * @param date - to be serialized.
+ * @returns https://smithy.io/1.0/spec/core/protocol-traits.html#timestampformat-trait date-time format.
```
syall, line 26:
```suggestion
 * @returns https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait date-time format.
```

kuhe, line 26:
```suggestion
 * @returns https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait date-time format.
```

syall (approved):

Issue #974, opened 2023-09-28
fetch-http-handler: abortSignal, requestTimeout issues


u873838:
If an abort signal is passed, it is first included in the request options:

https://github.com/awslabs/smithy-typescript/blob/2f70f105a1093c0c5d1caaa84ce292e5917f31ec/packages/fetch-http-handler/src/fetch-http-handler.ts#L99-L101

Then later, the abort signal will be added to the race promises:

https://github.com/awslabs/smithy-typescript/blob/2f70f105a1093c0c5d1caaa84ce292e5917f31ec/packages/fetch-http-handler/src/fetch-http-handler.ts#L144-L154

It seems like the latter block of code can be removed. Relying solely on `fetch`'s internal abort signal signally mechanism seems preferable.

kuhe:
can track release in https://github.com/aws/aws-sdk-js-v3/issues/4872

PR #1308, opened 2024-06-14
chore: switch to platform AbortController & Signal implementations after dropping Node.js 14

kuhe:
fixes https://github.com/aws/aws-sdk-js-v3/issues/4872
fixes https://github.com/smithy-lang/smithy-typescript/issues/974

this deprecates our own Abort types and implementation and switches to the platform types where available.

AbortController / AbortSignal are available in Node.js 16+.

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -165,8 +165,10 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
             reject(abortError);
           };
           if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+            // preferred.
             (abortSignal as AbortSignal).addEventListener("abort", onAbort);
           } else {
+            // backwards compatibility
             abortSignal.onabort = onAbort;
           }
         })
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -159,11 +159,16 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
     if (abortSignal) {
       raceOfPromises.push(
         new Promise<never>((resolve, reject) => {
-          abortSignal.onabort = () => {
+          const onAbort = () => {
             const abortError = new Error("Request aborted");
             abortError.name = "AbortError";
             reject(abortError);
           };
+          if (typeof abortSignal.addEventListener === "function") {
+            abortSignal.addEventListener("abort", onAbort);
+          } else {
+            abortSignal.onabort = onAbort;
+          }
         })
       );
     }
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -164,8 +164,8 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
             abortError.name = "AbortError";
             reject(abortError);
           };
-          if (typeof abortSignal.addEventListener === "function") {
-            abortSignal.addEventListener("abort", onAbort);
+          if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+            (abortSignal as AbortSignal).addEventListener("abort", onAbort);
           } else {
             abortSignal.onabort = onAbort;
           }
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -611,7 +611,7 @@ describe("NodeHttpHandler", () => {
       ).rejects.toHaveProperty("name", "TypeError");
     });
 
-    fit("will destroy the request when aborted", async () => {
+    it("will destroy the request when aborted", async () => {
       const mockResponse = {
         statusCode: 200,
         headers: {},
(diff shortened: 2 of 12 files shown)


packages/types/src/abort.ts
```diff
@@ -1,12 +1,13 @@
...
  * @public
  */
-export interface AbortHandler {
-  (this: AbortSignal, ev: any): any;
-}
+export { AbortHandler };
 
 /**
  * @public
+ * @deprecated use platform (global) type for AbortController | AbortSignal.
```
trivikr, line 10:
```suggestion
 * @deprecated use platform (global) type for AbortSignal.
```

packages/types/src/abort.ts
```diff
@@ -28,6 +29,7 @@ export interface AbortSignal {
...
```
trivikr, line 32:
```suggestion
 * @deprecated use platform (global) type for AbortController.
```

trivikr (approved):
nit: removing redundant names in docs

trivikr:
Verified that `AbortController` was made stable in Node.js v15.4.0

https://nodejs.org/api/globals.html#class-abortcontroller

haydenbaker (approved):

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -159,11 +159,16 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
...
+          const onAbort = () => {
             const abortError = new Error("Request aborted");
             abortError.name = "AbortError";
             reject(abortError);
           };
+          if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+            (abortSignal as AbortSignal).addEventListener("abort", onAbort);
+          } else {
+            abortSignal.onabort = onAbort;
+          }
```
syall, line 171: Can we add a comment here denoting which path is the deprecated legacy path?

packages/node-http-handler/src/node-http-handler.spec.ts
```diff
@@ -611,20 +611,20 @@ describe("NodeHttpHandler", () => {
       ).rejects.toHaveProperty("name", "TypeError");
     });
 
-    it("will destroy the request when aborted", async () => {
+    fit("will destroy the request when aborted", async () => {
```
syall, line 614: `fit` should be reverted and CI tested again

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -232,13 +232,18 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
...
+          req.destroy();
           const abortError = new Error("Request aborted");
           abortError.name = "AbortError";
           reject(abortError);
         };
+        if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+          (abortSignal as AbortSignal).addEventListener("abort", onAbort);
+        } else {
+          abortSignal.onabort = onAbort;
+        }
```
syall, line 246: Can we add a comment here denoting which path is the deprecated legacy path?

packages/node-http-handler/src/node-http2-handler.ts
```diff
@@ -181,12 +181,17 @@ export class NodeHttp2Handler implements HttpHandler<NodeHttp2HandlerOptions> {
...
           req.close();
           const abortError = new Error("Request aborted");
           abortError.name = "AbortError";
           rejectWithDestroy(abortError);
         };
+        if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+          (abortSignal as AbortSignal).addEventListener("abort", onAbort);
+        } else {
+          abortSignal.onabort = onAbort;
+        }
```
syall, line 194: Can we add a comment here denoting which path is the deprecated legacy path?

packages/types/src/abort-handler.ts
```diff
@@ -0,0 +1,8 @@
+import type { AbortSignal as DeprecatedAbortSignal } from "./abort";
+
+/**
+ * @public
+ */
+export interface AbortHandler {
+  (this: AbortSignal | DeprecatedAbortSignal, ev: any): any;
+}
```
syall, line 8: nit question: will this ever need to be scoped down or have extra arguments?

packages/util-waiter/src/createWaiter.ts
```diff
@@ -1,12 +1,17 @@
...
-const abortTimeout = async (abortSignal: AbortSignal): Promise<WaiterResult> => {
+const abortTimeout = async (abortSignal: AbortSignal | DeprecatedAbortSignal): Promise<WaiterResult> => {
   return new Promise((resolve) => {
-    abortSignal.onabort = () => resolve({ state: WaiterState.ABORTED });
+    const onAbort = () => resolve({ state: WaiterState.ABORTED });
+    if (typeof (abortSignal as AbortSignal).addEventListener === "function") {
+      (abortSignal as AbortSignal).addEventListener("abort", onAbort);
+    } else {
+      abortSignal.onabort = onAbort;
+    }
```
syall, line 14: Can we add a comment here denoting which path is the deprecated legacy path?

syall (changes requested):
The jest `fit` needs to be reverted.

Is there a reason for the `addEventListener()` and `.onabort` branching? Is it to differentiate the platform and custom controller implementation we had? 

It'd also be nice to import every reference of the deprecated `AbortController` as `DeprecatedAbortController as you did in some files already.

packages/node-http-handler/src/node-http-handler.spec.ts
kuhe, line 614:
```suggestion
    it("will destroy the request when aborted", async () => {
```

packages/types/src/abort-handler.ts
kuhe: this should be `ev: Event -> void`, but I wouldn't change it at this point for compatibility

syall (approved):

kuhe:
confirmed TS22 is action runner memory issue

Issue #1305, opened 2024-06-13
[BUG] Stale timeouts in node-http-handler


deftomat:
According to the [node-http-handler source code](https://github.com/smithy-lang/smithy-typescript/blob/74a4554ac4bcf3dab4a87a23e835c3fa82180c4a/packages/node-http-handler/src/node-http-handler.ts#L174-L182), there may be 3000ms timeouts which are never canceled when the request fails.

**Context:**

We are having Lambdas which are calling DynamoDB with on-demand scaling in a large burst as users starts coming to the system in the morning and there is a possibility that Dynamo responds with the error and AWS-SDK v3 will automatically retry the request which will then success.

Due to this, a timeout created in `node-http-handler` for the first request (failed one) will never be canceled and our lambdas took 3 seconds to respond even when our own code + DynamoDB call took less than 25ms.

Lambdas, by default, do not respond until the event loop is empty. So, this 3 seconds timeout is increasing our response times from a few milliseconds to 3 seconds unnecessarily.

kuhe:
> Lambdas, by default, do not respond until the event loop is empty

How do you demonstrate this?

```js
setTimeout(() => {
  console.log('after 3000');
}, 3000);
export const handler = async (event) => {
  setTimeout(() => {
    console.log('after 3000');
  }, 3000);
  const response = {
    statusCode: 200,
    body: JSON.stringify('Hello from Lambda!'),
  };
  return response;
};
```

This Lambda fn responds immediately because the handler function returns a value. The two timeouts in and outside the handler are not blocking.

deftomat:
Well, It will wait for an empty event loop when you use callback (3rd handler parameter) instead of returning promise.

We have been forced to used callbacks instead of returning promises as otherwise, even the one missing await may lead to unsaved data. By using callbacks, you can avoid this problem. A lot of libraries for lambdas (apollo-server, etc.) are using callbacks instead of promises as you cannot trust your code without it. We had a LOT of problems in production with that, so we switched to callbacks which makes sure that there is no pending request to database, etc.

kuhe:
I see, I'll put together a patch that clears the timeouts.

kuhe:
The fix is available in https://www.npmjs.com/package/@smithy/node-http-handler/v/3.1.0, if you update your lockfiles and use a recent AWS SDK version that has an open range of `^3.x` for that Smithy package.

[...]

PR #1311, opened 2024-06-17
fix(node-http-handler): add logger option and clear timeout on errors

kuhe:
https://github.com/smithy-lang/smithy-typescript/issues/1305

- add logger option to node-http-handler
- clear socket usage check timeout on error

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -1,6 +1,6 @@
 import { HttpHandler, HttpRequest, HttpResponse } from "@smithy/protocol-http";
 import { buildQueryString } from "@smithy/querystring-builder";
-import type { NodeHttpHandlerOptions } from "@smithy/types";
+import type { Logger, NodeHttpHandlerOptions } from "@smithy/types";
 import { HttpHandlerOptions, Provider } from "@smithy/types";
 import { Agent as hAgent, request as hRequest } from "http";
 import { Agent as hsAgent, request as hsRequest, RequestOptions } from "https";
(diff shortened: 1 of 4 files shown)


haydenbaker (approved):

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -79,11 +85,10 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
          * lockout.
          */
         if (socketsInUse >= maxSockets && requestsEnqueued >= 2 * maxSockets) {
-          console.warn(
-            "@smithy/node-http-handler:WARN",
-            `socket usage at capacity=${socketsInUse} and ${requestsEnqueued} additional requests are enqueued.`,
-            "See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html",
-            "or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config."
+          logger?.warn?.(
```
syall, line 88: nit: do we need the `?` on `warn`?

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -175,7 +182,11 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
       // This warning will be cancelled if the request resolves.
       socketCheckTimeoutId = setTimeout(
         () => {
-          this.socketWarningTimestamp = NodeHttpHandler.checkSocketUsage(agent, this.socketWarningTimestamp);
+          this.socketWarningTimestamp = NodeHttpHandler.checkSocketUsage(
+            agent,
+            this.socketWarningTimestamp,
+            this.config!.logger
```
syall, line 188: So if `config!.logger` is `undefined` or `null`, this won't throw because `logger?.warn?.(`?

syall (approved):

packages/node-http-handler/src/node-http-handler.ts
kuhe: we don't based on the type itself, but users sometimes don't provide the right logger implementation

packages/node-http-handler/src/node-http-handler.ts
kuhe: it's because config is resolved earlier in this method

packages/node-http-handler/src/node-http-handler.ts
kuhe: it's somewhat common and not just hypothetical

Issue #1296, opened 2024-06-03
fetch-http-handler: unable to pass credentials parameter


Zouarfi:
Hello,

When using any SDK generated in the browser, the `fetch-http-handler` package is used.

However, according the [definition](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/fetch-http-handler/src/fetch-http-handler.ts), it is impossible to change the value `credentials` of `fetch`. On SDKv2, it was possible to specify `xhrWithCredentials` (see [Notable changes](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/migrating/notable-changes/)).

Specifying `credentials` can be useful when working with different origin on browsers, especially if you have multiple API.

This can be added by modifying following files:

* Add `credentials?` to [`FetchHttpHandlerOptions` in `types` package](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/types/src/http/httpHandlerInitialization.ts#L69)
* Add `credentials` value to `requestOptions` in file [`fetch-http-handler`](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/fetch-http-handler/src/fetch-http-handler.ts#L104)

Is this change relevant to you ? I will gladly open a Pull Request if it's the case.

PR #1317, opened 2024-06-26
Adding support for setting the fetch API credentials mode

qnesingh:
*Issue #, if available:* [1296](https://github.com/smithy-lang/smithy-typescript/issues/1296)

*Description of changes:* Adds support to set credentials in the fetch API to `include`, `same-origin`, or `omit`.

---

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -77,7 +77,7 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
     }
     const requestTimeoutInMs = this.config!.requestTimeout;
     const keepAlive = this.config!.keepAlive === true;
-    const credentials = this.config!.credentials;
+    const credentials = this.config!.credentials as RequestCredentials;
 
     // if the request was already aborted, prevent doing extra work
     if (abortSignal?.aborted) {
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -77,6 +77,7 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
     }
     const requestTimeoutInMs = this.config!.requestTimeout;
     const keepAlive = this.config!.keepAlive === true;
+    const credentials = this.config!.credentials;
 
     // if the request was already aborted, prevent doing extra work
     if (abortSignal?.aborted) {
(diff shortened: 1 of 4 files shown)


packages/types/src/http/httpHandlerInitialization.ts
```diff
@@ -90,4 +90,9 @@ export interface FetchHttpHandlerOptions {
    * these limitations before enabling keepalive.
    */
   keepAlive?: boolean;
+
+    /** 
+     * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
+     */
+  credentials?: RequestCredentials;
```
kuhe, line 97:
```suggestion
  credentials?: "include" | "omit" | "same-origin"
```

to remove dependency on lib.dom types.

packages/types/src/http/httpHandlerInitialization.ts
```diff
@@ -90,4 +90,9 @@ export interface FetchHttpHandlerOptions {
    * these limitations before enabling keepalive.
    */
   keepAlive?: boolean;
+
+  /** 
+   * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
+   */
+  credentials?: "include" | "omit" | "same-origin" | undefined | string;
```
JordonPhillips, line 97: Nit: maybe credentialsMode instead?

JordonPhillips (approved):

packages/types/src/http/httpHandlerInitialization.ts
qnesingh: `credentials` matches the property on the Fetch request itself. I think it makes sense to keep this property name consistent with how it's used on the request.

kuhe (approved):

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -77,7 +77,7 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
     }
     const requestTimeoutInMs = this.config!.requestTimeout;
     const keepAlive = this.config!.keepAlive === true;
-    const credentials = this.config!.credentials;
+    const credentials = this.config!.credentials as RequestCredentials;
```
kuhe, line 80: the point of the change was to remove usage of a global type, this only moves it to here.

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -77,6 +77,7 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
     }
     const requestTimeoutInMs = this.config!.requestTimeout;
     const keepAlive = this.config!.keepAlive === true;
+    const credentials = this.config!.credentials as RequestInit["credentials"];
```
kuhe, line 80: we're already using `RequestInit`, so `RequestCredentials` is _probably_ globally available too, but I'm being conservative here.

PR #1324, opened 2024-07-02
fix(middleware-apply-body-checksum): fix request copying with clone

syall:
*Issue #, if available:*

N/A

*Description of changes:*

Fixed a subtle bug when request copying, otherwise methods like `clone()` may not be copied over.

---


packages/middleware-apply-body-checksum/src/applyMd5BodyChecksumMiddleware.ts
```diff
@@ -30,13 +30,12 @@ export const applyMd5BodyChecksumMiddleware =
           digest = options.streamHasher(options.md5, body);
         }
 
-        request = {
-          ...request,
-          headers: {
-            ...headers,
-            "content-md5": options.base64Encoder(await digest),
-          },
+        const cloned = request.clone();
```
kuhe, line 33: This isn't a safe assumption. 

`HttpRequest.isInstance(request)` is incorrectly implemented. It only checks that a request is duck-like an HttpRequest **interface**. The `clone` method only exists on the `HttpRequest` **class**.

It wasn't a good idea naming interfaces and classes the same, even worse that they aren't identical types.

kuhe (changes requested):
Note to self: implement static `HttpRequest.clone()` that accepts `HttpRequest` interface. Deprecate instance method.

kuhe (approved):

kuhe:
This is no longer necessary due to https://github.com/smithy-lang/smithy-typescript/pull/1345

kuhe:
re-opening because this PR is still good for code clarity

hpmellema (approved):

PR #1333, opened 2024-07-10
chore(protocol-http): guidance for HttpRequest cloning

kuhe:
This moves `HttpRequest#clone()` (instance method) to `HttpRequest.clone(IHttpRequest)` (static method) and deprecates the original instance method.

`HttpRequest#clone()` cannot safely be called, because the type-guard `HttpRequest.isInstance(unknown)` is not correctly implemented, but cannot safely be changed since it is widely used as-is. 

#### Context:
`HttpRequest.isInstance(unknown) -> boolean is HttpRequest` is actually 
`HttpRequest.isInstance(unknown) -> boolean is IHttpRequest`. 

It returns true when the input satisfies the data/structural interface, but the structural interface does not necessarily include the `.clone()` method.

### DIFF
--- a/.changeset/khaki-seahorses-cough.md
+++ b/.changeset/khaki-seahorses-cough.md
@@ -1,0 +1,5 @@
+---
+"@smithy/protocol-http": patch
+---
+
+add guidance for HttpRequest cloning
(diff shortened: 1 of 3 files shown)


trivikr (approved):

kuhe:
In a later major version, the `HttpRequest.isInstance(unknown)` function can be made less misleading with a non-backwards-compatible change.

milesziemer (approved):

PR #1336, opened 2024-07-12
feat(util-stream): splitStream and headStream utilities

kuhe:
- `splitStream` splits a stream into two streams using `pipe(PassThrough)` or `tee()` depending on Stream type.
- `headStream` collects the first N bytes of a stream.

Use case: peek the head of a stream's copy while leaving the 2nd copy of that original stream intact

Use case: obtain a parallel stream for checksumming

### DIFF
--- a/.changeset/metal-snakes-remember.md
+++ b/.changeset/metal-snakes-remember.md
@@ -1,0 +1,5 @@
+---
+"@smithy/util-stream": minor
+---
+
+add stream splitting to sdkStreamMixin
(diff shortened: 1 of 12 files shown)


kuhe:
I've tested this for memory leaks. As long as the additional split stream reaches reference count 0 it will be garbage collected without needing to explicitly call its destructor or draining it.

sugmanue (approved):

siddsriv (approved):

PR #1338, opened 2024-07-16
add default endpoint to protocol test stub

kuhe:
Due to creating a set of default plugins in https://github.com/smithy-lang/smithy-typescript/pull/1321, an endpoint must be provided for non-Endpoints-RuleSet service models' clients. 

Providing a default endpoint in protocol tests is required for the non-Endpoints-RuleSet case, and also non-breaking otherwise.


smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/protocol-test-stub.ts
```diff
@@ -137,17 +130,18 @@ const equivalentContents = (expected: any, generated: any): boolean => {
...
 
   return true;
-}
+};
 
 const clientParams = {
   region: "us-west-2",
-  credentials: { accessKeyId: "key", secretAccessKey: "secret" }
-}
+  endpoint: "https://localhost/",
```
kuhe, line 137: This is the only change. The rest is formatting to match TypeScript code standards present in the repository.

milesziemer (approved):

trivikr (approved):

PR #1350, opened 2024-07-26
Verify that removeSignalEventListener is a function before calling

trivikr:
*Issue #, if available:*
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/6269

*Description of changes:*
Verifies that removeSignalEventListener is a function before calling

Testing done while debugging in react-native environment in https://github.com/aws/aws-sdk-js-v3/issues/6269#issuecomment-2252850763

---

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -185,7 +185,11 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
         })
       );
     }
-    return Promise.race(raceOfPromises).finally(removeSignalEventListener);
+    return Promise.race(raceOfPromises).finally(() => {
+      if (typeof removeSignalEventListener === "function") {
+        removeSignalEventListener();
+      }
+    });
   }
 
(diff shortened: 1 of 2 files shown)


isaiahshall (approved):

kuhe (approved):

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -185,7 +185,11 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
         })
       );
     }
-    return Promise.race(raceOfPromises).finally(removeSignalEventListener);
+    return Promise.race(raceOfPromises).finally(() => {
+      if (typeof removeSignalEventListener === "function") {
```
kuhe, line 189: initialize `removeSignalEventListener = () => {}` instead?

could you add a comment that this is only needed in react-native?

trivikr: > initialize `removeSignalEventListener = () => {}` instead?

Verified that initializing `removeSignalEventListener` as an empty function works
https://github.com/aws/aws-sdk-js-v3/issues/6269#issuecomment-2253018138

trivikr: Posted a new fix in https://github.com/smithy-lang/smithy-typescript/pull/1351

PR #1351, opened 2024-07-26
Initialize removeSignalEventListener as an empty function

trivikr:
*Issue #, if available:*
* Fixes: https://github.com/aws/aws-sdk-js-v3/issues/6269
* A cleaner alternative to https://github.com/smithy-lang/smithy-typescript/pull/1350

*Description of changes:*
Initialize removeSignalEventListener as an empty function

Testing done in https://github.com/aws/aws-sdk-js-v3/issues/6269#issuecomment-2253018138

---

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -185,7 +185,11 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
         })
       );
     }
-    return Promise.race(raceOfPromises).finally(removeSignalEventListener);
+    return Promise.race(raceOfPromises).finally(() => {
+      if (typeof removeSignalEventListener === "function") {
+        removeSignalEventListener();
+      }
+    });
   }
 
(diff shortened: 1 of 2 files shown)


isaiahshall (approved):

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -127,7 +127,7 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
       requestOptions.keepalive = keepAlive;
     }
 
-    let removeSignalEventListener = null as null | (() => void);
+    let removeSignalEventListener = () => {};
```
kuhe, line 130:
```suggestion
    // react-native promises cannot accept non-function as a then/catch/finally step.
    let removeSignalEventListener = () => {};
```

trivikr: I don't think the comment is necessary, as this fix changes `removeSignalEventListener` type to function?

gosar (approved):

PR #1370, opened 2024-08-16
feat(cbor): add protocol resolution priority system

kuhe:
- adds a system for Protocol selection that has a notion of priority.
- adds an API for controlling overrides to the global priority order and service-specific priority ordering of Protocols

Example Scenario:

Smithy-TS only has 1 inherent protocol at the moment, RPCv2 CBOR.

AWS consumer may want to set a default priority order of JSON->CBOR->XML etc.
Specific AWS services that support both JSON and CBOR may want to prioritize differently from each other.


kuhe:
prior: https://github.com/smithy-lang/smithy-typescript/pull/1280

trivikr (approved):

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/TypeScriptSettingsTest.java
```diff
@@ -87,6 +98,121 @@ private static Stream<Arguments> providePackageDescriptionTestCases() {
         );
     }
 
+    @Test
+    public void resolveServiceProtocol(@Mock Model model,
+                                       @Mock ServiceShape service,
+                                       @Mock ServiceIndex serviceIndex) {
+        TypeScriptSettings subject = new TypeScriptSettings();
+
+        // note: these are mock protocol names.
```
kuhe, line 107:
```suggestion
        // these are mock protocol names.
```

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/TypeScriptSettingsTest.java
```diff
@@ -87,6 +98,121 @@ private static Stream<Arguments> providePackageDescriptionTestCases() {
...
+        );
+
+        {
+            // spec case 1.
+            when(serviceIndex.getProtocols(service)).thenReturn(MapUtils.of(
+                rpcv2Cbor, null,
+                json1_0, null
+            ));
+            ShapeId protocol = subject.resolveServiceProtocol(model, service, protocolShapeIds);
+            // Note: JS customization JSON higher default priority than CBOR.
```
kuhe, line 133:
```suggestion
            // JS customization has JSON at higher default priority than CBOR.
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/ProtocolPriority.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+
+/**
+ * Allows customization of protocol selection for specific services or a global default ordering.
+ */
+public final class ProtocolPriority {
```
sugmanue, line 18: Global mutable state it's a bad pattern, error prone and hard to debug, and, as coded not thread safe. Is there a better way to define this without having a global state?

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/TypeScriptSettingsTest.java
```diff
@@ -87,6 +98,121 @@ private static Stream<Arguments> providePackageDescriptionTestCases() {
         );
     }
 
+    @Test
+    public void resolveServiceProtocol(@Mock Model model,
```
sugmanue, line 102: Consider instead using parametrized tests. This test is testing all the cases instead of having a single test method for each.

sugmanue (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/ProtocolPriority.java
kuhe: turned this into a non-static class and is now instantiated under `TypeScriptSettings`, which is singular in a ProtocolGenerationContext.

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/TypeScriptSettingsTest.java
kuhe: Split the spec cases out into separate test methods

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/protocols/ProtocolPriorityConfig.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+
+/**
+ * Allows customization of protocol selection for specific services or a global default ordering.
+ */
+public final class ProtocolPriorityConfig {
```
sugmanue, line 18: Can we make `ProtocolPriorityConfig`  immutable? Other than in the test, are there any other scenarios in which we need to mutate this after reading?

kuhe: made the class immutable, but settings can still overwrite the entire object

sugmanue (approved):

PR #1374, opened 2024-08-27
Do not take protocol and port from custom Endpoint

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/pull/6410#issuecomment-2311320293

*Description of changes:*
Skips taking protocol and port from custom Endpoint.

The old tests added in https://github.com/aws/aws-sdk-js-v3/pull/4031 were likely incorrect.
Noticed while incorporating the endpoint tests at top-level in https://github.com/aws/aws-sdk-js-v3/pull/6410#issuecomment-2311320293

---

### DIFF
--- a/packages/util-endpoints/src/resolveEndpoint.ts
+++ b/packages/util-endpoints/src/resolveEndpoint.ts
@@ -36,18 +36,6 @@ export const resolveEndpoint = (ruleSetObject: RuleSetObject, options: EndpointR
 
   const endpoint = evaluateRules(rules, { endpointParams, logger, referenceRecord: {} });
 
-  if (options.endpointParams?.Endpoint) {
-    // take protocol and port from custom Endpoint if present.
-    try {
-      const givenEndpoint = new URL(options.endpointParams.Endpoint as string);
-      const { protocol, port } = givenEndpoint;
-      endpoint.url.protocol = protocol;
-      endpoint.url.port = port;
-    } catch (e) {
-      // ignored
-    }
-  }
-
   options.logger?.debug?.(`${debugId} Resolved endpoint: ${toDebugString(endpoint)}`);
 
   return endpoint;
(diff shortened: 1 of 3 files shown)


kuhe (approved):

packages/util-endpoints/src/resolveEndpoint.ts
```diff
@@ -36,18 +36,6 @@ export const resolveEndpoint = (ruleSetObject: RuleSetObject, options: EndpointR
...
-    try {
-      const givenEndpoint = new URL(options.endpointParams.Endpoint as string);
-      const { protocol, port } = givenEndpoint;
-      endpoint.url.protocol = protocol;
-      endpoint.url.port = port;
-    } catch (e) {
-      // ignored
-    }
-  }
-
```
kuhe, line 50: To clarify, this is not the removal of the user's ability to set a custom protocol and port. 

If a service's ruleset follows the base template, one of the initial conditions is to check for a "built-in" user-supplied `Endpoint` parameter, or `endpoint` typically in TypeScript SDKs generated from Smithy.

It looks like this:
```json
"conditions": [
  {
    "fn": "isSet",
    "argv": [
      {
        "ref": "Endpoint"
      }
    ]
  },
  {
    "fn": "parseURL",
    "argv": [
      {
        "ref": "Endpoint"
      }
    ],
    "assign": "url"
  }
],
```

And will resolve an endpoint template such as 
```json
"endpoint": {
  "url": "{url#scheme}://{url#authority}{url#path}",
  "...": "..."
}
```

This preserves the user's input scheme/protocol and port as part of authority.

----

In rare cases where the service ruleset ignores the user input endpoint's scheme/port, the fix would be either to patch the ruleset or override the `endpointProvider` configuration field.

kstich (approved):

PR #1382, opened 2024-08-29
fix(endpoints): resolve service-specific endpoint once per client

kuhe:

- resolve service specific endpoint once per config (client instance)
- this avoids the ENV and file system read from `loadConfig`

### DIFF
--- a/packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
+++ b/packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
@@ -39,9 +39,19 @@ export const getEndpointFromInstructions = async <
   context?: HandlerExecutionContext
 ): Promise<EndpointV2> => {
   if (!clientConfig.endpoint) {
-    const endpointFromConfig = await getEndpointFromConfig(clientConfig.serviceId || "");
+    let endpointFromConfig: string | undefined;
+
+    // This field is guaranteed by the type indicated by the config resolver, but is new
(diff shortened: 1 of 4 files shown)


packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
```diff
@@ -39,9 +39,19 @@ export const getEndpointFromInstructions = async <
   context?: HandlerExecutionContext
 ): Promise<EndpointV2> => {
   if (!clientConfig.endpoint) {
-    const endpointFromConfig = await getEndpointFromConfig(clientConfig.serviceId || "");
```
kuhe, line 42: the issue was that this call involved a call to `loadConfig` for every request, which _calls_ `memoize` rather than calling a memoized function, leading to repeated calls to fs.

trivikr (approved):

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
kuhe: this delayed every request by around 50ms in my testing

milesziemer (approved):

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
kuhe: edit: "slurpFile" has a cache, but it is several functions/modules deep in terms of unnecessary calls

PR #1383, opened 2024-08-30
feat(smithy-client): add handler cache

kuhe:

This adds a cache for the handler functions resolved from calls to 

```ts
command.resolveMiddleware(client.middlewareStack, client.config, (request) options)
```

when options is undefined. This occurs in the request path and can take 1-2ms, but is redundant for any client/command pair in most cases. 

Despite being cached, the handler functions has a reference to the client config, so any changes there will still be visible to the middleware within the resolved handler. ~Modifying a client's middlewareStack invalidates the cache.~

The cache is opt-in and also unique per client, so users retain control that way.

### DIFF
--- a/packages/smithy-client/src/client.ts
+++ b/packages/smithy-client/src/client.ts
@@ -25,6 +25,22 @@ export interface SmithyConfiguration<HandlerOptions> {
    * @internal
    */
   readonly apiVersion: string;
+  /**
+   * @public
+   *
+   * Default false.
+   *
+   * When true, the client will only resolve the middleware stack once per
+   * Command class. This means modifying the middlewareStack of the
+   * command or client after requests have been made will not be
+   * recognized.
+   *
+   * Calling client.destroy() also clears this cache.
+   *
+   * Enable this only if needing the additional time saved (0-1ms per request)
+   * and not needing middleware modifications between requests.
+   */
+  cacheMiddleware?: boolean;
 }
 
 /**
@@ -33,6 +49,7 @@ export interface SmithyConfiguration<HandlerOptions> {
 export type SmithyResolvedConfiguration<HandlerOptions> = {
   requestHandler: RequestHandler<any, any, HandlerOptions>;
   readonly apiVersion: string;
+  cacheMiddleware?: boolean;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/smithy-client/src/client.ts
```diff
@@ -87,6 +113,7 @@ export class Client<
   }
 
   destroy() {
-    if (this.config.requestHandler.destroy) this.config.requestHandler.destroy();
+    this.config.requestHandler.destroy?.();
+    clientHandlers.delete(this);
```
kuhe, line 117: if the concern is affecting the shape of the class, we can probably use `private handler?` i.e. optionality in the property, for backwards compatibility.

packages/smithy-client/src/client.ts
```diff
@@ -44,11 +45,21 @@ export class Client<
...
+      delete this.handlers;
+    },
+  });
+  /**
+   * May be used to cache the resolved handler function for a Command class.
+   */
+  private handlers?: WeakMap<Function, Handler<any, any>> | undefined;
+  private configRef?: ResolvedClientConfiguration | undefined;
+
+  constructor(public readonly config: ResolvedClientConfiguration) {
```
milesziemer, line 59: Maybe just me not being familiar with this syntax, but it looks like `this.configRef` is what's assigned in the ctor, not `this.config`. Is `this.config` still accessible even though it isn't declared or assigned to? I.e. in the `destroy` method.

kuhe: this is shorthand, these are equivalent

```ts
public constructor(public prop: Type) {}

public constructor(prop: Type) {
  this.prop = prop;
}
```

configRef is for detecting reference changes to the client's config property, but this whole PR is kind of iffy. Probably not worth putting change detectors on so much stuff just to save 1ms of handler resolution time, and it's leaky.

kuhe:
putting this into draft while I reevaluate whether this change is worth it or not

packages/smithy-client/src/client.ts
kuhe: no longer applicable in latest revision

milesziemer (approved):

siddsriv (approved):

PR #1385, opened 2024-09-02
feat(util-endpoint): add endpoint ruleset cache

kuhe:

This PR creates a cache for endpoint ruleset resolution.

As one would expect, the input and cache key are the endpoint params and the output and cache value is the endpoint.

Some services declare endpoint parameters in excess of what is actually evaluated in the ruleset, so codegen provides the explicit pared param list to the cache instance.

- [x] codegen to connect EndpointCache to the service's endpointResolver.ts file contents.
- [x] codegen to determine what the pared parameter list is to avoid cache pollution from irrelevant params

### DIFF
--- a/packages/util-endpoints/src/cache/EndpointCache.ts
+++ b/packages/util-endpoints/src/cache/EndpointCache.ts
@@ -15,7 +15,7 @@ export class EndpointCache {
    *                 before keys are dropped.
    * @param [params] - list of params to consider as part of the cache key.
    *
-   * If the params list is not populated, all object keys will be considered.
+   * If the params list is not populated, no caching will happen.
    * This may be out of order depending on how the object is created and arrives to this class.
    */
   public constructor({ size, params }: { size?: number; params?: string[] }) {
@@ -57,10 +57,16 @@ export class EndpointCache {
     return this.data.size;
   }
 
+  /**
+   * @returns cache key or false if not cachable.
+   */
   private hash(endpointParams: EndpointParams): string | false {
     let buffer = "";
-    const params = this.parameters.length ? this.parameters : Object.keys(endpointParams);
-    for (const param of params) {
+    const { parameters } = this;
+    if (parameters.length === 0) {
+      return false;
+    }
+    for (const param of parameters) {
       const val = String(endpointParams[param] ?? "");
       if (val.includes("|;")) {
         return false;
--- a/packages/util-endpoints/src/cache/EndpointCache.ts
+++ b/packages/util-endpoints/src/cache/EndpointCache.ts
@@ -11,6 +11,8 @@ export class EndpointCache {
   private parameters: string[] = [];
 
   /**
+   * @param [size] - desired average maximum capacity. A buffer of 10 additional keys will be allowed
+   *                 before keys are dropped.
    * @param [params] - list of params to consider as part of the cache key.
    *
    * If the params list is not populated, all object keys will be considered.
@@ -30,6 +32,10 @@ export class EndpointCache {
    */
   public get(endpointParams: EndpointParams, resolver: () => EndpointV2): EndpointV2 {
     const key = this.hash(endpointParams);
+    if (key === false) {
+      return resolver();
+    }
+
     if (!this.data.has(key)) {
       if (this.data.size > this.capacity + 10) {
         const keys = this.data.keys();
@@ -51,11 +57,15 @@ export class EndpointCache {
     return this.data.size;
   }
 
-  private hash(endpointParams: EndpointParams): string {
+  private hash(endpointParams: EndpointParams): string | false {
     let buffer = "";
     const params = this.parameters.length ? this.parameters : Object.keys(endpointParams);
     for (const param of params) {
-      buffer += endpointParams[param] ?? "" + "|";
+      const val = String(endpointParams[param] ?? "");
+      if (val.includes("|;")) {
+        return false;
+      }
+      buffer += val + "|;";
     }
     return buffer;
   }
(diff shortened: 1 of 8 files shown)


kuhe:
The S3 resolver would now look like this below.
Notably absent is `Key` from the parameter list.

```ts
import { awsEndpointFunctions } from "@aws-sdk/util-endpoints";
import { EndpointV2, Logger } from "@smithy/types";
import { customEndpointFunctions, EndpointCache, EndpointParams, resolveEndpoint } from "@smithy/util-endpoints";

import { EndpointParameters } from "./EndpointParameters";
import { ruleSet } from "./ruleset";

const cache = new EndpointCache({
  size: 50,
  params: [
    "Accelerate",
    "Bucket",
    "DisableAccessPoints",
    "DisableMultiRegionAccessPoints",
    "DisableS3ExpressSessionAuth",
    "Endpoint",
    "ForcePathStyle",
    "Region",
    "UseArnRegion",
    "UseDualStack",
    "UseFIPS",
    "UseGlobalEndpoint",
    "UseObjectLambdaEndpoint",
    "UseS3ExpressControlEndpoint",
  ],
});

export const defaultEndpointResolver = (
  endpointParams: EndpointParameters,
  context: { logger?: Logger } = {}
): EndpointV2 => {
  return cache.get(endpointParams as EndpointParams, () =>
    resolveEndpoint(ruleSet, {
      endpointParams: endpointParams as EndpointParams,
      logger: context.logger,
    })
  );
};

customEndpointFunctions.aws = awsEndpointFunctions;
```

kuhe:
In one benchmark, a batch of 10 parallel Amazon S3 HeadObject requests completed in 80ms, as opposed to 86ms without this cache.

packages/util-endpoints/src/cache/EndpointCache.ts
```diff
@@ -0,0 +1,62 @@
...
+  private parameters: string[] = [];
+
+  /**
+   * @param [params] - list of params to consider as part of the cache key.
+   *
+   * If the params list is not populated, all object keys will be considered.
+   * This may be out of order depending on how the object is created and arrives to this class.
+   */
+  public constructor({ size, params }: { size?: number; params?: string[] }) {
+    this.capacity = size ?? 50;
```
trivikr, line 20: nit: This is already initialized as `50`on line 9, and doesn't have to be overridden
```suggestion
    if (size) {
      this.capacity = size;
    }
```

trivikr (approved):

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinderTest.java
```diff
@@ -0,0 +1,114 @@
...
+                                },
+                                true
+                              ]
+                            },
+                            true
+                          ]
+                        }
+                      ],
+                      "endpoint": {
+                        "url": "https://www.{BasicParameter}.{UrlOnlyParameter}.com",
```
milesziemer, line 89: Could you include a test with a more complex url, maybe with a [template shorthand](https://smithy.io/2.0/additional-specs/rules-engine/specification.html#template-shorthand), as a sanity check on the regex and to make sure we don't break it down the road?

packages/util-endpoints/src/cache/EndpointCache.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  public size() {
+    return this.data.size;
+  }
+
+  private hash(endpointParams: EndpointParams): string {
+    let buffer = "";
+    const params = this.parameters.length ? this.parameters : Object.keys(endpointParams);
+    for (const param of params) {
+      buffer += endpointParams[param] ?? "" + "|";
```
milesziemer, line 58: Some questions about this hashing method:
1. Does the specific `param` not matter for the hash? At least in the case where `params` is going to be `Object.keys(endpointParams)`, seems like it could matter
2. Can `endpointParams[param]` contain a `|` that would mess with the hash? 
3. Should anything be appended in the case where the key `param` doesn't exist? Might matter in the same case as 1.

Maybe these are edge cases that don't sufficiently impact the effectiveness of the hash though.

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinderTest.java
kuhe: added test case for shorthand, which I didn't know existed

packages/util-endpoints/src/cache/EndpointCache.ts
kuhe: 1. params needs to be consistently sorted, so parameters are alphabetized by codegen. In the case where Object.keys are used, this is best-effort since they are not guaranteed to be in any order.
2. a parameter with a delimiter in it can affect the hash, yes. `A="|",B=void` and `A=void,B="|"` would both hash to `|||`. I'm going to make the delimiter `|;` and abandon caching if this sequence exists in a value.
3. A delimiter as placeholder is enough in this case.

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinderTest.java
milesziemer: I mean one so that it has the `#` in the url, sorry. Just so we make sure those kinds of urls are matched by the regex.

milesziemer: Nevermind, practically speaking we're only looking for endpoint param names, so the object path won't matter.

PR #1397, opened 2024-09-10
feat(fetch-http-handler): add customization option for requestInit

kuhe:
*Issue #, if available:*

*Description of changes:*

---


packages/types/src/http/httpHandlerInitialization.ts
```diff
@@ -97,4 +98,30 @@ export interface FetchHttpHandlerOptions {
...
+   * Cache settings for fetch.
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Request/cache
+   */
+  cache?: "default" | "force-cache" | "no-cache" | "no-store" | "only-if-cached" | "reload";
+
+  /**
+   * An optional function that produces additional RequestInit
+   * parameters for each httpRequest.
+   *
+   * This is applied last and merge-overrides any other values
```
sugmanue, line 112: Nit: about the wording, the "merge" part is confusing.

sugmanue (approved):

siddsriv (approved):

packages/types/src/http/httpHandlerInitialization.ts
kuhe, line 112:
```suggestion
   * This is applied last via merging with Object.assign() and overwrites other values
```

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -127,6 +129,10 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerConfig> {
       requestOptions.keepalive = keepAlive;
     }
 
+    if (typeof this.config.requestInit === "function") {
+      Object.assign(requestOptions, this.config.requestInit(request));
```
trivikr, line 133: We're already setting `cache` on line 115 from config.
Why do we need requestInit?

kuhe: in case other options need customization, 

```
interface RequestInit {
    /** A BodyInit object or null to set request's body. */
    body?: BodyInit | null;
    /** A string indicating how the request will interact with the browser's cache to set request's cache. */
    cache?: RequestCache;
    /** A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials. */
    credentials?: RequestCredentials;
    /** A Headers object, an object literal, or an array of two-item arrays to set request's headers. */
    headers?: HeadersInit;
    /** A cryptographic hash of the resource to be fetched by request. Sets request's integrity. */
    integrity?: string;
    /** A boolean to set request's keepalive. */
    keepalive?: boolean;
    /** A string to set request's method. */
    method?: string;
    /** A string to indicate whether the request will use CORS, or will be restricted to same-origin URLs. Sets request's mode. */
    mode?: RequestMode;
    priority?: RequestPriority;
    /** A string indicating whether request follows redirects, results in an error upon encountering a redirect, or returns the redirect (in an opaque fashion). Sets request's redirect. */
    redirect?: RequestRedirect;
    /** A string whose value is a same-origin URL, "about:client", or the empty string, to set request's referrer. */
    referrer?: string;
    /** A referrer policy to set request's referrerPolicy. */
    referrerPolicy?: ReferrerPolicy;
    /** An AbortSignal to set request's signal. */
    signal?: AbortSignal | null;
    /** Can only be null. Used to disassociate request from any Window. */
    window?: null;
}
```

trivikr (approved):

PR #1403, opened 2024-09-13
fix(types): add interface stub for RequestInit

kuhe:
- adds interface stub for `RequestInit` global

`RequestInit` was already used in the fetch-http-handler, but that is only conditionally compiled if the user imports it or bundles web-based code that imports the browser runtimeConfig files. 

The usage of `RequestInit` in the types package, which is not conditionally loaded, must be accompanied by the interface merging stub.

### DIFF
--- a/.changeset/forty-cooks-deny.md
+++ b/.changeset/forty-cooks-deny.md
@@ -1,0 +1,5 @@
+---
+"@smithy/types": patch
+---
+
+add interface stub for browser RequestInit type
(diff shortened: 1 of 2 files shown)


trivikr (approved):

sugmanue (approved):

kuhe:
revision 1: changed to use the `RequestInit` type unconditionally instead of `never` when its a stub

PR #1404, opened 2024-09-14
fix(signature-v4): sort query parameter keys after encoding

r-izumida:
When signing an HTTP request with SigV4, an incorrect signature will be generated if the query parameter key contains characters that need to be encoded.

The following query parameter reproduces it: `?params[page]=1&params[pageSize]=20`

According to the [AWS documentation][canonical-query-params], it seems that query parameters need to be sorted after encoding.

> `CanonicalQueryString` – The URI-encoded query string parameters. You URI-encode each name and value individually. You must also sort the parameters in the canonical query string alphabetically by key name. The sorting occurs after encoding.

[canonical-query-params]: https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html#create-canonical-request

*Description of changes:*

* Sort query parameter keys after URI encoding in `getCanonicalQuery`

---

### DIFF
--- a/.changeset/ninety-hairs-obey.md
+++ b/.changeset/ninety-hairs-obey.md
@@ -1,0 +1,5 @@
+---
+"@smithy/signature-v4": patch
+---
+
+fix: sort query parameter keys after encoding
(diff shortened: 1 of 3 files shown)


kuhe (approved):

milesziemer (approved):

kuhe:
released in https://www.npmjs.com/package/@smithy/signature-v4/v/4.1.4

PR #1409, opened 2024-09-19
feat(util-stream): create checksum stream adapters

kuhe:
creates the `createChecksumStream` function to wrap streams with a checksum verification that throws when the stream is read to completion if the checksum verification fails.

Related to flexible checksums in AWS SDKs, but implementation is generic here. 

API reference:
- https://developer.mozilla.org/en-US/docs/Web/API/TransformStream
- https://nodejs.org/api/stream.html#duplex-and-transform-streams

### DIFF
--- a/packages/util-stream/src/checksum/createChecksumStream.ts
+++ b/packages/util-stream/src/checksum/createChecksumStream.ts
@@ -103,9 +103,13 @@ class ChecksumStream extends Duplex {
    * calculate a step update of the checksum.
    */
   public _write(chunk: Buffer, encoding: string, callback: (err?: Error) => void): void {
-    this.checksum.update(chunk);
-    this.push(chunk);
-    callback();
+    try {
+      this.checksum.update(chunk);
+      this.push(chunk);
+    } catch (e: unknown) {
+      return callback(e as Error);
+    }
+    return callback();
   }
 
   /**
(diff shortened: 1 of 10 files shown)


packages/util-stream/src/checksum/createChecksumStream.ts
```diff
@@ -0,0 +1,133 @@
...
+          new Error(
+            `Checksum mismatch: expected "${this.expectedChecksum}" but received "${received}"` +
+              ` in response header "${this.checksumSourceLocation}".`
+          )
+        );
+      }
+    } catch (e: unknown) {
+      callback(e as Error);
+    }
+    this.push(null);
```
kuhe, line 131: probably needs to call `callback()` after `this.push(null)`

packages/util-stream/src/checksum/createChecksumStream.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * @internal
+ *
+ * Wrapper for throwing checksum errors for streams without
+ * buffering the stream.
+ *
+ */
+export class ChecksumStream extends Duplex {
```
trivikr, line 59: Move this to it's own file, say `ChecksumStream.ts`?

trivikr (approved):

kstich (approved):

PR #1411, opened 2024-09-20
fix(node-http-handler): omit setting cache setting on request init when using default value

### LINKED ISSUE #1407: Error: The 'cache' field on 'RequestInitializerDict' is not implemented.
Started seeing this error when using AWS S3 SDK to GetObject in a Cloudflare Pages App, which runs on workerd:

```
✘ [ERROR] Error: The 'cache' field on 'RequestInitializerDict' is not implemented.

 at FetchHttpHandler.handle
  (file:///.../node_modules/@smithy/fetch-http-handler/dist-es/fetch-http-handler.js:77:30)
      at null.<anonymous> (async
```

I will attempt a downgrade. I may try a standalone app to share.

Update pinning to 3.2.5 resolves the error.

nicksrandall:
Fixes #1407 

Omit setting cache setting on request init when using default value

### DIFF
--- a/.changeset/good-cheetahs-wait.md
+++ b/.changeset/good-cheetahs-wait.md
@@ -1,0 +1,5 @@
+---
+"@smithy/fetch-http-handler": patch
+---
+
+Omit setting cache setting on request init when using default value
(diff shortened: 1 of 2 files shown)


nicksrandall:
@kuhe do you see any issues with this?

kuhe (approved):

milesziemer (approved):

PR #1412, opened 2024-09-20
fix(codegen): allow empty string field values for headers

siddsriv:
requires: https://github.com/smithy-lang/smithy-typescript/pull/1414

*Issue #, if available:*
Internal JS-5444

*Description of changes:*
allows headers to have empty values. 

for more details, relevant RFC: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5

upstream of this: https://github.com/smithy-lang/smithy/pull/2403 protocol test change

---


smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/integration/http-binding-utils.ts
```diff
@@ -1,7 +1,2 @@
 const isSerializableHeaderValue = (value: any): boolean =>
-  value !== undefined &&
-  value !== null &&
-  value !== "" &&
-  (!Object.getOwnPropertyNames(value).includes("length") ||
-    value.length != 0) &&
-  (!Object.getOwnPropertyNames(value).includes("size") || value.size != 0);
+  value != undefined && value != null;
```
kuhe, line 2: `value != null` is enough, since `undefined` will match

nvllkwng, line 2: 1. we should keep `!==` as we did before for null and undef checks
2. Are we not worried about `value.size` anymore?

Le0Michine: is it possible to cover this with some unit tests? would be nice to see what cases we are considering

kuhe: 1. no
2. yes, value size will be ignored

kuhe: added unit tests

kuhe (approved):

kstich (approved):

PR #1417, opened 2024-09-24
fix(protocols): fix header serde, handle unset union payloads

kuhe:
this is a set of changes revealed by re-enabling the skipped protocol tests in https://github.com/aws/aws-sdk-js-v3/pull/6512


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -2466,16 +2466,29 @@ private HttpBinding readPayload(
...
+        if (target instanceof UnionShape) {
+            writer.openBlock(
+                "if (Object.keys(data ?? {}).length) {",
+                "}",
+                () -> {
+                    writer.write("contents.$L = __expectUnion($L);", binding.getMemberName(), getOutputValue(context,
+                        Location.PAYLOAD, "data", binding.getMember(), target));
+                }
+            );
+        } else {
```
kuhe, line 2487: old:
```ts
  const data: Record<string, any> | undefined = __expectUnion(await parseBody(output.body, context));
  contents.nested = de_UnionPayload(data, context);
```
(does not handle empty output correctly)

new:
```ts
  const data: Record<string, any> | undefined = await parseBody(output.body, context);
  if (Object.keys(data ?? {}).length) {
    contents.nested = __expectUnion(de_UnionPayload(data, context));
  }
```

packages/smithy-client/src/split-header.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * @param value - header string value.
+ * @returns value split by commas that aren't in quotes.
+ */
+export const splitHeader = (value: string): string[] => {
```
kuhe, line 5: This takes over `.split(",")` on the deserialization side. Splitting by commas is not enough, because commas within quotes do not count as delimiters, and once split, the items wrapped by double quotes should have those quotes removed.

Behavior specified here: https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L229-L232

```smithy
{
        id: "RestJsonInputAndOutputWithQuotedStringHeaders",
        documentation: "Tests responses with string list header bindings that require quoting",
        protocol: restJson1,
        code: 200,
        headers: {
            "X-StringList": "\"b,c\", \"\\\"def\\\"\", a"
        },
        params: {
            headerStringList: ["b,c", "\"def\"", "a"]
        }
}
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -517,7 +521,17 @@ private void writeHttpHeaderAssertions(HttpMessageTestCase testCase) {
...
+                writer.write("""
+                    expect(
+                      r.headers[$S].replaceAll(", ", ",")
+                    ).toBe(
+                      $S.replaceAll(", ", ",")
+                    );
+                    """, header, value);
+            } else {
+                writer.write("expect(r.headers[$S]).toBe($S);", header, value);
+            }
```
kuhe, line 534: I propose that we ignore comma delimiter spacing rather than changing the upstream Smithy test assertion or our and other language implementations.

trivikr (approved):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
milesziemer: Ah this is a pain. Although wouldn't it be easier to just add a space [here](https://github.com/smithy-lang/smithy-typescript/blob/80515513f9b4ff3e6140d335a8edfb2126b8c352/packages/middleware-compression/src/compressionMiddleware.ts#L85)?

packages/smithy-client/src/quote-header.ts
```diff
@@ -0,0 +1,11 @@
...
+ * @public
+ * @param part - header list element
+ * @returns quoted string if part contains delimiter.
+ */
+export function quoteHeader(part: string) {
+  if (part.includes(",") || part.includes('"')) {
+    part = `"${part.replace(/"/g, '\\"')}"`;
+  }
+  return part;
+}
```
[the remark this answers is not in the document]
milesziemer: Would this double-escape `part`s that already escaped their `"`'s?

packages/smithy-client/src/split-header.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    prevChar = char;
+  }
+
+  values.push(value.slice(anchor));
+
+  return values.map((v) => {
+    v = v.trim();
+    const z = v.length;
+    if (v[0] === `"` && v[z - 1] === `"`) {
```
milesziemer, line 37: Can `v` be empty? What if there's sequential `,`, or `,` only separated by ws?

milesziemer (changes requested):

packages/smithy-client/src/quote-header.ts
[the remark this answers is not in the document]
kuhe: Yes, but this is how all user input is preserved, which I believe is the intent.

user: `mystring\"hasquotesinit`

serialized: `"mystring\\"hasquotesinit"`

packages/smithy-client/src/split-header.ts
kuhe: added more test cases to describe what happens, i.e. `,,` is 3 empty strings, and so is `  ,   ,   `.

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
kuhe: _should_ be fine, so I pushed a commit to do so.

packages/smithy-client/src/quote-header.ts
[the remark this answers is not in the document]
kuhe: in JS the two would be `...\\"...` and `"...\\\\"..."`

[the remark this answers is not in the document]
milesziemer: Ah, I see now. It is wrapping the whole thing in `"`, so yea makes sense

milesziemer (approved):

PR #1424, opened 2024-10-02
feat: allow feature identification on the smithy context

kuhe:
Adds a typed `features` map as a subsection of the Smithy Context, which is an object that has a lifecycle corresponding to an SDK request. 

This map is not written to or read from in this PR, but integrations will be added to identify certain features being used. It will be read downstream in the SDK user agent middleware.

This functionality is not strictly relevant to the standalone Smithy runtime, but I am suggesting this is the least obtrusive way to implement it.

downstream: https://github.com/aws/aws-sdk-js-v3/pull/6536

### DIFF
--- a/packages/types/src/feature-ids.ts
+++ b/packages/types/src/feature-ids.ts
@@ -1,0 +1,18 @@
+/**
+ * @internal
+ */
+export type SmithyFeatures = Partial<{
+  RESOURCE_MODEL: "A";
+  WAITER: "B";
+  PAGINATOR: "C";
+  RETRY_MODE_LEGACY: "D";
+  RETRY_MODE_STANDARD: "E";
+  RETRY_MODE_ADAPTIVE: "F";
+  GZIP_REQUEST_COMPRESSION: "L";
+  PROTOCOL_RPC_V2_CBOR: "M";
+  ENDPOINT_OVERRIDE: "N";
+  SIGV4A_SIGNING: "S";
+  CREDENTIALS_CODE: "e";
+  CREDENTIALS_HTTP: "z";
+  CREDENTIALS_IMDS: "0";
+}>;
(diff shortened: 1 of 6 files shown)


packages/types/src/feature-ids.ts
```diff
@@ -0,0 +1,18 @@
...
+  RETRY_MODE_LEGACY: "D";
+  RETRY_MODE_STANDARD: "E";
+  RETRY_MODE_ADAPTIVE: "F";
+  GZIP_REQUEST_COMPRESSION: "L";
+  PROTOCOL_RPC_V2_CBOR: "M";
+  ENDPOINT_OVERRIDE: "N";
+  SIGV4A_SIGNING: "S";
+  CREDENTIALS_CODE: "e";
+  CREDENTIALS_HTTP: "z";
+  CREDENTIALS_IMDS: "0";
```
kuhe, line 17: some of these credentials features aren't technically Smithy, but they live in this repository as of the migration from AWS SDK.

kuhe: The full list can be seen in the paired PR https://github.com/aws/aws-sdk-js-v3/pull/6536/files

under a similar file `types/src/feature-ids.ts`

trivikr (approved):

packages/types/src/feature-ids.ts
JordonPhillips: > as of the migration from AWS SDK

I'm not sure what that means. If there's a separate list over there, why do AWSisms need to live here?

kuhe: All the JS code used to be in AWS SDK. The generic or core runtime packages were moved over to smithy but it wasn't completely separated and the instance metadata provider exists in the smithy-typescript repo. So, that feature may need to be known by the Smithy list of features.

However, I think I can remove credentials imds and credentials http feature detection from smithy and do it in the SDK code. I've pushed a commit to remove these two.

JordonPhillips (approved):

PR #1426, opened 2024-10-04
fix: skip streaming data for union check in readPayload

kuhe:
fix conditional branching for union readPayload codegen


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -2479,7 +2481,7 @@ private HttpBinding readPayload(
                     target.getType()));
         }
 
-        if (target instanceof UnionShape) {
+        if (!isStreaming && target instanceof UnionShape) {
```
kuhe, line 2484: this is to match the condition on line 2472

```java
} else if (target instanceof UnionShape) {
```

which isn't only `target instanceof UnionShape` but also that the prior if branch was not entered.

trivikr (approved):

milesziemer (approved):

PR #1459, opened 2024-11-22
fix(node-http-handler): skip sending body without waiting for a timeout on response, if "expect" request header with "100-continue" is provided

bolt-juri-gavshin:
*Issue #, if available:*

*Description of changes:*

Currently, S3 in AWS-SDK uses "expect: 100-continue" header.
In case of errors, for example "403 Forbidden", the request fails only on timeout. This happens because while sendRequestBody is waiting for "continue" event, a "response" event is fired and "continue" is not fired at all.

---

### DIFF
--- a/.changeset/hungry-eels-accept.md
+++ b/.changeset/hungry-eels-accept.md
@@ -1,0 +1,5 @@
+---
+"@smithy/node-http-handler": patch
+---
+
+skip sending body without waiting for a timeout on response, if "expect" request header with "100-continue" is provided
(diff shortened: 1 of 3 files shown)


kuhe (approved):

kuhe:
I don't have an ETA, but it may be a few weeks before this is released due to Thanksgiving holiday and re:Invent.

kuhe:
This was published in v3.3.2 of `@smithy/node-http-handler`.

https://github.com/smithy-lang/smithy-typescript/blob/main/packages/node-http-handler/CHANGELOG.md

Issue #1446, opened 2024-11-07
ERR_HTTP2_STREAM_CANCEL is not retried


jtmthf:
I have a scenario when using @aws-sdk/client-kinesis where it occasionally throws an error with the code `ERR_HTTP2_STREAM_CANCEL` and cause with code `ETIMEDOUT`. Looking at the `service-error-classification` package, an error with code `ETIMEDOUT` would normally be retried, however being that node now wraps the timeout error in an http2 error, it isn't properly handled.

Below are the properties from one of the errors thrown.

property | value
-- | --
err.cause.address | [REDACTED]
err.cause.code | ETIMEDOUT
err.cause.errno | -110
err.cause.message | connect ETIMEDOUT IP_ADDRESS:443
err.cause.port | 443
err.cause.stack | Error: connect ETIMEDOUT IP_ADDRESS:443     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1607:16)
err.cause.syscall | connect
err.cause.type | Error
err.code | ERR_HTTP2_STREAM_CANCEL
err.message | The pending stream has been canceled (caused by: connect ETIMEDOUT IP_ADDRESS:443)
err.stack | Error [ERR_HTTP2_STREAM_CANCEL]: The pending stream has been canceled (caused by: connect ETIMEDOUT IP_ADDRESS:443)     at closeSession (node:internal/http2/core:1175:20)     at ClientHttp2Session.destroy (node:internal/http2/core:1565:5)     at TLSSocket.socketOnError (node:internal/http2/core:3028:13)     at TLSSocket.emit (node:events:519:28)     at emitErrorNT (node:internal/streams/destroy:169:8)     at emitErrorCloseNT (node:internal/streams/destroy:128:3)     at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
err.type | Error

In order to resolve, one option may be to add `ERR_HTTP2_STREAM_CANCEL` along with others to `NODEJS_TIMEOUT_ERROR_CODES`. Another maybe safer option could be to have error classification recursively walk the `cause` property of errors and consider those for classification.

PR #1461, opened 2024-11-26
add support for error cause in transient error checks

jtmthf:
*Issue #1446 

*Description of changes:*

Updates `isTransientError` to recursively check the `cause` property if the underlying error is also transient. This continues until either the error cause chain is exhausted or a causing error is determined to be transient. This correctly annotates node http2 errors which wrap `ETIMEDOUT` errors.

---

### DIFF
--- a/.changeset/good-dots-run.md
+++ b/.changeset/good-dots-run.md
@@ -1,0 +1,6 @@
+---
+"@smithy/service-error-classification": patch
+"@smithy/types": patch
+---
+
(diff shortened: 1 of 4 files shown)


kuhe (approved):

kuhe:
Published in 3.0.11 of `@smithy/service-error-classification`. 

https://github.com/smithy-lang/smithy-typescript/blob/main/packages/service-error-classification/CHANGELOG.md

PR #1466, opened 2024-12-05
fix: binding the methods to the context

### LINKED ISSUE #1465: Getting `TypeError: Illegal invocation` when calling any AWS SDK method in a browser window in Electron
Hi, after updating the AWS SDK v3 to the latest version I started getting ``TypeError: Illegal invocation` every time the sdk is invoked (eg on `client.send()`).

I narrowed down the problem to the `timings.setTimeout` function. Timing is created as 
```
timings = {
  setTimeout: setTimeout,
  clearTimeout: clearTimeout,
}
```

(see https://github.com/smithy-lang/smithy-typescript/blob/b52b4e8c0a9a43aa5c23f73b65c02df93aeb6cfe/packages/node-http-handler/src/timing.ts#L5-L8)

This causes the two functions to loose the context (`this`) and cause the Illegal invocation error. I verified that reverting back to an SDK version earlier than the date this change was introduced solved the issue.

massi-ang:
*Issue #, if available:*
Fixes #1465 
*Description of changes:*
Added context binding

---

### DIFF
--- a/.changeset/large-squids-rush.md
+++ b/.changeset/large-squids-rush.md
@@ -1,0 +1,5 @@
+---
+"@smithy/node-http-handler": patch
+---
+
+Added context binding to the setTimeout and clearTimeout functions
--- a/packages/node-http-handler/src/timing.ts
+++ b/packages/node-http-handler/src/timing.ts
@@ -3,6 +3,6 @@
  * For test spies.
  */
 export const timing = {
-  setTimeout: (cb, timeout) => setTimeout(cb, timeout),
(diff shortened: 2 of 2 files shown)


massi-ang:
Apparently this is an issue only with Electron, or at least the setup I am using where, the `node-http-handler` is being used. I do not observe the same behaviour in a pure browser application.
Assuming you need the `timing` variable, the correct  fix is to use arrow functions (`bind` did not work as this after compilation is not what I expected). I pushed the new change.

kuhe (approved):

PR #1471, opened 2024-12-15
pass client config to loadNodeConfig calls

kuhe:
Pass the client config object to loadNodeConfig calls.

This does nothing for non-AWS clients, but allows code-level, i.e. client-scoped AWS profile value to be set.


.changeset/witty-pans-roll.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
trivikr, line 2: nit: empty changeset is not needed is only Java file are edited.

trivikr (approved):

.changeset/witty-pans-roll.md
kuhe: Yeah, I agree it's not necessary, but I sometimes create one to remind myself of what changes are pending since it creates an item in the changeset folder associated with this commit.

PR #1475, opened 2024-12-19
fix for operationContextParam codegen

kuhe:
Internal JS-5629

This PR fixes codegen for operationContextParam by changing the path expression result into a mapping function based on the command's input. The getter functions are made non-throwing so as not to perform unintentional client side validation.

### DIFF
--- a/packages/middleware-endpoint/src/types.ts
+++ b/packages/middleware-endpoint/src/types.ts
@@ -6,7 +6,8 @@ export interface EndpointParameterInstructions {
     | BuiltInParamInstruction
     | ClientContextParamInstruction
     | StaticContextParamInstruction
-    | ContextParamInstruction;
+    | ContextParamInstruction
+    | OperationContextParamInstruction;
 }
 
 /**
@@ -39,4 +40,12 @@ export interface StaticContextParamInstruction {
 export interface ContextParamInstruction {
   type: "contextParams";
   name: string; // The input structure's member name that has contextParams trait
+}
+
+/**
+ * @internal
+ */
+export interface OperationContextParamInstruction {
+  type: "operationContextParams";
+  get(input: any): any;
 }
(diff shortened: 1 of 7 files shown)


packages/middleware-endpoint/src/types.ts
```diff
@@ -40,3 +41,11 @@ export interface ContextParamInstruction {
   type: "contextParams";
   name: string; // The input structure's member name that has contextParams trait
 }
+
+/**
+ * @internal
+ */
+export interface OperationContextParamInstruction {
+  type: "operationContextParams";
+  get(input: any): any;
```
kuhe, line 50: can I assume that the path expression is solely based on the input? No other parameters are needed, right?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -344,7 +344,7 @@ private void generateEndpointParameterInstructionProvider() {
             operationContextParamValues.forEach((name, jmesPathForInputInJs) -> {
                 writer.write(
                     """
-                    $L: { type: \"operationContextParams\", name: $L },
+                    $L: { type: \"operationContextParams\", get: (input?: any) => $L },
```
kuhe, line 347: These need to be functions because there is no `this` at the time the instruction is evaluated. The instruction sits in the global scope due to the use of classBuilder to create commands.

```ts
export class XCommand extends $Command.
  classBuilder<...>()
  .ep({
    ...commonParams,
    MemberName: { type: "operationContextParams", get: (input?: any) => ... },
  })
  ...
  .build() {
  protected declare static __types: { ... }
}
```

packages/middleware-endpoint/src/types.ts
trivikr: Yes. As per internal documentation
> The `operationContextParams` trait defines one or more context parameters that **MUST** be bound to values specified in the operation input using a JMESPath expression

PR #1179, opened 2024-03-04
feat(fetch-http-handler): improve performance, replace FileReader with Blob.arrayBuffer()

richarddd:
*Description of changes:*
Improve performance of collecting blob body by using` arrayBuffer() `method.

If one or more of the packages in the `/packages` directory has been modified, be sure `yarn changeset add` has been run and its output has
been committed and included in this pull request. See CONTRIBUTING.md.

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1483, opened 2025-01-02
fix(fetch-http-handler): add polyfill to collect Blob in react-native

trivikr:
*Issue #, if available:*
* https://github.com/aws-samples/aws-sdk-js-tests/issues/223
* https://github.com/aws/aws-sdk-js-v3/issues/6733
* https://github.com/aws/aws-sdk-js-v3/issues/6636
* https://github.com/aws/aws-sdk-js-v3/issues/6626#issuecomment-2506443781

*Description of changes:*

Adds polyfill to collect Blob in react-native environment where arrayBuffer API is not available.
This re-introduces the code removed in https://github.com/smithy-lang/smithy-typescript/pull/1179, but only when `Blob.prototype.arrayBuffer` implementation is not available.

Tested by copying fetch-http-handler artifacts to https://github.com/aws-samples/aws-sdk-js-tests/issues/223, and verifying that API call is successful.

---

### DIFF
--- a/packages/fetch-http-handler/src/stream-collector.ts
+++ b/packages/fetch-http-handler/src/stream-collector.ts
@@ -1,12 +1,22 @@
 import { StreamCollector } from "@smithy/types";
+import { fromBase64 } from "@smithy/util-base64";
 
 export const streamCollector: StreamCollector = async (stream: Blob | ReadableStream): Promise<Uint8Array> => {
   if ((typeof Blob === "function" && stream instanceof Blob) || stream.constructor?.name === "Blob") {
-    return new Uint8Array(await (stream as Blob).arrayBuffer());
+    if (Blob.prototype.arrayBuffer !== undefined) {
+      return new Uint8Array(await (stream as Blob).arrayBuffer());
+    }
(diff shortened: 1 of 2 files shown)


siddsriv (approved):

packages/fetch-http-handler/src/stream-collector.ts
```diff
@@ -32,3 +42,26 @@ async function collectStream(stream: ReadableStream): Promise<Uint8Array> {
...
+      const commaIndex = result.indexOf(",");
+      const dataOffset = commaIndex > -1 ? commaIndex + 1 : result.length;
+      resolve(result.substring(dataOffset));
+    };
+    reader.onabort = () => reject(new Error("Read aborted"));
+    reader.onerror = () => reject(reader.error);
+    // reader.readAsArrayBuffer is not always available
+    reader.readAsDataURL(blob);
+  });
+}
```
kuhe, line 67: is this the same code that was there before?

kuhe: nm, answered in description

PR #1484, opened 2025-01-02
feat(smithy-client): add isInstance method for ServiceException class

siddsriv:
*Issue #, if available:*
Internal JS-5478

*Description of changes:*
Adds an `isInstance` type check helper method for generic `ServiceException` class in `smithy-client`.

`ServiceException.isInstance(e)`/ `e instanceof ServiceException` should now tell if something is an instance of the generic `ServiceException` class

---

### DIFF
--- a/packages/smithy-client/src/exceptions.ts
+++ b/packages/smithy-client/src/exceptions.ts
@@ -51,7 +51,7 @@ export class ServiceException extends Error implements SmithyException, Metadata
     );
   }
 
-  static [Symbol.hasInstance](instance: unknown) {
+  public static [Symbol.hasInstance](instance: unknown) {
     return ServiceException.isInstance(instance);
   }
 }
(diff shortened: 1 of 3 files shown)


packages/smithy-client/src/exceptions.ts
```diff
@@ -37,6 +37,19 @@ export class ServiceException extends Error implements SmithyException, Metadata
...
+   */
+  public static isInstance(value: unknown): value is ServiceException {
+    if (!value) return false;
+    const candidate = value as ServiceException;
+    return (
+      Boolean(candidate.$fault) &&
+      Boolean(candidate.$metadata) &&
+      (candidate.$fault === "client" || candidate.$fault === "server")
+    );
+  }
```
kuhe, line 52: should we implement `[Symbol.hasInstance]()` here too?

packages/smithy-client/src/exceptions.ts
```diff
@@ -37,6 +37,23 @@ export class ServiceException extends Error implements SmithyException, Metadata
...
+
+  static [Symbol.hasInstance](instance: unknown) {
```
kuhe, line 54: should have public keyword

kuhe (approved):

PR #1490, opened 2025-01-06
fix(smithy-client): instanceof support for ServiceException class

siddsriv:
*Issue #, if available:*
Internal JS-5478, Internal JS-5642

*Description of changes:*
adds proper support for the `instanceof` operator to ServiceException while ensuring subclasses maintain strict prototype checking.

1. adds `[Symbol.hasInstance]` to ServiceException that supports `instanceof` checks according to the assertion table noted below (in the comments).
 

2. fixes prototype chain setup in ServiceException constructor to properly support inheritance.
```typescript
Object.setPrototypeOf(this, Object.getPrototypeOf(this).constructor.prototype);
```
this resolves an issue where duck-typed objects were incorrectly matching subclass `instanceof` checks (e.g. `dummyServiceException instanceof NoSuchKey` would return true even when it wasn't actually a NoSuchKey instance). 

*Testing*
for `smithy-client`:
```console
✓ src/get-value-from-text-node.spec.ts (3)
 ✓ src/exceptions.spec.ts (16)
 ✓ src/lazy-json.spec.ts (7)
 ✓ src/date-utils.spec.ts (155)
 ✓ src/client.spec.ts (7)
 ✓ src/parse-utils.spec.ts (430)
 ✓ src/command.spec.ts (2)
 ✓ src/object-mapping.spec.ts (8)
 ✓ src/is-serializable-header-value.spec.ts (4)
 ✓ src/quote-header.spec.ts (4)
 ✓ src/create-aggregated-client.spec.ts (4)
 ✓ src/split-every.spec.ts (5)
 ✓ src/split-header.spec.ts (2)
 ✓ src/ser-utils.spec.ts (4)
 ✓ src/serde-json.spec.ts (3)
 ↓ src/emitWarningIfUnsupportedVersion.spec.ts (7) [skipped]

 Test Files  15 passed | 1 skipped (16)
      Tests  654 passed | 7 skipped (661)
   Start at  18:24:37
   Duration  1.77s (transform 1.01s, setup 2ms, collect 1.71s, tests 565ms, environment 3ms, prepare 2.46s)
```

---

### DIFF
--- a/packages/smithy-client/src/exceptions.spec.ts
+++ b/packages/smithy-client/src/exceptions.spec.ts
@@ -34,90 +34,120 @@ it("ExceptionOptionType allows specifying message", () => {
   expect(exception.code).toBe("code");
 });
 
-describe("ServiceException type checking", () => {
-  class TestServiceException extends ServiceException {
+describe("Exception Hierarchy Tests", () => {
+  // test classes to represent the hierarchy
+  class ClientServiceException extends ServiceException {
     constructor() {
       super({
-        name: "TestServiceException",
+        name: "ClientServiceException",
         $fault: "client",
-        $metadata: {},
+        $metadata: {}
       });
-      Object.setPrototypeOf(this, TestServiceException.prototype);
+      Object.setPrototypeOf(this, ClientServiceException.prototype);
     }
   }
 
-  const baseError = new ServiceException({
-    name: "Error",
-    $fault: "client",
-    $metadata: {},
-  });
+  class ModeledClientServiceException extends ClientServiceException {
+    constructor() {
+      super();
+      this.name = "ModeledClientServiceException";
+      Object.setPrototypeOf(this, ModeledClientServiceException.prototype);
+    }
+  }
 
-  const subclassError = new TestServiceException();
-
-  const duckTyped = {
-    $fault: "server",
-    $metadata: {},
-  };
-
-  describe("isInstance", () => {
-    it("should return true for ServiceException instances", () => {
-      expect(ServiceException.isInstance(baseError)).toBe(true);
-    });
-
-    it("should return true for duck-typed objects", () => {
-      expect(ServiceException.isInstance(duckTyped)).toBe(true);
-    });
-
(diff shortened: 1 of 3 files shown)


packages/smithy-client/src/exceptions.spec.ts
```diff
@@ -35,20 +35,33 @@ it("ExceptionOptionType allows specifying message", () => {
...
 describe("ServiceException type checking", () => {
-  const error = new ServiceException({
+  class TestServiceException extends ServiceException {
+    constructor() {
+      super({
+        name: "TestServiceException",
+        $fault: "client",
+        $metadata: {},
+      });
+      Object.setPrototypeOf(this, TestServiceException.prototype);
```
siddsriv, line 45: just a small callout: changing prototypes is known to be slow

kuhe:
This is the functionality table I think we should have:

```ts
// hierarchy

// language base error
const E = Error;

// SDK base service exception for any exception originating from a service response
const A = class ServiceException extends Error { ... }

// base exception for an individual service client
const B = class ClientServiceException extends ServiceException {}

// modeled exception shape from a service client or package
const C = class ModeledClientServiceException extends ClientServiceException {}
```

object | instanceof
-- | --
`{}` | -
`new Error` | E
`new ServiceException` | E, A
`new ClientServiceException` | E, A, B
`new ModeledClientServiceException` | E, A, B, C
`{ name: "Error" }` | -
`{ $fault, $metadata }` | A
`{ name: "ServiceException", $fault, $metadata }` | A
`{ name: "ClientServiceException", $fault, $metadata }` | A, B
`{ name: "ModeledClientServiceException", $fault, $metadata }` | A, C

For the last row, the C-level exception like `{ name: "ModeledClientServiceException", $fault, $metadata }`. If it were prototyped it should also identify as `B`, but because it's a duck-type, we can only use the $-props to identify as `A`, and its name to identify as `C`.

kuhe (approved):

kuhe (approved):

PR #1491, opened 2025-01-06
fix(util-waiter): fix waiter validation

smilkuri:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/6767

*Description of changes:*
The waitUntilObjectExists function was incorrectly throwing an error:
Error: WaiterConfiguration.minDelay must be greater than 0
When minDelay was set to 0.5.

Changes Made:
Updated validateWaiterOptions to only throw an error if minDelay, maxDelay, or maxWaitTime are less than or equal to 0 instead of < 1.
Added a test case to verify that small values like 0.01 do not throw an error.
This ensures that fractional values (e.g., 0.5, 0.01) work correctly while keeping 0 invalid. 

*Testing*
for util-waiter:

 ✓ src/waiter.spec.ts (5)
 ✓ src/utils/validate.spec.ts (7)
 ✓ src/poller.spec.ts (5)
 ✓ src/index.spec.ts (1)
 ✓ src/createWaiter.spec.ts (3)

 Test Files  5 passed (5)
      Tests  21 passed (21)
   Start at  12:01:40
   Duration  223ms (transform 105ms, setup 0ms, collect 212ms, tests 15ms, environment 0ms, prepare 227ms)

---

### DIFF
--- a/.changeset/eleven-poems-bake.md
+++ b/.changeset/eleven-poems-bake.md
@@ -1,0 +1,5 @@
+---
+"@smithy/util-waiter": minor
+---
+
+modify the waitUntilObjectExists Validation for minDelay
--- a/.changeset/eleven-poems-bake.md
+++ b/.changeset/eleven-poems-bake.md
@@ -1,5 +1,5 @@
 ---
-"@smithy/util-waiter": minor
+"@smithy/util-waiter": patch
 ---
 
-modify the waitUntilObjectExists Validation for minDelay
+this validation is fixed for all waiters
--- a/.changeset/eleven-poems-bake.md
+++ b/.changeset/eleven-poems-bake.md
@@ -2,4 +2,4 @@
 "@smithy/util-waiter": patch
 ---
 
-this validation is fixed for all waiters
+fix range validation in waiters
--- a/packages/util-waiter/src/utils/validate.spec.ts
+++ b/packages/util-waiter/src/utils/validate.spec.ts
@@ -54,11 +54,7 @@ describe(validateWaiterOptions.name, () => {
     waiterOptions.maxWaitTime = 0.5;
     waiterOptions.minDelay = 0.0001;
     waiterOptions.maxDelay = 0.4;
-    try {
-      validateWaiterOptions(waiterOptions);
-    } catch (e) {
-      expect(e).toBe("SHOULD NOT ERROR HERE");
-    }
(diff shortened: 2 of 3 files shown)


.changeset/eleven-poems-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/util-waiter": minor
+---
+
+modify the waitUntilObjectExists Validation for minDelay
```
kuhe, line 5: The changeset message should say this validation is fixed for all waiters.

Also, don't capitalize the word Validation.

.changeset/eleven-poems-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/util-waiter": minor
```
kuhe, line 2: This is a fix and there is no API addition, so the changeset type should be "patch".

packages/util-waiter/src/utils/validate.spec.ts
```diff
@@ -50,6 +50,17 @@ describe(validateWaiterOptions.name, () => {
...
 
+  it("should not throw an error with small decimal numbers", () => {
+    waiterOptions.maxWaitTime = 0.5;
+    waiterOptions.minDelay = 0.0001;
+    waiterOptions.maxDelay = 0.4;
+    try {
+      validateWaiterOptions(waiterOptions);
+    } catch (e) {
+      expect(e).toBe("SHOULD NOT ERROR HERE");
+    }
```
kuhe, line 61: This catch block is unnecessary. 

The try-branch, if not throwing, will not assert anything. The catch block assertion will never pass. 

That means having no try-catch block at all would produce the same result. If you just want to assert that some code does not throw, simply run it with no assertions.

.changeset/eleven-poems-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/util-waiter": patch
+---
+
+this validation is fixed for all waiters
```
kuhe, line 5: This message appears in the changelog, so "this validation" is going to be ambiguous. 

The message can be "fix range validation in waiters".

kuhe:
run `yarn format` at the repository root to format the files.

kuhe (approved):

PR #1500, opened 2025-01-09
fix(pagination): only use startingToken if it exists

icholy:
### Description of changes:

The existing value at `inputTokenName` should not be overwritten if `startingToken` was not provided.

### Use-Case:

I was surprised when the following DynamoDB scan didn't respect the `ExclusiveStartKey` property, but did respect the `Limit` property.

```
const paginator = paginateScan(
	{ client: documentClient },
	{
		TableName: 'my-table',
		ExclusiveStartKey: last_key,
		Limit: 1000,
	},
);
```


kuhe (approved):

kuhe:
released in https://www.npmjs.com/package/@smithy/core/v/3.1.0

PR #1505, opened 2025-01-15
fix(node-http-handler): Write request body if 100 Continue response takes more than 1 second

Nevon:
Fixes the issue reported here: https://github.com/aws/aws-sdk-js-v3/issues/6805

*Description of changes:*

When the client makes a request with the `Expect: 100-continue` header, it will wait for 1000ms to get either a `100 Continue` response or an error. If it does not receive any response within the 1000ms, the intention seems to be that the client should write the body anyway. This behavior broke in #1459 because in the timeout case `resolve` is called with no arguments, which means that `sendBody` becomes undefined and the body is never sent. Even if the server returns a 100 Continue after 1001ms, the client is no longer waiting for that response and instead just idles until eventually the server returns an error.

This change restores the previous behavior, where the client will send the body after waiting 1000ms.

---

### DIFF
--- a/.changeset/good-buttons-relax.md
+++ b/.changeset/good-buttons-relax.md
@@ -1,0 +1,5 @@
+---
+"@smithy/node-http-handler": patch
+---
+
+fix sending request when 100 Continue response takes more than 1 second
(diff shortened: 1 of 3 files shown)


kuhe (approved):

kuhe:
> the client will send the body after waiting 1000ms.

I changed this to 6000ms, explained inline comments.

PR #1516, opened 2025-01-29
chore(docs): move description block before deprecated tag

aBurmeseDev:
*Issue #, if available:*
- https://github.com/aws/aws-sdk-js-v3/issues/6672
- P187212400

*Description of changes:*
- Currently the description for parameter is missing when param has deprecated tag. By moving deprecated tag after block, description should render correctly.

Resource: https://tsdoc.org/pages/tags/deprecated/

---


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -266,7 +266,7 @@ boolean writeMemberDocs(Model model, MemberShape member) {
                     docs = docs.replace("{", "\\{")
                         .replace("}", "\\}");
                     if (member.getTrait(DeprecatedTrait.class).isPresent() || isTargetDeprecated(model, member)) {
-                        docs = "@deprecated\n\n" + docs;
+                        docs = docs + "\n\n@deprecated";
```
kuhe, line 269: this is needed since the tsdoc does not support explicit `@description` block like jsdoc. tsdoc uses only the implied description block and it must come first in the docblock order.

kuhe (approved):

PR #1517, opened 2025-01-30
feat(docs): add client config option reference

aBurmeseDev:
*Issue #, if available:*
- https://github.com/aws/aws-sdk-js-v3/issues/6858

*Description of changes:*
- Add a comment with client config options to developer guide link in code example syntax for reference
- `// See AWS SDK config options: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/configuring-the-jssdk.html`

---

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
@@ -254,7 +254,7 @@ final class CommandGenerator implements Runnable {
                         packageName)
                 + String.format("// const { %s, %s } = require(\"%s\"); // CommonJS import%n", serviceName, commandName,
                         packageName)
-                + String.format("const client = new %s(config);%n", serviceName)
+                + String.format("const client = new %s(config); // See AWS SDK config options: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/configuring-the-jssdk.html%n", serviceName)
                 + String.format("const input = %s%n",
                         StructureExampleGenerator.generateStructuralHintDocumentation(
                                 model.getShape(operation.getInputShape()).get(), model, false, true))


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -254,7 +254,7 @@ private String getCommandExample(String serviceName, String configName, String c
                         packageName)
                 + String.format("// const { %s, %s } = require(\"%s\"); // CommonJS import%n", serviceName, commandName,
                         packageName)
-                + String.format("const client = new %s(config);%n", serviceName)
+                + String.format("const client = new %s(config); // See AWS SDK config options: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/configuring-the-jssdk.html%n", serviceName)
```
kuhe, line 257: resulting line might be too long, move to above this line

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
```diff
@@ -254,6 +254,7 @@ private String getCommandExample(String serviceName, String configName, String c
                         packageName)
                 + String.format("// const { %s, %s } = require(\"%s\"); // CommonJS import%n", serviceName, commandName,
                         packageName)
+                + "// See AWS SDK config options: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/configuring-the-jssdk.html\n"
                 + String.format("const client = new %s(config);%n", serviceName)
```
kuhe, line 258: this is smithy, we shouldn't link to AWS docs.

A preface block to the line containing the client init should be injectable by the downstream code generator. You can refer to the `CodeSection` interface usage in this repo on how to do that.

kuhe (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CommandGenerator.java
kuhe: we could replace `config` with `{}` so at least there would be no question about an undefined variable.

kuhe: instead of a link, there could be a comment saying that the exported type of the config is the client class name plus the word "Config".

e.g.

```ts
import { S3Client, S3ClientConfig } from "@aws-sdk/client-s3";
```

PR #1523, opened 2025-02-13
feat(util-stream): buffering of stream chunks

kuhe:
add `createBufferedReadable` fn to util-stream to wrap Readable and ReadableStream, buffering chunks to a configurable minimum byte length.


kuhe:
this PR will be used to address https://github.com/aws/aws-sdk-js-v3/issues/6859

packages/util-stream/src/ByteArrayCollector.ts
```diff
@@ -0,0 +1,28 @@
...
+  private byteArrays = [] as Uint8Array[];
+
+  public constructor(public readonly ctor: (size: number) => Uint8Array) {}
+
+  public push(byteArray: Uint8Array) {
+    this.byteArrays.push(byteArray);
+    this.byteLength += byteArray.byteLength;
+  }
+
+  public flush() {
```
siddsriv, line 16: will flush() always create a new aggregate buffer, even if there's only one chunk?

kuhe: updated to return single bytearray without copy

Issue #1530, opened 2025-02-24
node-http-handler: socket timeout handlers leaking for keep alive connections


tomekt-bolt:
If `NodeHttpHandler` is used with keepAlive agent, _every_ request going through a socket registers its own `timeout` handler on this socket and does not clear it.

This leads to the amount of callbacks growing unnecessarily throughout the socket's life time and `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 timeout listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit` warning being emitted.

Each request should remove its related `timeout` listener from the underlying socket upon request closure.

**Repro**:
Run the program below.

**Expected result**:
Now warning is emitted.

**Actual result**:
`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 timeout listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit` is emited

```
import {HttpRequest} from "@smithy/protocol-http";
import * as https from "https";

import {NodeHttpHandler} from "./node-http-handler";

/*
 * REQUEST_DELAY_MS has to be greater than DEFER_EVENT_LISTENER_TIME in `set-socket-timeout.ts`, so that the socket
 * timeout listeners are _actually_ registered.
 */
const REQUEST_DELAY_MS = 4000;

const handler = new NodeHttpHandler({
    httpsAgent: new https.Agent({
        /*
         * Using just one socket, to easily simulate many requests handled with the same keep-alive connection.
         */
        maxSockets: 1,
        keepAlive: true,
    }),
    requestTimeout: 30_000,
});

async function main() {
    for (let i = 0; i < 11; i++) {
        const response = await handler.handle(new HttpRequest({
            method: "GET",
            hostname: "httpstat.us",
            protocol: "https:",
            path: `/200?sleep=${REQUEST_DELAY_MS}`
        }));

        response.response.body.on("data", (chunk) => {
            console.log(chunk.toString())
        });
        console.log(`Response ${i}: ${response.response.statusCode}`);
    }
}

main().catch(console.error);
```

kuhe:
thanks for your contribution, the fix was published in 
https://www.npmjs.com/package/@smithy/node-http-handler/v/4.0.3

tomekt-bolt:
Thanks for your support!

PR #1531, opened 2025-02-24
fix(node-http-handler): clear obsolete socket timeout listeners

tomekt-bolt:
*Issue #1530*

*Description of changes:*

This PR adds cleaning the socket timeout listener from the underlying socket upon requests `close`, to avoid leaking the listeners throughout a keepAlive socket's lifetime.

---

### DIFF
--- a/packages/node-http-handler/src/set-socket-timeout.ts
+++ b/packages/node-http-handler/src/set-socket-timeout.ts
@@ -19,6 +19,7 @@ export const setSocketTimeout = (
 
     if (request.socket) {
       request.socket.setTimeout(timeout, onTimeout);
+      request.on("close", () => request.socket?.removeListener("timeout", onTimeout));
     } else {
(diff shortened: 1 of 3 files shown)


kuhe (approved):

packages/node-http-handler/src/set-socket-timeout.ts
```diff
@@ -19,6 +19,7 @@ export const setSocketTimeout = (
 
     if (request.socket) {
       request.socket.setTimeout(timeout, onTimeout);
+      request.on("close", () => request.socket?.removeListener("timeout", onTimeout));
```
kuhe, line 22: `onTimeout` is a unique reference per request

kuhe:
published in https://www.npmjs.com/package/@smithy/node-http-handler/v/4.0.3

PR #1537, opened 2025-02-27
Support JMESPath flatten in OperationContextParams

trivikr:
*Issue #, if available:*
* Internal JS-5749
* JMESPath docs: https://jmespath.org/specification.html#flatten-operator
* Array.prototype.flat() https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

*Description of changes:*
Support JMESPath flatten in OperationContextParams

---

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
@@ -366,10 +366,17 @@ public class RuleSetParameterFinder {
         }
 
         // Process Flatten operator https://jmespath.org/specification.html#flatten-operator
-        if (part.endsWith("[*][]")) {
+        if (part.endsWith("[]")) {
             // Get key to run hash wildcard on.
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
```diff
@@ -359,6 +365,21 @@ private String getJmesPathExpressionSection(String separator, String value, Stri
...
+        if (part.endsWith("[]")) {
+            // Get key to run hash wildcard on.
+            String key = part.substring(0, part.length() - 2);
+
+            // If key is on list item
+            if (key.endsWith("[*]")) {
+                value = value + separator + key.substring(0, key.length() - 3) + ".flat()";
+            } else {
+                // key is on object
+                value = value + separator + key + ").flat()";
```
siddsriv, line 378: will there always be an opening parenthesis? since we are doing a concatenation with `").flat()"`

kuhe (approved):

siddsriv (approved):

PR #1571, opened 2025-04-24
feat(core): resolve auth schemes based on the preference list

trivikr:
*Issue #, if available:*
Internal JS-5811

*Description of changes:*
Resolves auth schemes based on the preference list

Testing is done by installing `@aws-sdk/client-s3` and `@smithy/core`, and validating signature
```console
$ yarn add ../aws-sdk-js-v3/clients/client-s3/aws-sdk-client-s3-3.796.0.tgz
$ yarn add ../smithy-typescript/packages/core/smithy-core-3.2.0.tgz
$ cp -R node_modules/@smithy/core/* node_modules/@aws-sdk/client-s3/node_modules/@smithy/core
```

 Test code
```js
import "@aws-sdk/signature-v4-crt";

const interceptorMiddleware = (next, context) => (args) => {
  // middleware intercept the request and return it early
  const request = args.request;
  return Promise.resolve({
    output: {
      $metadata: { attempts: 0, httpStatusCode: 200 },
      request,
      context,
    },
    response: {},
  });
};

const OutpostId = "op-01234567890123456";
const AccountId = "123456789012";
const region = "us-west-2";
const clientRegion = "us-east-1";
const credentials = { accessKeyId: "key", secretAccessKey: "secret" };

const client = new S3({
  region: clientRegion,
  credentials,
  useArnRegion: true,
  authSchemePreference: ["sigv4"],
});

client.middlewareStack.add(interceptorMiddleware, {
  step: "finalizeRequest",
  priority: "low",
});

const result = await client.putObject({
  Bucket: `arn:aws:s3-outposts:${region}:${AccountId}:outpost/${OutpostId}/accesspoint/abc-111`,
  Key: "key",
  Body: "body",
});

const headers = result.request.headers;
const authorizationHeader =
  headers["authorization"] || headers["Authorization"];

// Prints "AWS4-ECDSA-P256-SHA256" when authSchemePreference starts with "sigv4a", or is not set.
// Prints "AWS4-HMAC-SHA256" when authSchemePreference is set to "sigv4".
console.log(authorizationHeader.split(" ")[0]);
```

---


packages/core/src/middleware-http-auth-scheme/httpAuthSchemeMiddleware.ts
```diff
@@ -84,18 +88,20 @@ export const httpAuthSchemeMiddleware =
     const options = config.httpAuthSchemeProvider(
       await mwOptions.httpAuthSchemeParametersProvider(config, context as TContext, args.input)
     );
-    const authSchemes = convertHttpAuthSchemesToMap(config.httpAuthSchemes);
+    const optionsMap = convertHttpAuthOptionsToMap(options);
+
+    const authSchemePreference = await config.authSchemePreference();
```
trivikr, line 93: Add check for `config.authSchemePreference` for backward compatibility, since `@smithy` packages use `^`
```suggestion
    const authSchemePreference = config.authSchemePreference ? await config.authSchemePreference() : [];
```

trivikr:
The PR is blocked on sigv4 being selected when autheSchemePreference is not passed.
This happens as `config.httpAuthSchemes` doesn't maintain order `["signv4a", "sigv4"]` as defined in the models.

packages/core/src/middleware-http-auth-scheme/resolveAuthOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // add any remaining candidates that weren't in the preference list
+  for (const candidateAuthOption of candidateAuthOptions) {
+    if (!preferredAuthOptions.find(({ schemeId }) => schemeId === candidateAuthOption.schemeId)) {
+      preferredAuthOptions.push(candidateAuthOption);
+    }
+  }
+
+  return preferredAuthOptions;
```
kuhe, line 38: because the lists are so small, I feel that this is optional, but you can do this without nested for loops like this:

```ts
// runnable in console
{
const candidates = [{ id: 'a', }, { id: 'b' }];
const preference = ['c', 'b'];

const candidateMap = candidates.reduce((map, item) => {
  map.set(item.id, item);
  return map;
}, new Map);

const preferredMap = new Map;

for (const pref of preference) {
  if (candidateMap.has(pref)) {
    preferredMap.set(pref, candidateMap.get(pref));
  }
}

for (const [id, item] of candidateMap) {
  preferredMap.set(id, item);
}

console.log([...preferredMap.values()])
}
```

kuhe (approved):

packages/core/src/middleware-http-auth-scheme/resolveAuthOptions.ts
trivikr: I think we should retain the resolve function, as it's recommended in the internal cross SDK proposal.
Also, the JS SDK implementation uses map for authSchemes which do not have order preserved. The order is preserved in authOptions instead.

kuhe: I'm not sure if you're referring to the Map in the example, but those retain insertion order like modern objects. 

But let's keep the function as is since it is in the design.

PR #1574, opened 2025-05-01
fix(service-error-classification): add TypeError as transient error

siddsriv:
*Issue #, if available:*
Internal JS-5890

*Description of changes:*
fetch API surfaces connectivity errors as TypeErrors, example: `TypeError: Failed to fetch` in Chrome. 

For browsers, we should retry for connectivity issues (offline), this is a fix to retry on TypeErrors then. Since we don't want to retry for `TypeErrors` that *aren't* network errors, I've paired this check with the actual error messages that different browsers surface (not an exhaustive list, but covers some major ones). 

Fetch has an open issue about this: https://github.com/whatwg/fetch/issues/526

additional resources
- https://stackoverflow.com/questions/61113344/fetch-api-how-to-determine-if-an-error-is-a-network-error

---

### DIFF
--- a/packages/service-error-classification/src/index.ts
+++ b/packages/service-error-classification/src/index.ts
@@ -20,6 +20,28 @@ export const isClockSkewError = (error: SdkError) => CLOCK_SKEW_ERROR_CODES.incl
  */
 export const isClockSkewCorrectedError = (error: SdkError) => error.$metadata?.clockSkewCorrected;
 
+/**
+ *
+ * @internal
+ */
+export const isBrowserNetworkError = (error: SdkError) => {
+  const errorMessages = new Set([
+    "Failed to fetch", // Chrome
+    "NetworkError when attempting to fetch resource", // Firefox
+    "The Internet connection appears to be offline", // Safari 16
+    "Load failed", // Safari 17+
+    "Network request failed", // `cross-fetch`
+  ]);
+
+  const isValid = error && error instanceof TypeError;
+
+  if (!isValid) {
+    return false;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/service-error-classification/src/index.ts
```diff
@@ -20,6 +20,28 @@ export const isClockSkewError = (error: SdkError) => CLOCK_SKEW_ERROR_CODES.incl
...
+    'Network request failed', // `cross-fetch`
+  ]);
+  
+  const isValid = error && error instanceof TypeError;
+
+  if(!isValid) {
+    return false;
+  }
+
+  return errorMessages.has(error.message);
```
siddsriv, line 42: doing a check for messages as well to avoid unintended `TypeError`s from being retried

packages/service-error-classification/src/index.ts
```diff
@@ -20,6 +20,28 @@ export const isClockSkewError = (error: SdkError) => CLOCK_SKEW_ERROR_CODES.incl
  */
 export const isClockSkewCorrectedError = (error: SdkError) => error.$metadata?.clockSkewCorrected;
 
+/**
+ * 
+ * @internal 
+ */
+export const isBrowserNetworkError = (error: SdkError) => {
+  const errorMessages = new Set([
```
siddsriv, line 28: this constant could be placed somewhere else, like the constants file in the same dir. I'd like this here since:
1. these are not as strong a check as error codes or status codes, so I want to keep them away from other constants
2. this function will be easily(-ier?) visible in the future if someone is taking a look at transient errors.

kuhe (approved):

PR #1584, opened 2025-05-07
fix(node-config-provider): pass logger to environment variable selector

trivikr:
*Issue #, if available:*
https://github.com/smithy-lang/smithy-typescript/pull/1583/files#r2076091590

*Description of changes:*
Passes logger to environment variable selector

---

### DIFF
--- a/packages/node-config-provider/src/configLoader.ts
+++ b/packages/node-config-provider/src/configLoader.ts
@@ -37,7 +37,15 @@ export const loadConfig = <T = string>(
   { environmentVariableSelector, configFileSelector, default: defaultValue }: LoadedConfigSelectors<T>,
   configuration: LocalConfigOptions = {}
 ): Provider<T> => {
-  const envOptions = configuration.signingName ? { signingName: configuration.signingName } : undefined;
+  const envOptions: EnvOptions = {};
+
+  if (configuration.signingName !== undefined) {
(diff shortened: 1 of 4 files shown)


siddsriv (approved):

packages/node-config-provider/src/configLoader.ts
```diff
@@ -37,7 +37,15 @@ export const loadConfig = <T = string>(
...
 ): Provider<T> => {
-  const envOptions = configuration.signingName ? { signingName: configuration.signingName } : undefined;
+  const envOptions: EnvOptions = {};
+
+  if (configuration.signingName !== undefined) {
+    envOptions.signingName = configuration.signingName;
+  }
+  if (configuration.logger !== undefined) {
+    envOptions.logger = configuration.logger;
+  }
```
kuhe, line 47: since they're already optional, assign without check?

```ts
const { signingName, logger } = configuration;
const envOptions: EnvOptions = { signingName, logger };
```

kuhe (approved):

packages/node-config-provider/src/configLoader.ts
trivikr: Done in [`05c748f` (#1584)](https://github.com/smithy-lang/smithy-typescript/pull/1584/commits/05c748f5f5cc77782afcf70c596b9d1df99a1e9b)

PR #1586, opened 2025-05-08
chore(node-config-provider): export Getters and their configs

trivikr:
*Issue #, if available:*
Internal JS-5860

*Description of changes:*
The EnvOptions created in https://github.com/smithy-lang/smithy-typescript/pull/1583 need to be exported for consumption by providers. Similar other configurations and getters also need to be exported is other packages want to explicitly use them.

---


packages/node-config-provider/src/index.ts
```diff
@@ -1 +1,4 @@
 export * from "./configLoader";
+export { EnvOptions, GetterFromEnv } from "./fromEnv";
+export { GetterFromConfig, SharedConfigInit } from "./fromSharedConfigFiles";
+export { FromStaticConfig } from "./fromStatic";
```
kuhe, line 4: There should be a form of "internal package, do not use" in the top level readme.

kuhe (approved):

PR #1588, opened 2025-05-12
fix(middleware-serde): add response metadata to deserialization errors

kuhe:
This adds the `error.$metadata` field to deserialization errors, so that the (earlier in order) retry middleware can decide to retry the error if the response code is retryable, regardless of the format of the error.

This scenario can happen when a server origin response is supposed to be e.g. JSON but its load balancer is XML or does not reach the point of checking the correct response format. If the LB sheds a request with a retryable error code but unparsable body, the parsing error should be treated as non-terminal, retryable.

### DIFF
--- a/packages/middleware-serde/src/deserializerMiddleware.ts
+++ b/packages/middleware-serde/src/deserializerMiddleware.ts
@@ -66,7 +66,7 @@ export const deserializerMiddleware =
         try {
           // if the deserializer failed, then $metadata may still be set
           // by taking information from the response.
-          if (HttpResponse.isInstance(response) && typeof response.statusCode === "number") {
+          if (HttpResponse.isInstance(response)) {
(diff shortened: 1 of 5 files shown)


packages/middleware-serde/src/deserializerMiddleware.ts
```diff
@@ -60,8 +62,35 @@ export const deserializerMiddleware =
             error.$response.body = error.$responseBodyText;
           }
         }
+
+        try {
+          // if the deserializer failed, then $metadata may still be set
+          // by taking information from the response.
+          if (HttpResponse.isInstance(response) && typeof response.statusCode === "number") {
```
kuhe, line 69:
```suggestion
          if (HttpResponse.isInstance(response)) {
```

trivikr (approved):

PR #1589, opened 2025-05-12
chore: generate default endpoint ruleset

kuhe:
Generates a default endpoint ruleset to make code generation consistent between service models that do and don't provide a ruleset trait.


private/smithy-rpcv2-cbor/src/RpcV2ProtocolClient.ts
```diff
@@ -249,15 +257,16 @@ export class RpcV2ProtocolClient extends __Client<
     let _config_0 = __getRuntimeConfig(configuration || {});
     super(_config_0 as any);
     this.initConfig = _config_0;
-    let _config_1 = resolveCustomEndpointsConfig(_config_0);
```
kuhe, line 252: The main point of the change is here. The deprecated `resolveCustomEndpointsConfig` input interface requires a user-supplied endpoint. Equivalent functionality will be provided by the default endpointRuleSet.

trivikr (approved):

PR #1590, opened 2025-05-13
chore: deprecation markers for pre-endpointsRuleSet components

kuhe:
adds deprecation markers for some components not used by endpoints 2.0


.changeset/long-experts-confess.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
kuhe, line 2: no runtime code changes and this does not warrant a version bump

trivikr (approved):

PR #1595, opened 2025-05-14
feat(core/schema): add schema classes

kuhe:
Adds the `Schema` class group, and schema serde plugin, all of which are still unused in the core/schema submodule.

----

original working branch: https://github.com/smithy-lang/smithy-typescript/pull/1592

group:
https://github.com/smithy-lang/smithy-typescript/pull/1593
https://github.com/smithy-lang/smithy-typescript/pull/1594
https://github.com/smithy-lang/smithy-typescript/pull/1595


packages/core/src/submodules/schema/TypeRegistry.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import { ErrorSchema } from "./schemas/ErrorSchema";
+
+/**
+ * A way to look up schema by their ShapeId values.
+ *
+ * @alpha
+ */
+export class TypeRegistry {
+  public static active: TypeRegistry | null = null;
```
siddsriv, line 11: what is this property for? i don't think it's used anywhere in the current revision

siddsriv (approved):

packages/core/src/submodules/schema/TypeRegistry.ts
kuhe, line 11:
```suggestion
```

PR #1596, opened 2025-05-15
chore(smithy-client): build-fix, hide submodule export from api-extractor

kuhe:
This change hides `core/serde` from api-extractor during the doc build process because it does not understand how to import it. It is ok to hide this export because it contains internal interfaces used in serde-functions.

This does not change the runtime behavior of any package or the actual type interface during regular build and publish steps.

Tested with:
```
npx turbo run extract:docs -F=@smithy/smithy-client
```


trivikr (approved):

kuhe:
https://github.com/microsoft/rushstack/issues/2142

PR #1529, opened 2025-02-21
fix(node-http-handler): fix constructor socketAcquisitionWarningTimeout does not work

yuji-hatakeyama:
*Issue:* https://github.com/smithy-lang/smithy-typescript/issues/1528

*Description of changes:*

[NodeHttpHandler](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-node-http-handler/Interface/NodeHttpHandlerOptions/) “socketAcquisitionWarningTimeout” option does not work correctly.

Fix it

PR #1598, opened 2025-05-18
test(node-http-handler): add test for socketAcquisitionWarningTimeout

yuji-hatakeyama:
*Issue None*

*Description of changes:*

Add test case for the fix in PR #1529 to verify socketAcquisitionWarningTimeout parameter is properly passed from constructor options to config.

Relates to: #1529

[Question]
I only need to modify test, but do I need to bump the version?


yuji-hatakeyama:
@kuhe 
Thanks for the reply.
I have corrected the issue you mentioned at https://github.com/smithy-lang/smithy-typescript/pull/1598#issuecomment-2892023034

1ba2fd17823e037797b67ce5c31484eb35126556

kuhe (approved):

PR #1602, opened 2025-05-22
test: run tests with vitest, fix test type errors

kuhe:
- fixes retry script for tests
- fixes type errors in tests due to upgrading to ts5.8 and its new eslint standards (in a quick way, mostly with `any`)
- updates integration test model with new ruleset


package.json
```diff
@@ -7,8 +7,8 @@
   "scripts": {
     "clean": "turbo run clean --force --parallel",
     "build": "turbo run build",
-    "test": "turbo run test",
-    "test:integration": "yarn build-test-packages && turbo run test:integration",
```
kuhe, line 11: as with jsv3, turbo is a slow way to run tests because it spins up separate vite processes. 

A singular vitest process with a view of all tests is much faster for test compilation.

siddsriv (approved):

PR #1621, opened 2025-06-16
feat: add support for accountId in imds

smilkuri:
*Issue #, if available:*
Internal JS-5966

*Description of changes:*

This PR adds support for version 2.1 of the IMDS credentials provider. With this update, the provider can now retrieve credentials that include an account ID. 

To support this functionality, the credentials provider first attempts to access the extended API endpoint, which ends with `-extended`. If this endpoint returns a 404, the provider falls back to the legacy API endpoint—the same one used in the IMDS credentials provider v2.0.

This "try extended API, then fall back to legacy API" pattern is applied to both retrieving the IMDS instance profile name and fetching credentials. 
Only the extended API can return credentials that include an account ID.

The PR also made the following IMDS credentials providers options configurable:

- `Disable IMDS credentials fetching`
- `IMDS instance profile name`

Both options can be configured via environment variables or a shared config file, with environment variables taking precedence over the config file.

ToDo:

- Add support for feature Id tracking 

---


packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -1,15 +1,18 @@
+import { loadConfig } from "@smithy/node-config-provider";
 import { CredentialsProviderError } from "@smithy/property-provider";
 import { afterEach, beforeEach, describe, expect, test as it, vi } from "vitest";
 
 import { InstanceMetadataV1FallbackError } from "./error/InstanceMetadataV1FallbackError";
-import { fromInstanceMetadata } from "./fromInstanceMetadata";
+import { checkIfImdsDisabled, fromInstanceMetadata } from "./fromInstanceMetadata";
+import * as fromInstanceMetadataModule from "./fromInstanceMetadata";
```
kuhe: use explicit imports instead of `*`

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -49,18 +52,22 @@ describe("fromInstanceMetadata", () => {
...
     secretAccessKey: mockImdsCreds.SecretAccessKey,
     sessionToken: mockImdsCreds.Token,
     expiration: new Date(mockImdsCreds.Expiration),
+    accountId: mockImdsCreds.AccountId,
   });
 
   beforeEach(() => {
     vi.mocked(staticStabilityProvider).mockImplementation((input) => input);
     vi.mocked(getInstanceMetadataEndpoint).mockResolvedValue({ hostname } as any);
+    vi.mocked(loadConfig).mockReturnValue(() => Promise.resolve(false));
```
kuhe, line 72: this is hard to understand because the mock is too broad. What config property is being targeted by this mock return value?

Can the mock be made more specific so we can tell what config property it is trying to intercept?

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -72,6 +79,65 @@ describe("fromInstanceMetadata", () => {
     vi.resetAllMocks();
   });
 
+  it("returns no credentials when AWS_EC2_METADATA_DISABLED=true", async () => {
+    vi.mocked(loadConfig).mockReturnValueOnce(() => Promise.resolve(true));
```
kuhe: same as above: mock is too broad

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -72,6 +79,65 @@ describe("fromInstanceMetadata", () => {
...
+    const profileName = "my-profile-0002";
+
+    vi.mocked(httpRequest)
+      .mockResolvedValueOnce(mockToken as any)
+      .mockResolvedValueOnce(JSON.stringify(mockImdsCreds) as any);
+
+    vi.mocked(retry).mockImplementation((fn: any) => fn());
+    vi.mocked(fromImdsCredentials).mockReturnValue(mockCreds);
+
+    const result = await fromInstanceMetadata({ ec2InstanceProfileName: profileName })();
```
kuhe: variable name too generic

is this "credentials"?

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -72,6 +79,65 @@ describe("fromInstanceMetadata", () => {
...
+      .mockResolvedValueOnce(mockToken as any)
+      .mockResolvedValueOnce(mockProfile as any)
+      .mockResolvedValueOnce(JSON.stringify(mockImdsCreds) as any);
+
+    vi.mocked(retry).mockImplementation((fn: any) => fn());
+    vi.mocked(fromImdsCredentials).mockReturnValue(mockCreds);
+
+    const provider = fromInstanceMetadata({});
+
+    const result = await provider();
```
kuhe: result should be called credentials

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -213,6 +282,73 @@ describe("fromInstanceMetadata", () => {
...
+      vi.mocked(httpRequest).mockClear();
+      vi.mocked(loadConfig).mockClear();
+      vi.mocked(retry).mockImplementation((fn: any) => fn());
+    });
+
+    it("uses ec2InstanceProfileName from init if provided", async () => {
+      const profileName = "profile-from-init";
+      const options = { hostname } as any;
+
+      // Only use vi.spyOn for imported functions
```
kuhe: what does this comment mean?

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -213,6 +282,73 @@ describe("fromInstanceMetadata", () => {
...
+  describe("getImdsProfileHelper", () => {
+    beforeEach(() => {
+      vi.mocked(httpRequest).mockClear();
+      vi.mocked(loadConfig).mockClear();
+      vi.mocked(retry).mockImplementation((fn: any) => fn());
+    });
+
+    it("uses ec2InstanceProfileName from init if provided", async () => {
+      const profileName = "profile-from-init";
+      const options = { hostname } as any;
```
kuhe: try to avoid using `any`

packages/credential-provider-imds/src/fromInstanceMetadata.spec.ts
```diff
@@ -213,6 +282,73 @@ describe("fromInstanceMetadata", () => {
...
+      expect(result).toBe(configProfileName);
+      expect(httpRequest).not.toHaveBeenCalled();
+
+      // 2. Simulate config file missing: should call IMDS (extended fails, legacy succeeds)
+      vi.mocked(loadConfig).mockReturnValue(() => Promise.resolve(null));
+      vi.mocked(httpRequest)
+        .mockRejectedValueOnce(Object.assign(new Error(), { statusCode: 404 }))
+        .mockResolvedValueOnce(legacyProfileName as any);
+
+      result = await fromInstanceMetadataModule.getImdsProfileHelper(options, mockMaxRetries, {});
```
kuhe: rename all variables called "result" to be more specific

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -144,6 +141,60 @@ const getInstanceMetadataProvider = (init: RemoteProviderInit = {}) => {
   };
 };
 
+/**
+ * @internal
+ * Gets IMDS profile with proper error handling and retries
+ */
+export const getImdsProfileHelper = async (
```
kuhe: `getImdsProfile`

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -144,6 +141,60 @@ const getInstanceMetadataProvider = (init: RemoteProviderInit = {}) => {
...
+          apiVersion = "legacy";
+          const response = await httpRequest({ ...options, path: IMDS_LEGACY_PATH });
+          resolvedProfile = response.toString().trim();
+          return resolvedProfile;
+        } else {
+          throw error;
+        }
+      }
+    } catch (err) {
+      throw err;
```
kuhe: there's no point in catching an error just to throw it, remove the outer try-catch

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +205,108 @@ const getMetadataToken = async (options: RequestOptions) =>
     },
   });
 
-const getProfile = async (options: RequestOptions) => (await httpRequest({ ...options, path: IMDS_PATH })).toString();
+/**
+ * @internal
+ * Checks if IMDS credential fetching is disabled through configuration
+ */
+export const checkIfImdsDisabled = async (profile?: string, logger?: any): Promise<void> => {
```
kuhe: function name should be `throwIfImdsTurnedOff`

"check" doesn't imply it will do anything

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +205,108 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+      environmentVariableSelector: (env) => env[ENV_PROFILE_NAME],
+      // Check config file
+      configFileSelector: (profile) => profile[CONFIG_PROFILE_NAME],
+      default: null,
+    },
+    { profile }
+  )();
+
+  // Check runtime config (highest priority)
+  const name = init.ec2InstanceProfileName || profileName;
```
kuhe:
```suggestion
  const name = init.ec2InstanceProfileName ?? profileName;
```

kuhe:
- [ ] create e2e test suite runnable in an EC2 instance. Extra testing is important for credentials-related code

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,107 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+ * Checks if IMDS credential fetching is disabled through configuration
+ */
+export const throwIfImdsTurnedOff = async (profile?: string, logger?: any): Promise<void> => {
+  // Load configuration in priority order
+  const disableImds = await loadConfig(
+    {
+      // Check environment variable
+      environmentVariableSelector: (env) => {
+        const envValue = env[ENV_IMDS_DISABLED];
+        return envValue === "true";
```
kuhe, line 218: this needs to return undefined to move onto the next selector. Returning `envValue === "true"` will always be a boolean.

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,107 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+      },
+      // Check config file
+      configFileSelector: (profile) => {
+        const profileValue = profile[CONFIG_IMDS_DISABLED];
+        return profileValue === "true";
```
kuhe, line 226: needs to return undefined in some cases to move to next selector

packages/credential-provider-imds/src/fromInstanceMetadata.e2e.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  });
+
+  afterEach(() => {
+    process.env = { ...originalEnv };
+  });
+
+  it("should fetch metadata token successfully", async () => {
+    if (!imdsAvailable) {
+      return;
```
kuhe: skip the test if not available, don't return. It will look like the test succeeded.

kuhe:
```ts
const test = imdsAvailable ? it : it.skip;

test("should...
```

packages/credential-provider-imds/src/fromInstanceMetadata.e2e.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const credentials = await provider();
+
+    expect(credentials).toHaveProperty("accessKeyId");
+    expect(credentials).toHaveProperty("secretAccessKey");
+    expect(typeof credentials.accessKeyId).toBe("string");
+    expect(typeof credentials.secretAccessKey).toBe("string");
+
+    if (!credentials.accountId) {
+      console.log("Skipping account ID test not an allowlisted instance");
+      return;
```
kuhe: make this a separate test and skip the actual test instead of returning from the test

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -115,6 +111,7 @@ const getInstanceMetadataProvider = (init: RemoteProviderInit = {}) => {
 
   return async () => {
     const endpoint = await getInstanceMetadataEndpoint();
+    await throwIfImdsTurnedOff(profile, logger);
```
kuhe: should this go before the endpoint resolution?

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -144,7 +141,56 @@ const getInstanceMetadataProvider = (init: RemoteProviderInit = {}) => {
...
+  init: RemoteProviderInit = {},
+  profile?: string,
+  resetCache?: boolean
+): Promise<string> => {
+  let apiVersion: "unknown" | "extended" | "legacy" = "unknown";
+  let resolvedProfile: string | null = null;
+
+  // If resetCache is true, clear the cached profile name
+  if (resetCache) {
+    resolvedProfile = null;
```
kuhe: why set a value that was already null to null?

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
     },
   });
 
-const getProfile = async (options: RequestOptions) => (await httpRequest({ ...options, path: IMDS_PATH })).toString();
+/**
+ * @internal
+ * Checks if IMDS credential fetching is disabled through configuration
```
kuhe, line 199: description must be the first part of the tsdoc block, above the `@internal` access level annotation

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+
+  // If IMDS is disabled, throw error
+  if (disableImds) {
+    throw new CredentialsProviderError("IMDS credential fetching is disabled", { logger });
+  }
+};
+
+/**
+ * @internal
+ * Gets configured profile name from various sources
```
kuhe, line 240: description goes above annotations

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+  if (typeof name === "string" && name.trim() === "") {
+    throw new CredentialsProviderError("EC2 instance profile name cannot be empty");
+  }
+
+  return name;
+};
 
+/**
+ * @internal
+ * Gets credentials from profile
```
kuhe: description goes above annotations

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+        throw legacyError;
+      }
+    }
+    throw error;
+  }
+};
 
+/**
+ * @internal
+ * Gets credentials from specified IMDS path
```
kuhe, line 298: description goes above annotations

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+ */
+async function getCredentialsFromPath(path: string, options: RequestOptions) {
+  const response = await httpRequest({
+    ...options,
+    path,
+  });
+
+  const credentialsResponse = JSON.parse(response.toString());
```
kuhe: catch JSON parse error and treat as invalid response same as below? or does the specification require different handling for JSON errors?

packages/credential-provider-imds/src/fromInstanceMetadata.ts
```diff
@@ -154,23 +200,113 @@ const getMetadataToken = async (options: RequestOptions) =>
...
+  } catch (error) {
+    // If extended API returns 404, fall back to legacy API
+    if (error.statusCode === 404) {
+      try {
+        return await getCredentialsFromPath(IMDS_LEGACY_PATH + profile, options);
+      } catch (legacyError) {
+        if (legacyError.statusCode === 404 && init.ec2InstanceProfileName === undefined) {
+          // If legacy API also returns 404 and we're using a cached profile name,
+          // the profile might have changed - clear cache and retry
+          const newProfileName = await getImdsProfile(options, init.maxRetries ?? 3, init, profile, true);
```
kuhe: how does `init.ec2InstanceProfileName === undefined` indicate that we're using a cached profile name?

packages/credential-provider-imds/src/fromInstanceMetadata.ts
smilkuri: Yes

packages/credential-provider-imds/src/fromInstanceMetadata.ts
smilkuri: If `ec2InstanceProfileName` is provided by the user it's a direct profile specification and we would use that. 
When ec2InstanceProfileName is undefined, the profile name came from either env vars, config files or IMDS discovery which would be cached in `resolvedProfile` and it might have changed when we make subsequent calls, since the time we cached it. so we need to reset cache and try again.

kuhe:
comments for the PR as of June 26:
- variable names need to differentiate "AWS_PROFILE" init.profile, "imdsProfile", and "ec2InstanceProfile". How do these conceptually relate to each other? Are profile and imds profile separate? Is ec2InstanceProfile a subset of imdsProfile?
- I updated the PR to not use positional arguments. There were too many redundancies between elements of the `RemoteProviderInit` and positional args like maxRetries, logger, "profile" etc. and the use of "profile" as an argument name was too ambiguous.
- more tests are needed to cover variations in inputs/config

kuhe (changes requested):
the PR is on hold

PR #1624, opened 2025-06-20
fix: criteria for event stream member detection

kuhe:
corrects the criteria from https://github.com/smithy-lang/smithy-typescript/pull/1623

http payload trait is not required and the member itself will not have the streaming trait.


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/EventStreamGenerator.java
```diff
@@ -454,7 +452,7 @@ private void generateEventStreamDeserializer(GenerationContext context, UnionSha
                             });
                         });
                     });
-                    writer.write("return {$$unknown: output};");
```
kuhe, line 457: this fixes a bug introduced in https://github.com/smithy-lang/smithy-typescript/pull/572

kuhe: there's no impact because $unknown events are dropped from the asyncIterable

siddsriv (approved):

Issue #1600, opened 2025-05-20
Schemas


kuhe:
This is a tracking issue for implementing schema objects and associated features.

Since schemas are an implementation detail that is not relevant to the typical generated-client consumer, I won't go into a long explanation of what they do. They are instruction objects for request serialization and response deserialization.

Going along with this change, two features are planned:

- [x] protocol selection can be done at runtime for clients, enabling different communication formats with services supporting more than one protocol without needing to generate separate clients.
- [x] BigDecimal and BigInteger support for all protocols and formats.

----
Early PoC:  
https://github.com/smithy-lang/smithy-typescript/pull/1592
https://github.com/aws/aws-sdk-js-v3/pull/6986

----
### PRs

https://github.com/smithy-lang/smithy-typescript/pull/1593 - schema types
https://github.com/smithy-lang/smithy-typescript/pull/1594 - migrate serde functions
https://github.com/smithy-lang/smithy-typescript/pull/1595 - schema runtime classes
https://github.com/smithy-lang/smithy-typescript/pull/1599 - java codegen
https://github.com/smithy-lang/smithy-typescript/pull/1603 - BigInt and BigDecimal support in CBOR codec
https://github.com/smithy-lang/smithy-typescript/pull/1661 - event streams
https://github.com/smithy-lang/smithy-typescript/pull/1671 - event stream initial-request initial-response
https://github.com/aws/aws-sdk-js-v3/pull/7290 - awsQueryCompat
https://github.com/smithy-lang/smithy-typescript/pull/1672 - schema bucketing (later superseded by https://github.com/smithy-lang/smithy-typescript/pull/1739)

#### AWS SDK for JavaScript (downstream consumer)

https://github.com/aws/aws-sdk-js-v3/pull/7110 - AWS protocol runtime classes
https://github.com/aws/aws-sdk-js-v3/pull/7117 - AWS protocol codegen and protocol tests

#### Misc patches (pre-release)
https://github.com/aws/aws-sdk-js-v3/pull/7109 - (chore) import core submodules instead of index
https://github.com/smithy-lang/smithy-typescript/pull/1613 - namespace of synthetic base exceptions
https://github.com/smithy-lang/smithy-typescript/pull/1609 - allow schema structIterator to be run on unit schema
https://github.com/smithy-lang/smithy-typescript/pull/1610 - allow httpPrefixHeader and httpHeader bindings to read from same header
[...]

PR #1638, opened 2025-06-27
fix(core/schema): fix http binding struct iteration

kuhe:
https://github.com/smithy-lang/smithy-typescript/issues/1600

since iteration over the input was changed from input to schema keys, we need checked handling for all the member fields


.changeset/plenty-apes-accept.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
kuhe, line 2: this change doesn't need to be independently published

PR #1652, opened 2025-07-26
test(node-http-handler): add comprehensive test coverage for NodeHttpHandler

yuji-hatakeyama:
$(cat <<'EOF'
This PR adds comprehensive test coverage for the NodeHttpHandler class.

## Changes

- Added tests for connectionTimeout configuration with both options hash and provider
- Added tests for requestTimeout configuration with both options hash and provider  
- Added tests for NodeHttpHandler.create() static factory method
  - Tests return existing handler instance when passed
  - Tests create new instance for undefined/empty options
  - Tests proper configuration passing when creating new instances

## Test Plan

- All new tests pass
- No changes to existing functionality
- Follows existing test patterns in the codebase
EOF
)


yuji-hatakeyama:
Sorry I fixed format 4d097a7d3764f1ece0ec7270e179b30a8d0982bf .

On my MacOS 15.5 (M3), “yarn format” does not seem to be detected correctly.
This may be due to the cache related to turbo.

```console
% git co df469f4403738435410922d2e30b25ddc5e89269
HEAD is now at df469f44 chore: add empty changeset

% yarn format 1>/dev/null && git diff
turbo 2.3.3

% rm -rf .turbo/cache && yarn format 1>/dev/null && git diff
turbo 2.3.3

diff --git a/packages/node-http-handler/src/node-http-handler.spec.ts b/packages/node-http-handler/src/node-http-handler.spec.ts
index 2f924b54..5f219a82 100644
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -252,9 +252,12 @@ describe(“NodeHttpHandler”, () => {

     it.each([
       [“existing handler instance”, new NodeHttpHandler()],
-      [“custom HttpHandler object”, {
-        handle: vi.fn(),
-      } as any],
+      [
+        “custom HttpHandler object”,
+        {
+          handle: vi.fn(),
+        } as any,
+      ],
     ])(“returns the input handler when passed %s”, (_, handler) => {
       const result = NodeHttpHandler.create(handler);
       expect(result).toBe(handler);
```

Disabling the cache when formatting in turbo.json resolved the issue, so I will apply this only to my local environment.

kuhe (approved):

PR #1657, opened 2025-08-01
feat: replace uuid.v4 with crypto.randomUUID()

kuhe:
https://github.com/aws/aws-sdk-js-v3/issues/7212

replace uuid v4 with crypto.randomUUID()

execution environments without global crypto.randomUUID will need a replacement polyfill.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
@@ -873,13 +873,8 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
         Shape target = model.expectShape(binding.getMember().getTarget());
 
         boolean isIdempotencyToken = binding.getMember().hasTrait(IdempotencyTokenTrait.class);
-        if (isIdempotencyToken) {
-            writer
-                .addDependency(TypeScriptDependency.UUID_TYPES)
-                .addImport("v4", "generateIdempotencyToken", TypeScriptDependency.UUID);
-        }
         boolean isRequired = binding.getMember().isRequired();
-        String idempotencyComponent = (isIdempotencyToken && !isRequired) ? " ?? generateIdempotencyToken()" : "";
+        String idempotencyComponent = (isIdempotencyToken && !isRequired) ? " ?? crypto.randomUUID()" : "";
         String memberAssertionComponent = (idempotencyComponent.isEmpty() ? "!" : "");
 
         String queryValue = getInputValue(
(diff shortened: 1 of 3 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -873,13 +873,8 @@ private void writeRequestQueryParam(
...
 
         boolean isIdempotencyToken = binding.getMember().hasTrait(IdempotencyTokenTrait.class);
-        if (isIdempotencyToken) {
-            writer
-                .addDependency(TypeScriptDependency.UUID_TYPES)
-                .addImport("v4", "generateIdempotencyToken", TypeScriptDependency.UUID);
-        }
         boolean isRequired = binding.getMember().isRequired();
-        String idempotencyComponent = (isIdempotencyToken && !isRequired) ? " ?? generateIdempotencyToken()" : "";
+        String idempotencyComponent = (isIdempotencyToken && !isRequired) ? " ?? crypto.randomUUID()" : "";
```
trivikr, line 877: This is runtime agnostic file, and `crypto` is not available globally the minimum version of Node.js, i.e. 18.0.0, as per my tests

```console
$ nvm use 18.0.0
Now using node v18.0.0 (npm v8.6.0)

$ cat randomUUID.mjs 
console.log(crypto.randomUUID());

$ node randomUUID.mjs 
file:///Users/trivikr/workspace/test/randomUUID.mjs:1
console.log(crypto.randomUUID());
            ^

ReferenceError: crypto is not defined
    at file:///Users/trivikr/workspace/test/randomUUID.mjs:1:13
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)
    at async loadESM (node:internal/process/esm_loader:85:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)

Node.js v18.0.0
```

trivikr (changes requested):

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java
trivikr: The crypto was added globally as follows:
* v23.0.0 | No longer experimental.
* v19.0.0 | No longer behind --experimental-global-webcrypto CLI flag.
* v17.6.0, v16.15.0 |  Added in: v17.6.0, v16.15.0

<br class="Apple-interchange-newline">

Docs: https://nodejs.org/api/globals.html#crypto

trivikr: We can replace `uuid` with a utility in core though, where Node.js one will get it from crypto.
That utility can be removed once we drop support for Node.js 22.x

wojtekmaj: Is there something blocking us from simply _importing_ crypto module the old way? `randomUUID()` method in `node:crypto` module was added in Node.js v15.6.0, v14.17.0.

kuhe: The code that currently uses randomUUID would no longer be "isomorphic" compatible. 

We will revisit this soon, since I am in the process of rewriting that part of the SDK code.

PR #1674, opened 2025-09-02
fix: event stream serialization

kuhe:
internal P291987404

Fixes event stream serialization when the `@eventStreamPayload` member is a string. 

This case is not present in current AWS models.


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/EventStreamGenerator.java
```diff
@@ -387,7 +387,7 @@ private void writeEventBody(
                     writer.write("body = input.$L;", payloadMemberName);
                 } else if (payloadShape instanceof StringShape) {
                     writer.write("body = context.utf8Decoder(input.$L);", payloadMemberName);
-                } else if (payloadShape instanceof BlobShape || payloadShape instanceof StringShape) {
```
kuhe, line 390: this was a copy-paste error from https://github.com/smithy-lang/smithy-typescript/pull/572

smilkuri (approved):

PR #1681, opened 2025-09-04
chore: schema code size optimization

kuhe:
- Reduce the bundled code size of the runtime for schema serde, from about 2k to 1k lines of code (non-minified)

### DIFF
--- a/packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
+++ b/packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
@@ -95,13 +95,18 @@ export class NormalizedSchema implements INormalizedSchema {
    * Static constructor that attempts to avoid wrapping a NormalizedSchema within another.
    */
   public static of(ref: SchemaRef): NormalizedSchema {
-    if (Array.isArray(ref)) {
-      // An aggregate schema must be initialized with members and the member retrieved through the aggregate
-      // container.
-      throw new Error("@smithy/core/schema - may not init member schema.");
-    }
     if (ref instanceof NormalizedSchema) {
       return ref;
+    }
+    if (Array.isArray(ref)) {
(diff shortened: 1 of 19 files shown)


packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
```diff
@@ -21,23 +22,23 @@ import { StructureSchema } from "./StructureSchema";
...
+  private readonly memberTraits: SchemaTraits;
   private normalizedTraits?: SchemaTraitsObject;
 
   /**
    * @param ref - a polymorphic SchemaRef to be dereferenced/normalized.
    * @param memberName - optional memberName if this NormalizedSchema should be considered a member schema.
    */
-  public constructor(
-    private readonly ref: SchemaRef,
+  private constructor(
```
kuhe, line 40: alpha API breaking change: the only public constructor is now the static constructor.

packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
```diff
@@ -83,34 +80,30 @@ export class NormalizedSchema implements INormalizedSchema {
...
-    }
-    return isPrototype;
+    return Schema[Symbol.hasInstance].bind(this)(lhs);
   }
 
   /**
    * Static constructor that attempts to avoid wrapping a NormalizedSchema within another.
    */
-  public static of(ref: SchemaRef, memberName?: string): NormalizedSchema {
+  public static of(ref: SchemaRef): NormalizedSchema {
```
kuhe, line 97: alpha API breaking change: static public ctor no longer allows a member name.

Members must not be initialized on their own. They must be retrieved as part of an aggregate schema (struct, union, list, map, document etc.). 

This simplifies reasoning about the schema data that is present.

PR #1698, opened 2025-09-18
Bump uuid to ^11.0.0

TastyPi:
*Description of changes:*

Bump `uuid` to a version that provides types itself, allowing the removal of the dependency on `@types/uuid` which has been deprecated.

Stopped at 11 because 12.0.0 dropped CommonJS support, which this package appears to have, I assume you still want that.

I've been having issues with tools getting confused between the types from `@types/uuid` and the `uuid` package, this is one of only two projects that cause us to indirectly depend on `@types/uuid` so I'm hoping this will be the easier way to solve my issues instead of fighting the tools.

---

### DIFF
--- a/.changeset/eight-rats-roll.md
+++ b/.changeset/eight-rats-roll.md
@@ -1,0 +1,6 @@
+---
+"@smithy/middleware-retry": patch
+"@smithy/core": patch
+---
+
+Bump uuid dependency to ^11.0.0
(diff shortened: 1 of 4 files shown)


trivikr:
Thank you, @TastyPi, for your contribution.

We already had plans to remove dependency on `uuid` in https://github.com/smithy-lang/smithy-typescript/pull/1657
It was removed in https://github.com/smithy-lang/smithy-typescript/pull/1706

PR #1702, opened 2025-09-19
feat(uuid): add '@smithy/uuid' with v4 polyfill

trivikr:
*Issue #, if available:*
Required to implement https://github.com/aws/aws-sdk-js-v3/issues/7212

*Description of changes:*
Adds '@smithy/uuid' with polyfill for v4, which prefers native implementations

---

### DIFF
--- a/packages/uuid/src/v4.ts
+++ b/packages/uuid/src/v4.ts
@@ -2,6 +2,24 @@ import { randomUUID } from "./randomUUID";
 
 const decimalToHex = Array.from({ length: 256 }, (_, i) => i.toString(16).padStart(2, "0"));
 
+/**
+ * Generates a RFC4122 version 4 UUID
+ *
+ * This function generates a random UUID using one of two methods:
+ * 1. The native randomUUID() function if available
+ * 2. A fallback implementation using crypto.getRandomValues()
+ *
+ * The fallback implementation:
+ * - Generates 16 random bytes using crypto.getRandomValues()
+ * - Sets the version bits to indicate version 4
+ * - Sets the variant bits to indicate RFC4122
+ * - Formats the bytes as a UUID string with dashes
+ *
+ * @returns A version 4 UUID string in the format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
+ * where x is any hexadecimal digit and y is one of 8, 9, a, or b.
+ *
+ * @internal
+ */
 export const v4 = () => {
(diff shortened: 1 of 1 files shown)


packages/uuid/src/v4.ts
```diff
@@ -0,0 +1,39 @@
+import { randomUUID } from "./randomUUID";
+
+const decimalToHex = Array.from({ length: 256 }, (_, i) => i.toString(16).padStart(2, "0"));
+
+export const v4 = () => {
```
kuhe, line 23: add `@internal` annotation with description?

packages/uuid/README.md
```diff
@@ -0,0 +1,10 @@
+# @smithy/uuid
+
+[![NPM version](https://img.shields.io/npm/v/@smithy/uuid/latest.svg)](https://www.npmjs.com/package/@smithy/uuid)
+[![NPM downloads](https://img.shields.io/npm/dm/@smithy/uuid.svg)](https://www.npmjs.com/package/@smithy/uuid)
+
+> An internal package
+
+## Usage
+
+You probably shouldn't, at least directly.
```
kuhe: we should update the template part of this, e.g.

```
This is an internal `@smithy` package. You should not install it as a direct dependency in your applications.
```

kuhe (approved):

packages/uuid/README.md
trivikr:
```suggestion
This is an internal `@smithy` package. You should not install it as a direct dependency in your applications.
```

packages/uuid/src/v4.ts
trivikr:
```suggestion
/**
 * Generates a RFC4122 version 4 UUID
 * 
 * This function generates a random UUID using one of two methods:
 * 1. The native randomUUID() function if available
 * 2. A fallback implementation using crypto.getRandomValues()
 * 
 * The fallback implementation:
 * - Generates 16 random bytes using crypto.getRandomValues()
 * - Sets the version bits to indicate version 4
 * - Sets the variant bits to indicate RFC4122
 * - Formats the bytes as a UUID string with dashes
 * 
 * @returns A version 4 UUID string in the format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
 * where x is any hexadecimal digit and y is one of 8, 9, a, or b.
 * 
 * @internal
 */
export const v4 = () => {
```

PR #1713, opened 2025-09-26
feat: use rollup for cjs distributions

kuhe:
*Issue #, if available:*

*Description of changes:*

---


packages/service-client-documentation-generator/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "4.1.0",
   "scripts": {
     "build": "concurrently 'yarn:build:cjs' 'yarn:build:es' 'yarn:build:types && yarn build:types:downlevel'",
-    "build:cjs": "node ../../scripts/inline service-client-documentation-generator",
```
kuhe, line 6: this is not a runtime package so it doesn't need inlining

tsconfig.cjs.json
```diff
@@ -4,7 +4,7 @@
     "importHelpers": true,
     "module": "commonjs",
     "noEmitHelpers": false,
-    "target": "ES2018",
+    "target": "es2022",
```
kuhe, line 7: Why is it safe to upgrade to es2022?

Smithy-TS has the same language level support as AWS SDK JSv3. JSv3 already uses the `@tsconfig/node18` preset which has a language target of es2022.

The dist code diff for a sampling of language constructs:

source:
```ts
const x: any = undefined;

(async () => {
    const y = {
        async *[Symbol.asyncIterator]() {
            for await (const p of x) {
                yield 1;
                yield 2;
                yield x ?? x;
                yield x?.x?.()?.[0];
            }
        }
    }
})();
```

es2018:
```js
"use strict";
const x = undefined;
(async () => {
    const y = {
        async *[Symbol.asyncIterator]() {
            var _a, _b;
            for await (const p of x) {
                yield 1;
                yield 2;
                yield x !== null && x !== void 0 ? x : x;
                yield (_b = (_a = x === null || x === void 0 ? void 0 : x.x) === null || _a === void 0 ? void 0 : _a.call(x)) === null || _b === void 0 ? void 0 : _b[0];
            }
        }
    };
})();
```

es2022:
```js
"use strict";
const x = undefined;
(async () => {
    const y = {
        async *[Symbol.asyncIterator]() {
            for await (const p of x) {
                yield 1;
                yield 2;
                yield x ?? x;
                yield x?.x?.()?.[0];
            }
        }
    };
})();

```

smilkuri (approved):

Issue #1779, opened 2025-11-15
Generated client does not add headers resolved from Endpoint Ruleset to the request


humanzz:
Hello,

I'm using `typescript-client-codegen` plugin to generate a typescript client for a model where I use https://smithy.io/2.0/additional-specs/rules-engine/specification.html to add `stage` and `apiKey` client context parameters so that my service's client users can have an easy way of selecting the service's endpoint based on `stage`, and also to set the `apiKey` I provide for them.

For the generated typescript client, and despite setting the `apiKey` which should, in the case of the model below, lead to the selection of an endpoint configuration that has the `x-api-key` header being set to the value of the `apiKey` provided, the final outgoing request never has that header set.

For context, I'm generating an AWS SDK v2-based client (starting from Smithy -> C2J -> generated Java Client, so I know it's a bit different), and following the same approach I'm able to set the `apiKey` on the client, and the endpoint header is included in the outgoing request.

I tried debugging the generated typescript client, and I could see that
- Till https://github.com/smithy-lang/smithy-typescript/blob/a0e0035db609963933ac8499d0117384e7dbc8a2/packages/middleware-endpoint/src/endpointMiddleware.ts#L57, the endpoint is resolved correctly, and has the `x-api-key` header
- I suspect it's in https://github.com/smithy-lang/smithy-typescript/blob/a0e0035db609963933ac8499d0117384e7dbc8a2/packages/core/src/submodules/protocols/requestBuilder.ts#L33 where the `HttpRequest` is created, ignoring the headers from the endpoint

Is that an intentional behavior? If not, then this is a feature request to respect the endpoint headers.

```smithy

$version: "2.0"

namespace software.MyService

use smithy.rules#clientContextParams
use smithy.rules#endpointRuleSet
use smithy.rules#endpointTests

apply MyService @clientContextParams(
    Stage: { type: "string", documentation: "Service stage (beta|prod)" }
    ApiKey: { type: "string", documentation: "ApiKey" }
)

apply MyService @endpointRuleSet({
    version: "1.0"
    parameters: {
        Stage: { required: false, documentation: "Specify the stage (beta|prod) so the endpoint is automatically set. If unspecified, prod endpoint will be used. Setting an explicit endpoint override is supported.", type: "String" }
        ApiKey: { required: false, documentation: "ApiKey", type: "string" }
        Endpoint: { builtIn: "SDK::Endpoint", required: false, documentation: "Override the endpoint used to send requests", type: "String" }
[...]

PR #1788, opened 2025-11-24
fix: handle clientContextParam collisions with builtin config keys 

smilkuri:
*Issue #, if available:*

https://github.com/smithy-lang/smithy-typescript/issues/1779
codegen v3:https://github.com/aws/aws-sdk-js-v3/pull/7549

*Description of changes:*

This fixes TypeScript client codegen conflict between known config keys and clientContextParams by adding the nested clientContextParams object to isolate conflicting parameters while maintaining backward compatibility and proper precedence.

--

### DIFF
--- a/private/my-local-model-schema/src/XYZServiceClient.ts
+++ b/private/my-local-model-schema/src/XYZServiceClient.ts
@@ -249,7 +249,8 @@ export class XYZServiceClient extends __Client<
         httpAuthSchemeParametersProvider: defaultXYZServiceHttpAuthSchemeParametersProvider,
         identityProviderConfigProvider: async (config: XYZServiceClientResolvedConfig) =>
           new DefaultIdentityProviderConfig({
-            "smithy.api#httpApiKeyAuth": config.apiKey,}),
+            "smithy.api#httpApiKeyAuth": config.apiKey,
+          }),
       })
     );
     this.middlewareStack.use(getHttpSigningPlugin(this.config));
--- a/private/my-local-model-schema/src/endpoint/endpointResolver.ts
+++ b/private/my-local-model-schema/src/endpoint/endpointResolver.ts
@@ -7,8 +7,7 @@ import { ruleSet } from "./ruleset";
 
 const cache = new EndpointCache({
   size: 50,
-  params: ["ApiKey",
- "endpoint"],
+  params: ["ApiKey", "endpoint"],
 });
 
 /**
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ClientConfigKeys.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ClientConfigKeys.java
@@ -2,7 +2,6 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-
 package software.amazon.smithy.typescript.codegen.endpointsV2;
 
 import java.util.Map;
@@ -103,13 +102,15 @@ public final class ClientConfigKeys {
      * @return filtered custom context parameters
      */
     public static Map<String, String> getCustomContextParams(
-            Map<String, String> clientContextParams,
-            Map<String, String> builtInParams
-        ) {
+        Map<String, String> clientContextParams,
+        Map<String, String> builtInParams
+    ) {
         Map<String, String> customContextParams = new java.util.HashMap<>();
         for (Map.Entry<String, String> entry : clientContextParams.entrySet()) {
-            if (!builtInParams.containsKey(entry.getKey())
-                    && !KNOWN_CONFIG_KEYS.contains(entry.getKey())) {
+            if (
+                !builtInParams.containsKey(entry.getKey())
+                    && !KNOWN_CONFIG_KEYS.contains(entry.getKey())
+            ) {
                 customContextParams.put(entry.getKey(), entry.getValue());
             }
         }
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
@@ -109,7 +109,8 @@ public final class EndpointsV2Generator implements Runnable {
                 Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
                 builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
                 Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
-                    clientContextParams, builtInParams
+                    clientContextParams,
+                    builtInParams
                 );
 
                 writer.writeDocs("@public");
@@ -125,8 +126,13 @@ public final class EndpointsV2Generator implements Runnable {
                             writer.indent();
                             ObjectNode ruleSet = endpointRuleSetTrait.getRuleSet().expectObjectNode();
                             ruleSet.getObjectMember("parameters").ifPresent(parameters -> {
-                                parameters.accept(new RuleSetParametersVisitor(writer,
-                                    clientContextParamsExcludingBuiltIns, true));
+                                parameters.accept(
+                                    new RuleSetParametersVisitor(
+                                        writer,
+                                        clientContextParamsExcludingBuiltIns,
+                                        true
+                                    )
+                                );
                             });
                             writer.dedent();
                             writer.write("};");
@@ -140,8 +146,10 @@ public final class EndpointsV2Generator implements Runnable {
                             String paramName = entry.getKey();
                             String localName = EndpointsParamNameMap
                                 .getLocalName(paramName);
-                            if (!ClientConfigKeys.isKnownConfigKey(paramName)
-                                && !ClientConfigKeys.isKnownConfigKey(localName)) {
+                            if (
+                                !ClientConfigKeys.isKnownConfigKey(paramName)
+                                    && !ClientConfigKeys.isKnownConfigKey(localName)
+                            ) {
                                 directParams.put(paramName, entry.getValue());
                             }
                         });
@@ -149,7 +157,8 @@ public final class EndpointsV2Generator implements Runnable {
                         ruleSet.getObjectMember("parameters").ifPresent(parameters -> {
                             parameters.accept(new RuleSetParametersVisitor(writer, directParams, true));
                         });
-                });
+                    }
+                );
 
                 writer.write("");
                 writer.writeDocs("@public");
@@ -176,15 +185,15 @@ public final class EndpointsV2Generator implements Runnable {
                             .ifPresent(parameters -> {
                                 parameters.accept(new RuleSetParametersVisitor(writer, true));
                             });
-                            writer.write(
-                                "defaultSigningName: \"$L\",",
-                                settings.getDefaultSigningName()
-                            );
-                            if (ruleSetParameterFinder.hasCustomClientContextParams()) {
-                                ruleSetParameterFinder.writeConfigResolverNestedClientContextParams(writer);
-                            }
-                        });
-                    }
+                        writer.write(
+                            "defaultSigningName: \"$L\",",
+                            settings.getDefaultSigningName()
+                        );
+                        if (ruleSetParameterFinder.hasCustomClientContextParams()) {
+                            ruleSetParameterFinder.writeConfigResolverNestedClientContextParams(writer);
+                        }
+                    });
+                }
                 );
 
                 writer.write("");
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
@@ -173,7 +173,8 @@ public class RuleSetParameterFinder {
         Map<String, String> builtInParams = getBuiltInParams();
         builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
         Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
-            clientContextParams, builtInParams
+            clientContextParams,
+            builtInParams
         );
         return !customContextParams.isEmpty();
     }
@@ -186,7 +187,8 @@ public class RuleSetParameterFinder {
         Map<String, String> builtInParams = getBuiltInParams();
         builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
         Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
-            clientContextParams, builtInParams
+            clientContextParams,
+            builtInParams
         );
         ObjectNode ruleSet = ruleset.getRuleSet().expectObjectNode();
         ruleSet.getObjectMember("parameters").ifPresent(parameters -> {
@@ -236,13 +238,15 @@ public class RuleSetParameterFinder {
         Map<String, String> builtInParams = getBuiltInParams();
         builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
         Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
-            clientContextParams, builtInParams
+            clientContextParams,
+            builtInParams
         );
         ObjectNode ruleSet = ruleset.getRuleSet().expectObjectNode();
         boolean hasDefaultsForResolve = false;
         if (ruleSet.getObjectMember("parameters").isPresent()) {
             ObjectNode parameters = ruleSet.getObjectMember("parameters").get().expectObjectNode();
-            hasDefaultsForResolve = customContextParams.entrySet().stream()
+            hasDefaultsForResolve = customContextParams.entrySet()
+                .stream()
                 .anyMatch(entry -> {
                     ObjectNode paramNode = parameters.getObjectMember(entry.getKey()).orElse(null);
                     return paramNode != null && paramNode.containsMember("default");
@@ -251,7 +255,7 @@ public class RuleSetParameterFinder {
         if (hasDefaultsForResolve) {
             writer.write(
                 "clientContextParams: Object.assign(clientContextParamDefaults, "
-                + "options.clientContextParams),"
+                    + "options.clientContextParams),"
             );
         } else {
             writer.write(
--- a/smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2GeneratorTest.java
+++ b/smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2GeneratorTest.java
@@ -183,7 +183,9 @@ public class EndpointsV2GeneratorTest {
                   };
                   stage?: string | undefined | Provider<string | undefined>;
                   endpoint?: string | Provider<string> | Endpoint | Provider<Endpoint> | EndpointV2 | Provider<EndpointV2>;
-                }"""));
+                }"""
+            )
+        );
     }
 
     private MockManifest testEndpoints(String filename) {


packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
```diff
@@ -78,6 +78,11 @@ export const resolveParams = async <
   instructionsSupplier: EndpointParameterInstructionsSupplier,
   clientConfig: Partial<EndpointResolvedConfig<T>> & Config
 ) => {
+  // Initialize clientContextParams to empty object if undefined
+  const config = clientConfig as typeof clientConfig & { clientContextParams?: Record<string, unknown> };
+  if (config.clientContextParams === undefined) {
+    config.clientContextParams = {};
+  }
```
kuhe: no need to do this, since the only place where it's read also accounts for it being undefined

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
```diff
@@ -18,7 +18,11 @@ export const createConfigValueProvider = <Config extends Record<string, unknown>
   config: Config
 ) => {
   const configProvider = async () => {
-    const configValue: unknown = config[configKey] ?? config[canonicalEndpointParamKey];
+    // Check clientContextParams first for client context parameters
+    const clientContextParams = config.clientContextParams as Record<string, unknown> | undefined;
+    const nestedValue: unknown = clientContextParams?.[configKey] ?? clientContextParams?.[canonicalEndpointParamKey];
+    // Fall back to direct config properties
+    const configValue: unknown = nestedValue ?? config[configKey] ?? config[canonicalEndpointParamKey];
```
kuhe: This should only happen when the configuration being read is a client context parameter. 

Users should not be able to place other kinds of config into this object and have them be read. It will create a contract where none is intended.

smithy-typescript-codegen/src/test/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2GeneratorTest.java
```diff
@@ -56,11 +56,16 @@ public void containsExtraContextParameter() {
                   return Object.assign(options, {
                     stage: options.stage ?? "production",
                     defaultSigningName: "",
+                    clientContextParams: Object.assign(clientContextParamDefaults, options.clientContextParams ?? {}),
```
kuhe: no need to default the input to `?? {}`, since object assign will ignore undefined values.

.changeset/stale-phones-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/middleware-endpoint": minor
+---
+
+handle clientContextParam collisions with builtin config keys
```
kuhe, line 5: add the diff for `make build generate-protocol-tests` to the PR

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -109,39 +111,119 @@ private void generateEndpointParameters() {
...
+                writer.addImport("EndpointParameters", "__EndpointParameters", TypeScriptDependency.SMITHY_TYPES);
+                writer.addImport("Provider", null, TypeScriptDependency.SMITHY_TYPES);
+                Map<String, String> clientContextParams =
+                    ruleSetParameterFinder.getClientContextParams();
+                Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
+                builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+                Set<String> knownConfigKeys = Set.of(
+                    "apiKey", "retryStrategy", "requestHandler");
+                // Generate clientContextParams with all params excluding built-ins
+                Map<String, String> customerContextParams = new HashMap<>();
```
kuhe: call this `customContextParams`, not customer

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -109,39 +111,119 @@ private void generateEndpointParameters() {
         this.delegator.useFileWriter(
             Paths.get(CodegenUtils.SOURCE_FOLDER, ENDPOINT_FOLDER, ENDPOINT_PARAMETERS_FILE).toString(),
             writer -> {
-                writer.addTypeImport("EndpointParameters", "__EndpointParameters", TypeScriptDependency.SMITHY_TYPES);
-                writer.addTypeImport("Provider", null, TypeScriptDependency.SMITHY_TYPES);
+                writer.addImport("EndpointParameters", "__EndpointParameters", TypeScriptDependency.SMITHY_TYPES);
+                writer.addImport("Provider", null, TypeScriptDependency.SMITHY_TYPES);
```
kuhe: no reason to change these from type import to import

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -109,39 +111,119 @@ private void generateEndpointParameters() {
...
+                Map<String, String> clientContextParams =
+                    ruleSetParameterFinder.getClientContextParams();
+                Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
+                builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+                Set<String> knownConfigKeys = Set.of(
+                    "apiKey", "retryStrategy", "requestHandler");
```
kuhe: make this static final on the class

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -1,5 +1,5 @@
 // smithy-typescript generated code
-import type { Endpoint, EndpointParameters as __EndpointParameters, EndpointV2, Provider } from "@smithy/types";
+import { type Endpoint, type EndpointV2, EndpointParameters as __EndpointParameters, Provider } from "@smithy/types";
 
 /**
  * @public
```
kuhe, line 5: define a custom endpoint ruleset on the local model so that this codegen diff shows the feature is working

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
kuhe: on second thought this should be handled by a separate class. 

seed the class with the config keys seen in https://github.com/aws/aws-sdk-js-v3/blob/f806a2b9eded9f82d488f3d81db36b9cba9da220/private/aws-client-api-test/src/client-interface-tests/client-s3/impl/initializeWithMaximalConfiguration.ts except those specific to S3, then create a public Java static method to add keys to the list.

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
```diff
@@ -9,16 +9,29 @@ import type { Endpoint, EndpointV2 } from "@smithy/types";
...
-    const configValue: unknown = config[configKey] ?? config[canonicalEndpointParamKey];
+    let configValue: unknown;
+
+    if (isClientContextParam) {
+      // For client context parameters, check clientContextParams first
+      const clientContextParams = config.clientContextParams as Record<string, unknown> | undefined;
+      const nestedValue: unknown = clientContextParams?.[configKey] ?? clientContextParams?.[canonicalEndpointParamKey];
+      configValue = nestedValue ?? config[configKey] ?? config[canonicalEndpointParamKey];
+    } else {
+      // For built-in parameters, only check direct config properties
```
kuhe: not just for built-ins, for anything other than client context params

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
```diff
@@ -9,16 +9,29 @@ import type { Endpoint, EndpointV2 } from "@smithy/types";
...
+  isClientContextParam = false
 ) => {
   const configProvider = async () => {
-    const configValue: unknown = config[configKey] ?? config[canonicalEndpointParamKey];
+    let configValue: unknown;
+
+    if (isClientContextParam) {
+      // For client context parameters, check clientContextParams first
+      const clientContextParams = config.clientContextParams as Record<string, unknown> | undefined;
+      const nestedValue: unknown = clientContextParams?.[configKey] ?? clientContextParams?.[canonicalEndpointParamKey];
```
kuhe: is there any case where the nested object does not use the canonical parameter key?

If not, there's no reason to check the `configKey` entry for the nested object.

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
```diff
@@ -90,8 +90,10 @@ export const resolveParams = async <
         endpointParams[name] = commandInput[instruction.name] as string | boolean;
         break;
       case "clientContextParams":
+        endpointParams[name] = await createConfigValueProvider<Config>(instruction.name, name, clientConfig, true)();
+        break;
       case "builtInParams":
-        endpointParams[name] = await createConfigValueProvider<Config>(instruction.name, name, clientConfig)();
+        endpointParams[name] = await createConfigValueProvider<Config>(instruction.name, name, clientConfig, false)();
```
kuhe:
```ts
case "clientContextParams":
case "builtInParams":
  endpointParams[name] = await createConfigValueProvider<Config>(
    instruction.name, name, clientConfig, instruction.type !== "builtInParams"
  )();
```

private/my-local-model/src/endpoint/EndpointParameters.ts
```diff
@@ -5,25 +5,53 @@ import type { Endpoint, EndpointParameters as __EndpointParameters, EndpointV2,
...
+const clientContextParamDefaults = {
+  apiKey: "default-api-key",
+  customParam: "default-custom-value",
+} as const;
+
 export const resolveClientEndpointParameters = <T>(
   options: T & ClientInputEndpointParameters
 ): T & ClientResolvedEndpointParameters => {
   return Object.assign(options, {
+    apiKey: options.apiKey ?? "default-api-key",
```
kuhe: This is the main point of the fix, but this shouldn't appear here, since it will conflict with the same key potentially be added by the api key auth scheme.

private/my-local-model/src/endpoint/ruleset.ts
```diff
@@ -5,34 +5,33 @@ export const ruleSet: RuleSetObject = {
...
+      type: "String",
+      required: true,
+      default: "default-api-key",
+      documentation: "API key for service authentication",
+    },
+    customParam: {
+      type: "String",
+      required: true,
+      default: "default-custom-value",
+      documentation: "Custom parameter for testing",
```
kuhe, line 27: add a few more client context params like `boolean` and I believe there is one more type which is `string[]`, right?

add a conflicting and non-conflicting param for each type

kuhe: and add some that don't have default values (or are they required?)

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ClientConfigKeys.java
```diff
@@ -0,0 +1,116 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
kuhe, line 4: use the short copyright header

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ClientConfigKeys.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Set;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Manages known client configuration keys that should not be placed in
+ * clientContextParams.
+ */
+@SmithyInternalApi
+public final class ClientConfigKeys {
+    private static final Set<String> KNOWN_CONFIG_KEYS = new HashSet<>();
```
kuhe: make this a concurrent hashset?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
                     "export interface ClientInputEndpointParameters {",
                     "}",
                     () -> {
-                        Map<String, String> clientInputParams = ruleSetParameterFinder.getClientContextParams();
-                        //Omit Endpoint params that should not be a part of the ClientInputEndpointParameters interface
-                        Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
-                        builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
-                        clientInputParams.putAll(builtInParams);
-
+                        if (!customContextParams.isEmpty()) {
```
kuhe:
```suggestion
                        if (ruleSetParameterFinder.hasCustomClientContextParams()) {
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
             writer -> {
                 writer.addTypeImport("EndpointParameters", "__EndpointParameters", TypeScriptDependency.SMITHY_TYPES);
                 writer.addTypeImport("Provider", null, TypeScriptDependency.SMITHY_TYPES);
+                Map<String, String> clientContextParams =
+                    ruleSetParameterFinder.getClientContextParams();
+                Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
```
kuhe, line 109: I'm going to suggest some new signatures for this class (ruleSetParameterFinder) to encapsulate the new logic in this file diff

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
-                        clientInputParams.putAll(builtInParams);
-
+                        if (!customContextParams.isEmpty()) {
+                            writer.write("clientContextParams?: {");
+                            writer.indent();
+                            ObjectNode ruleSet = endpointRuleSetTrait.getRuleSet().expectObjectNode();
+                            ruleSet.getObjectMember("parameters").ifPresent(parameters -> {
+                                parameters.accept(new RuleSetParametersVisitor(
+                                    writer, customContextParams, true));
+                            });
```
kuhe, line 136:
```java
writer.indent();
ruleSetParameterFinder.writeInputConfigCustomClientContextParams(writer);
writer.dedent();
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
+                            writer.dedent();
+                            writer.write("};");
+                        }
+                        // Add direct params (built-ins + custom context params)
```
kuhe: for the rest of the block:

```java
ruleSetParameterFinder.writeInputConfigDirectClientContextParams(writer);
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
                 writer.openBlock(
-                    """
-                    export type ClientResolvedEndpointParameters = Omit<ClientInputEndpointParameters, "endpoint"> & {
-                    """,
+                    "export type ClientResolvedEndpointParameters = "
+                        + "Omit<ClientInputEndpointParameters, " + omitFields + "> & {",
                     "};",
                     () -> {
                         writer.write("defaultSigningName: string;");
+                        if (!customContextParams.isEmpty()) {
```
kuhe:
```suggestion
                        if (ruleSetParameterFinder.hasCustomClientContextParams()) {
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
+                            writer.write("clientContextParams: {");
+                            writer.indent();
+                            ObjectNode ruleSet = endpointRuleSetTrait.getRuleSet().expectObjectNode();
+                            ruleSet.getObjectMember("parameters").ifPresent(parameters -> {
+                                parameters.accept(new RuleSetParametersVisitor(
+                                    writer, customContextParams, true));
+                            });
+                            writer.dedent();
```
kuhe:
```java
writer.indent();
ruleSetParameterFinder.writeResolvedConfigCustomClientContextParams(writer);
writer.dedent();
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,37 +113,104 @@ private void generateEndpointParameters() {
...
+                            writer.write("};");
+                        }
                     }
                 );
+                // Generate clientContextParamDefaults only if there are customer context params
+                if (!customContextParams.isEmpty()) {
```
kuhe:
```java
if (ruleSetParameterFinder.hasCustomClientContextParams()) {
  ruleSetParameterFinder.writeNestedClientContextParamDefaults(writer);
}
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -157,6 +226,34 @@ private void generateEndpointParameters() {
                                 "defaultSigningName: \"$L\",",
                                 settings.getDefaultSigningName()
                             );
+                            // Only generate clientContextParams if there are customer context params
+                            if (!customContextParams.isEmpty()) {
```
kuhe:
```java
ruleSetParameterFinder.writeConfigResolverNestedClientContextParams(writer);
```

smithy-typescript-protocol-test-codegen/model/my-local-model/main.smithy
```diff
@@ -3,9 +3,34 @@ $version: "2.0"
...
 
 use smithy.protocols#rpcv2Cbor
+use smithy.rules#clientContextParams
+use smithy.rules#endpointRuleSet
 
 @rpcv2Cbor
 @documentation("xyz interfaces")
+@clientContextParams(
+    apiKey: { type: "string", documentation: "API key for authentication" }
+    customParam: { type: "string", documentation: "Custom parameter" }
```
kuhe, line 13: going to comment again here since this is the source file and ruleset.ts is generated:

add some more client context params with varying permutations:
- conflicting / non-conflicting
- string / string[] / boolean
- has default + required / no default + optional

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
smilkuri: I don't think so

private/my-local-model/src/endpoint/EndpointParameters.ts
smilkuri: This is generated because the parameter has a default value set defined in `@endpointRuleSet`. If a default value is not defined it doesn't generate this.

kuhe: that does not matter - a conflicting key cannot be generated here

packages/eventstream-serde-universal/src/eventstream-cbor.integ.spec.ts
```diff
@@ -9,7 +9,8 @@ describe("local model integration test for cbor eventstreams", () => {
   it("should read and write cbor event streams", async () => {
     const client = new XYZService({
       endpoint: "https://localhost",
-    });
+      apiKey: { apiKey: "test-api-key" },
+    } as any);
```
kuhe: there should be no `as any` here. This test is from a user perspective (integ test means calls user-facing public APIs only), so the types must align with the user intention.

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.spec.ts
```diff
@@ -58,4 +58,31 @@ describe(createConfigValueProvider.name, () => {
     expect(await createConfigValueProvider("v1", "endpoint", config)()).toEqual(sampleUrl);
     expect(await createConfigValueProvider("v2", "endpoint", config)()).toEqual(sampleUrl);
   });
+
+  it("should prioritize clientContextParams over direct properties", async () => {
+    const config = {
+      apiKey: "direct-api-key",
+      clientContextParams: {
+        apiKey: "nested-api-key",
```
kuhe: Although conflicts don't matter at the unit level (and this code is not aware of what is considered a conflict),

change this test key to something else that doesn't conflict, like `stage: string = "beta" | "prod"`. This implies that apiKey is a string at the base config level, but it isn't (when present).

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.spec.ts
```diff
@@ -58,4 +58,31 @@ describe(createConfigValueProvider.name, () => {
...
+      },
+    };
+    expect(await createConfigValueProvider("apiKey", "apiKey", config, true)()).toEqual("nested-api-key");
+  });
+
+  it("should fall back to direct property when clientContextParams is not provided", async () => {
+    const config = {
+      customParam: "direct-value",
+    };
```
kuhe, line 75: add
```js
{
  clientContextParams: {
    customParam: undefined
  }
}
```

to emphasize the desired behavior

packages/util-retry/src/retries.integ.spec.ts
```diff
@@ -20,7 +20,8 @@ describe("retries", () => {
   it("should retry throttling and transient-error status codes", async () => {
     const client = new XYZService({
       endpoint: "https://localhost/nowhere",
-    });
+      apiKey: { apiKey: "test-api-key" },
+    } as any);
```
kuhe: remove `as any`

private/my-local-model-schema/src/auth/httpAuthSchemeProvider.ts
```diff
@@ -1,12 +1,16 @@
...
+  type HandlerExecutionContext,
+  type HttpAuthOption,
+  type HttpAuthScheme,
+  type HttpAuthSchemeParameters,
+  type HttpAuthSchemeParametersProvider,
+  type HttpAuthSchemeProvider,
+  type Provider,
+  ApiKeyIdentity,
+  ApiKeyIdentityProvider,
+  HttpApiKeyAuthLocation,
```
kuhe, line 13: these new imports look like types. Change the codegen part to use `addTypeImport()` instead, or if imported via symbol, ensure the symbol has the `typeOnly` property.

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -5,37 +5,83 @@ import type { Endpoint, EndpointParameters as __EndpointParameters, EndpointV2,
...
 
+/**
+ * @internal
+ */
+const clientContextParamDefaults = {
+  nonConflictingParam: "non-conflict-default",
+  customParam: "default-custom-value",
+  debugMode: false,
+  enableFeature: true,
+} as const;
```
kuhe, line 36: I realized that this set of defaults cannot be the full set of clientContextParams that have default values.

For example, consider a non-conflicting clientContextParam `stage` with default of `"prod"`. The user may instantiate like this:
```js
new Client({
  stage: "prod"
});
```

but because the default value is written to the nested property, it will take precedence. 

```js
new Client({
  stage: "beta" // <- set by user
  clientContextParams: {
    stage: "prod" // <- set by default, but higher priority
  }
});
```

Therefore, this set of defaults must only be the _conflicting_ clientContextParam defaults. In such cases, the default value applied at the root config level will take care of setting the correct default value.

```js
const resolvedConfig = {
  stage: "prod", // set by resolveClientEndpointParameters as default 
  conflictingParam: "(any)", // conflicting param may have separate default
  clientContextParams: {
    stage: undefined // no default needs to be assigned here
    conflictingParam: "conflictingParamDefault" // do set values for conflicting params
  } 
}
```

Please also add an integration test for this scenario.

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -5,37 +5,83 @@ import type { Endpoint, EndpointParameters as __EndpointParameters, EndpointV2,
...
     defaultSigningName: "",
+    clientContextParams: Object.assign(clientContextParamDefaults, options.clientContextParams),
   });
 };
 
 /**
  * @internal
  */
 export const commonParams = {
+  ApiKey: { type: "clientContextParams", name: "apiKey" },
```
kuhe, line 58: wasn't there an automatic conversion of the param name to camelCase in the endpoints generator? If it couldn't find an existing built-in name.

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -5,37 +5,83 @@ import type { Endpoint, EndpointParameters as __EndpointParameters, EndpointV2,
...
+  enableFeature: { type: "clientContextParams", name: "enableFeature" },
   endpoint: { type: "builtInParams", name: "endpoint" },
 } as const;
 
 /**
  * @internal
  */
 export interface EndpointParameters extends __EndpointParameters {
-  endpoint?: string | undefined;
+  endpoint: string;
```
kuhe, line 72: not saying it's wrong necessarily, but why did endpoint become required in the change?

private/my-local-model/src/auth/httpAuthExtensionConfiguration.ts
```diff
@@ -1,5 +1,5 @@
 // smithy-typescript generated code
-import type { HttpAuthScheme } from "@smithy/types";
+import { type HttpAuthScheme, ApiKeyIdentity, ApiKeyIdentityProvider } from "@smithy/types";
```
kuhe: change these to type imports

private/my-local-model/src/auth/httpAuthSchemeProvider.ts
```diff
@@ -1,12 +1,16 @@
...
+  type HandlerExecutionContext,
+  type HttpAuthOption,
+  type HttpAuthScheme,
+  type HttpAuthSchemeParameters,
+  type HttpAuthSchemeParametersProvider,
+  type HttpAuthSchemeProvider,
+  type Provider,
+  ApiKeyIdentity,
+  ApiKeyIdentityProvider,
+  HttpApiKeyAuthLocation,
```
kuhe, line 13: use type imports

packages/util-retry/src/retries.integ.spec.ts
smilkuri: If I removed `any` it was throwing a type incompatibility error.

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
smilkuri: Makes sense. Will modify.

packages/eventstream-serde-universal/src/clientContextParams-precedence.integ.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+describe("client context parameters precedence integration test", () => {
+  it("should handle conflicting vs non-conflicting parameter precedence correctly", async () => {
+    // For non-conflicting parameter root level takes precedence over nested defaults
+    const clientWithNonConflicting = new XYZService({
+      endpoint: "https://localhost",
+      apiKey: async () => ({ apiKey: "test-key" }),
+      customParam: "user-custom-value",
+      clientContextParams: {
+        apiKey: "test-key",
+        customParam: "default-custom-value",
```
kuhe:
```suggestion
        customParam: "nested-custom-value",
```

this is not the default, it is given by the user at the client instantiation site

packages/eventstream-serde-universal/src/clientContextParams-precedence.integ.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+    });
+
+    // Verify the resolved config has the user's root-level value
+    const resolvedConfig = (clientWithNonConflicting as any).config;
+    expect(resolvedConfig.customParam).toBe("user-custom-value");
```
kuhe: This should be "nested-custom-value". When a value exists in both places the nested value should be used.

packages/eventstream-serde-universal/src/clientContextParams-precedence.integ.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    const clientWithConflicting = new XYZService({
+      endpoint: "https://localhost",
+      apiKey: async () => ({ apiKey: "auth-key" }),
+      clientContextParams: {
+        apiKey: "endpoint-key",
+      },
+    });
+
+    // Verify that both auth and endpoint contexts can coexist
+    const resolvedConfigConflicting = (clientWithConflicting as any).config;
```
kuhe: why `as any`?

private/my-local-model-schema/src/XYZServiceClient.ts
```diff
@@ -255,19 +250,19 @@ export class XYZServiceClient extends __Client<
...
     this.middlewareStack.use(getSchemaSerdePlugin(this.config));
     this.middlewareStack.use(getRetryPlugin(this.config));
     this.middlewareStack.use(getContentLengthPlugin(this.config));
     this.middlewareStack.use(
       getHttpAuthSchemeEndpointRuleSetPlugin(this.config, {
         httpAuthSchemeParametersProvider: defaultXYZServiceHttpAuthSchemeParametersProvider,
         identityProviderConfigProvider: async (config: XYZServiceClientResolvedConfig) =>
-          new DefaultIdentityProviderConfig({}),
+          new DefaultIdentityProviderConfig({
+            "smithy.api#httpApiKeyAuth": config.apiKey,}),
```
kuhe: a new line is needed after the key-value pair

private/my-local-model-schema/src/auth/httpAuthExtensionConfiguration.ts
```diff
@@ -1,5 +1,5 @@
 // smithy-typescript generated code
-import type { HttpAuthScheme } from "@smithy/types";
+import { type HttpAuthScheme, ApiKeyIdentity, ApiKeyIdentityProvider } from "@smithy/types";
```
kuhe: these must be type imports

private/my-local-model-schema/src/endpoint/endpointResolver.ts
```diff
@@ -7,7 +7,8 @@ import { ruleSet } from "./ruleset";
 
 const cache = new EndpointCache({
   size: 50,
-  params: ["endpoint"],
+  params: ["ApiKey",
+ "endpoint"],
```
kuhe: no line break between params

smithy-typescript-protocol-test-codegen/model/my-local-model/main.smithy
```diff
@@ -9,12 +9,14 @@ use smithy.rules#endpointRuleSet
 @rpcv2Cbor
 @documentation("xyz interfaces")
 @httpApiKeyAuth(name: "X-Api-Key", in: "header")
+@httpApiKeyAuth(name: "X-Api-Key", in: "header")
```
smilkuri: Duplicate entry will remove it

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/ClientConfigKeys.java
```diff
@@ -0,0 +1,117 @@
...
+     * Get custom context parameters by filtering out built-in and known config
+     * keys.
+     *
+     * @param clientContextParams all client context parameters
+     * @param builtInParams       built-in parameters
+     * @return filtered custom context parameters
+     */
+    public static Map<String, String> getCustomContextParams(
+            Map<String, String> clientContextParams,
+            Map<String, String> builtInParams) {
```
kuhe:
```suggestion
            Map<String, String> builtInParams
    ) {
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/EndpointsV2Generator.java
```diff
@@ -111,21 +112,46 @@ private void generateEndpointParameters() {
             writer -> {
                 writer.addTypeImport("EndpointParameters", "__EndpointParameters", TypeScriptDependency.SMITHY_TYPES);
                 writer.addTypeImport("Provider", null, TypeScriptDependency.SMITHY_TYPES);
+                Map<String, String> clientContextParams =
+                    ruleSetParameterFinder.getClientContextParams();
+                Map<String, String> builtInParams = ruleSetParameterFinder.getBuiltInParams();
+                builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+                Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
+                    clientContextParams, builtInParams);
```
kuhe:
```suggestion
                    clientContextParams, builtInParams
                );
```
align close parens with the indentation of the line containing its starting counterpart

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
```diff
@@ -173,6 +174,98 @@ public Map<String, String> getBuiltInParams() {
...
 
+    /**
+     * Check if there are custom client context parameters.
+     */
+    public boolean hasCustomClientContextParams() {
+        Map<String, String> clientContextParams = getClientContextParams();
+        Map<String, String> builtInParams = getBuiltInParams();
+        builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+        Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
+            clientContextParams, builtInParams);
```
kuhe:
```suggestion
            clientContextParams, builtInParams
        );
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
```diff
@@ -173,6 +174,98 @@ public Map<String, String> getBuiltInParams() {
...
+
+    /**
+     * Write custom client context parameters to TypeScript writer.
+     */
+    public void writeInputConfigCustomClientContextParams(TypeScriptWriter writer) {
+        Map<String, String> clientContextParams = getClientContextParams();
+        Map<String, String> builtInParams = getBuiltInParams();
+        builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+        Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
+            clientContextParams, builtInParams);
```
kuhe:
```suggestion
            clientContextParams, builtInParams
        );
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java
```diff
@@ -173,6 +174,98 @@ public Map<String, String> getBuiltInParams() {
...
+
+    /**
+     * Write config resolver nested client context parameters to TypeScript writer.
+     */
+    public void writeConfigResolverNestedClientContextParams(TypeScriptWriter writer) {
+        Map<String, String> clientContextParams = getClientContextParams();
+        Map<String, String> builtInParams = getBuiltInParams();
+        builtInParams.keySet().removeIf(OmitEndpointParams::isOmitted);
+        Map<String, String> customContextParams = ClientConfigKeys.getCustomContextParams(
+            clientContextParams, builtInParams);
```
kuhe:
```suggestion
            clientContextParams, builtInParams
        );
```

smithy-typescript-protocol-test-codegen/model/my-local-model/main.smithy
```diff
@@ -3,9 +3,66 @@ $version: "2.0"
...
+    version: "1.0"
+    parameters: {
+        endpoint: { builtIn: "SDK::Endpoint", required: true, documentation: "The endpoint used to send the request.", type: "String" }
+        ApiKey: { required: false, documentation: "ApiKey", type: "String" }
+        region: { type: "String", required: false, documentation: "AWS region" }
+        customParam: { type: "String", required: true, default: "default-custom-value", documentation: "Custom parameter for testing" }
+        enableFeature: { type: "Boolean", required: true, default: true, documentation: "Feature toggle with default" }
+        debugMode: { type: "Boolean", required: true, default: false, documentation: "Debug mode with default" }
+        nonConflictingParam: { type: "String", required: true, default: "non-conflict-default", documentation: "Non-conflicting with default" }
+        logger: { type: "String", required: true, default: "default-logger", documentation: "Conflicting logger with default" }
```
kuhe: all endpoint param types need to be lowercase `string` `boolean`. Add a `stringArray ` one too, like "additionalFeatures".

private/my-local-model/src/auth/httpAuthSchemeProvider.ts
smilkuri: `HttpApiKeyAuthLocation` can't be changed to type only. It's accessing a runtime value `HttpApiKeyAuthLocation.HEADER`

packages/eventstream-serde-universal/src/clientContextParams-precedence.integ.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { describe, expect, test as it } from "vitest";
+import { XYZService } from "xyz";
+
+describe("client context parameters precedence integration test", () => {
+  it("should handle conflicting vs non-conflicting parameter precedence correctly", async () => {
+    // For non-conflicting params
```
siddsriv, line 6: clarifying nit: does this mean non-conflicting because the value set for `apiKey` is the same at root and nested? my interpretation of "non-conflicting" was more strict, in that the entire interface is clean and no duplicate params exist at all at the root and nested level. but that's not a blocker/problem, just wanted to double-check

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
```diff
@@ -9,16 +9,29 @@ import type { Endpoint, EndpointV2 } from "@smithy/types";
...
 ) => {
   const configProvider = async () => {
-    const configValue: unknown = config[configKey] ?? config[canonicalEndpointParamKey];
+    let configValue: unknown;
+
+    if (isClientContextParam) {
+      // For client context parameters, check clientContextParams first
+      const clientContextParams = config.clientContextParams as Record<string, unknown> | undefined;
+      const nestedValue: unknown = clientContextParams?.[canonicalEndpointParamKey];
+      configValue = nestedValue ?? config[configKey] ?? config[canonicalEndpointParamKey];
```
siddsriv, line 29: could `canonicalEndpointParamKey` and `configKey` differ? (ApiKey -- from the ruleset vs apiKey -- from the client config) would that create any problems for the precedence lookup?

siddsriv: for example, would this case correctly prefer the nested value? 
```ts
const client = new XYZService({
  apiKey: async () => ({ apiKey: "auth-secret" }),   
  clientContextParams: {
    apiKey: "endpoint-header-key"
  }
});
```
if the nested value looks for the canonical name, would it find `ApiKey` (`name` in `commonParams`) and then fall back to the root value when it doesn't find `clientContextParams.[ApiKey]`?

packages/eventstream-serde-universal/src/clientContextParams-precedence.integ.spec.ts
siddsriv: from context reading ahead, it does seem like that's the case. cool!

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
smilkuri: With the current logic using canonicalEndpointParamKey. It would look for `clientContextParams["ApiKey"]` → not found. So it would fall back to `config["apiKey"]` → finds "auth-secret"
But it should have found `"endpoint-header-key"` from clientContextParams["apiKey"].

I think this should be  `const nestedValue: unknown = clientContextParams?.[configKey]`;

packages/middleware-endpoint/src/adaptors/createConfigValueProvider.ts
kuhe:
```ts
clientContextParams?.[configKey] ?? clientContextParams?.[canonicalEndpointParamKey];
```

is `canonicalEndpointParamKey` ApiKey and `configKey` apiKey?
if they are different then both still need to be checked

smilkuri: The clientContextParams uses a [local name](https://github.com/smithy-lang/smithy-typescript/pull/1788/files#diff-66cacae2905f9523659a509fab2d638e9818ad9c5b432c7d810f3345ff748f77R137) so wouldn't it be just `clientContextParams?.[configKey]` do we still need to check `clientContextParams?.[canonicalEndpointParamKey]`?

kuhe (approved):

PR #1794, opened 2025-12-02
fix(codegen): generate access level annotations on endpoint components

kuhe:
Adds public or internal tsdoc tags to generated endpoint components.

### DIFF
--- a/private/my-local-model-schema/src/endpoint/EndpointParameters.ts
+++ b/private/my-local-model-schema/src/endpoint/EndpointParameters.ts
@@ -8,10 +8,16 @@ export interface ClientInputEndpointParameters {
   endpoint?: string | Provider<string> | Endpoint | Provider<Endpoint> | EndpointV2 | Provider<EndpointV2>;
 }
 
+/**
+ * @public
+ */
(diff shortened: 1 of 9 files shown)


private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -8,10 +8,16 @@ export interface ClientInputEndpointParameters {
   endpoint?: string | Provider<string> | Endpoint | Provider<Endpoint> | EndpointV2 | Provider<EndpointV2>;
 }
 
+/**
+ * @public
```
kuhe, line 12: resolved params are public because the client's resolved config type intersection is itself public

private/my-local-model-schema/src/endpoint/EndpointParameters.ts
```diff
@@ -8,10 +8,16 @@ export interface ClientInputEndpointParameters {
...
+ */
 export type ClientResolvedEndpointParameters = Omit<ClientInputEndpointParameters, "endpoint"> & {
   defaultSigningName: string;
 };
 
+/**
+ * @internal
```
kuhe, line 19: the other elements are internal because they do not give information to the reader/user on how to use the client, nor would they be imported by a consumer.

PR #1824, opened 2025-12-31
set default of generateSchemas to true

kuhe:
set default of generateSchemas to `true`.

The option to set this value at all will be removed in a future version.

### DIFF
--- a/smithy-typescript-protocol-test-codegen/smithy-build.json
+++ b/smithy-typescript-protocol-test-codegen/smithy-build.json
@@ -25,6 +25,7 @@
             "license": "Apache-2.0"
           },
           "private": true,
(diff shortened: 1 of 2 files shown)


smithy-typescript-protocol-test-codegen/smithy-build.json
```diff
@@ -24,7 +24,7 @@
             },
             "license": "Apache-2.0"
           },
-          "private": true,
+          "private": true,    "generateSchemas": false,
```
trivikr: nit: two different lines
```suggestion
          "private": true,
          "generateSchemas": false,
```

trivikr (approved):

PR #1839, opened 2026-01-13
feat(typecheck): add api for runtime typecheck

kuhe:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/7610

*Description of changes:*
Adds rttc package

### DIFF
--- a/packages/typecheck/src/getRuntimeTypecheckPlugin.ts
+++ b/packages/typecheck/src/getRuntimeTypecheckPlugin.ts
@@ -22,7 +22,7 @@ export function getRuntimeTypecheckPlugin<InputType extends object = any, Output
 ): Pluggable<InputType, OutputType> {
   return {
     applyToStack: (commandStack: MiddlewareStack<InputType, OutputType>) => {
-      commandStack.add(rttcMiddleware(options), rttcMiddlewareOptions);
+      commandStack.add(runtimeTypecheckMiddleware(options), runtimeTypecheckOptions);
     },
   };
 }
@@ -30,7 +30,7 @@ export function getRuntimeTypecheckPlugin<InputType extends object = any, Output
 /**
  * @internal
  */
-const rttcMiddlewareOptions: InitializeHandlerOptions = {
+const runtimeTypecheckOptions: InitializeHandlerOptions = {
   name: "runtimeTypecheckMiddleware",
   step: "initialize",
   tags: ["RUNTIME_TYPECHECK"],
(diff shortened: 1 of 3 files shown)


packages/typecheck/src/getRuntimeTypecheckPlugin.ts
```diff
@@ -0,0 +1,85 @@
...
+  name: "runtimeTypecheckMiddleware",
+  step: "initialize",
+  tags: ["RUNTIME_TYPECHECK"],
+  override: true,
+};
+
+/**
+ * @internal
+ */
+const rttcMiddleware =
```
trivikr: nit: should this be named runtimeTypecheckMiddleware, same as line 34?
```suggestion
const runtimeTypecheckMiddleware =
```

packages/typecheck/src/validateSchema.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      for (let i = 0; i < data.length; ++i) {
+        const value = data[i];
+        if (value == null) {
+          if (!sparse) {
+            errors.push(`${path}[${i}]: should be non-null.`);
+          }
+          continue;
+        }
+        errors.push(...validateSchema(list$, value, path + `[${i}]`));
```
trivikr: nit: use else, since early return is not used at other places?
```suggestion
        } else {
          errors.push(...validateSchema(list$, value, path + `[${i}]`));
         }
```

packages/typecheck/src/validateSchema.ts
```diff
@@ -0,0 +1,106 @@
...
+      const sparse = !!$.getMergedTraits().sparse;
+      const map$ = $.getValueSchema();
+      for (const [key, value] of Object.entries(data)) {
+        if (value == null) {
+          if (!sparse) {
+            errors.push(`${path}[${key}]: should be non-null.`);
+          }
+          continue;
+        }
+        errors.push(...validateSchema(map$, value, path + `["${key}"]`));
```
trivikr: nit: use else?
```suggestion
        } else {
          errors.push(...validateSchema(map$, value, path + `["${key}"]`));
         }
```

packages/typecheck/src/validateSchema.ts
```diff
@@ -0,0 +1,106 @@
...
+      const keys = new Set(Object.keys(data));
+      let required = ($.getSchema() as StaticStructureSchema)?.[6] ?? 0;
+      for (const [member, member$] of $.structIterator()) {
+        keys.delete(member);
+        const isRequired = required-- > 0;
+        if (isRequired && (data as any)[member] == null) {
+          errors.push(`${path}.${member}: is required.`);
+        } else {
+          errors.push(...validateSchema(member$, (data as any)[member], path + `.${member}`));
+        }
```
trivikr, line 98: nit: store value in a variable?
```suggestion
        const value = (data as any)[member];
        const isRequired = required-- > 0;
        if (isRequired && value == null) {
          errors.push(`${path}.${member}: is required.`);
        } else {
          errors.push(...validateSchema(member$, value, path + `.${member}`));
        }
```

PR #1853, opened 2026-01-23
feat(smithy-client): add pagination/waiters to aggregate clients

kuhe:
*Issue #, if available:*
n/a

*Description of changes:*

Adds paginators and waiters as methods on aggregated clients.

## "bare-bones" modular client
```ts
// modular syntax
import {
  XYZClient,
  GetNumbersCommand,
  paginateGetNumbers,
  waitUntilNumbersAligned,
} from "@namespace/client-xyz";

const client = new XYZClient();

await client.send(new GetNumbersCommand({ bigInteger: 1n }));

for await (const page of paginateGetNumbers({ client }, { bigInteger: 1n })) {
  // page
}

await waitUntilNumbersAligned({ client }, { bigInteger: 1n });
```

## aggregated client
```ts
// aggregated syntax (calls modular syntax internally)

import { XYZ } from "@namespace/client-xyz";

const xyz = new XYZ();

await xyz.getNumbers({ bigInteger: 1n });

for await (const page of xyz.paginateGetNumbers(
  { bigInteger: 1n },
  {
    /* optional paginator config */
  },
)) {
  // page
}

await xyz.waitUntilNumbersAligned(
  { bigInteger: 1n },
  {
    /* optional waiter config */
  },
);

```

### DIFF
--- a/packages/smithy-client/src/create-aggregated-client.ts
+++ b/packages/smithy-client/src/create-aggregated-client.ts
@@ -1,19 +1,44 @@
+import type { PaginationConfiguration, WaiterConfiguration } from "@smithy/types";
+
 import type { Client } from "./client";
+
+/**
+ * @internal
+ */
+type AggregatedClientPaginationConfiguration = Omit<PaginationConfiguration, "client">;
+
+/**
+ * @internal
+ */
+type AggregatedClientWaiterConfiguration<C> = Omit<WaiterConfiguration<C>, "client">;
 
 /**
  * @internal
  *
  * @param commands - command lookup container.
- * @param client - client instance on which to add aggregated methods.
+ * @param Client - client instance on which to add aggregated methods.
+ * @param options
+ * @param options.paginators - paginator functions.
+ * @param options.waiters - waiter functions.
+ *
  * @returns an aggregated client with dynamically created methods.
  */
 export const createAggregatedClient = (
   commands: Record<string, any>,
-  Client: { new (...args: any): Client<any, any, any, any> }
+  Client: { new (...args: any): Client<any, any, any, any> },
+  options?: {
+    paginators?: Record<string, any>;
+    waiters?: Record<string, any>;
+  }
 ): void => {
-  for (const command of Object.keys(commands)) {
-    const CommandCtor = commands[command];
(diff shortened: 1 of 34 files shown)


private/my-local-model-schema/src/XYZService.ts
```diff
@@ -55,11 +69,32 @@ export interface XYZService {
...
   ): void;
+
+  /**
+   * @see {@link GetNumbersCommand}
+   * @param args - command input.
+   * @param paginationConfig - optional pagination config.
+   * @returns AsyncIterable of {@link GetNumbersCommandOutput}.
+   */
+  paginateGetNumbers(
+    args?: GetNumbersCommandInput,
```
siddsriv, line 80: why is this optional?

packages/smithy-client/src/create-aggregated-client.ts
```diff
@@ -1,19 +1,41 @@
...
+ * @param waiters - waiter functions.
+ *
  * @returns an aggregated client with dynamically created methods.
  */
 export const createAggregatedClient = (
   commands: Record<string, any>,
-  Client: { new (...args: any): Client<any, any, any, any> }
+  Client: { new (...args: any): Client<any, any, any, any> },
+  paginators?: Record<string, any>,
+  waiters?: Record<string, any>
```
trivikr: Make this an options object?
```suggestion
  options?: {
    paginators?: Record<string, any>;
    waiters?: Record<string, any>;
  }
```

private/my-local-model-schema/src/XYZService.ts
kuhe: you can paginate something with no args. It implies you're starting from the first page, and the resource is account-specific rather than data within a resource that has multiples.

For example, you can paginate buckets without args, but you cannot paginate dynamodb items because they are separated by tables.

packages/smithy-client/src/create-aggregated-client.ts
kuhe: updated in https://github.com/smithy-lang/smithy-typescript/compare/8deccb89930947af619af246606faaa56769d32f..885a11af9854aecb006eaae2dd44f312639df1b2

PR #1854, opened 2026-01-26
test: generate serde benchmark based on testCase tag

kuhe:
*Issue #, if available:*
JS-6588

*Description of changes:*

This adds a customization to protocol test generation that when a serde-benchmark tag is seen on the testcase, generate a variant form of test that measures serde benchmarks instead of request/response correctness.

Example output:
```
GetNumbersRequestExample:SerdeBenchmark:Request
┌─────────┬──────────────┐
│ (index) │ Values       │
├─────────┼──────────────┤
│ n       │ '    10,000' │
│ p50     │ '    37,667' │
│ p95     │ '    43,959' │
│ p99     │ '    60,292' │
│ mean    │ '    40,856' │
│ stdDev  │ '    36,843' │
└─────────┴──────────────┘
=============================== Distribution Viz ===============================
 <= 4395
. <= 8791
. <= 13187
. <= 17583
. <= 21979
. <= 26375
. <= 30771
. <= 35167
............................................................................ <= 39563
............ <= 43959
... <= 48354
. <= 52750
. > 52750
================================================================================

stdout | test/functional/rpcv2cbor.spec.ts > GetNumbersResponseExample:SerdeBenchmark:Response
GetNumbersResponseExample:SerdeBenchmark:Response
┌─────────┬──────────────┐
│ (index) │ Values       │
├─────────┼──────────────┤
│ n       │ '    10,000' │
│ p50     │ '    11,167' │
│ p95     │ '    13,542' │
│ p99     │ '    22,458' │
│ mean    │ '    12,919' │
│ stdDev  │ '    29,984' │
└─────────┴──────────────┘
=============================== Distribution Viz ===============================
 <= 1354
. <= 2708
. <= 4062
. <= 5416
. <= 6771
. <= 8125
. <= 9479
... <= 10833
................................................................................ <= 12187
...... <= 13542
... <= 14896
. <= 16250
. > 16250
================================================================================

 ✓ test/functional/rpcv2cbor.spec.ts (2 tests) 1705ms
   ✓ GetNumbersRequestExample:SerdeBenchmark:Request  891ms
   ✓ GetNumbersResponseExample:SerdeBenchmark:Response  813ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Start at  11:36:12
   Duration  2.09s (transform 180ms, setup 0ms, import 293ms, tests 1.70s, environment 0ms)
```

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
@@ -57,6 +57,7 @@ import software.amazon.smithy.protocoltests.traits.HttpResponseTestCase;
 import software.amazon.smithy.protocoltests.traits.HttpResponseTestsTrait;
 import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator;
 import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator.GenerationContext;
+import software.amazon.smithy.typescript.codegen.knowledge.ServiceClosure;
 import software.amazon.smithy.typescript.codegen.util.PropertyAccessor;
 import software.amazon.smithy.utils.IoUtils;
 import software.amazon.smithy.utils.MapUtils;
@@ -81,6 +82,10 @@ public final class HttpProtocolTestGenerator implements Runnable {
 
     private static final Logger LOGGER = Logger.getLogger(HttpProtocolTestGenerator.class.getName());
     private static final String TEST_CASE_FILE_TEMPLATE = "test/functional/%s.spec.ts";
+    private static final String SERDE_BENCHMARK_TAG = "serde-benchmark";
+    private static final String WARMUP_ITERATIONS = "10_000";
+    private static final String BENCHMARK_ITERATIONS = "10_000";
+    private static final String BENCHMARK_TIMEOUT = "60_000";
 
     private final TypeScriptSettings settings;
     private final Model model;
@@ -93,6 +98,7 @@ public final class HttpProtocolTestGenerator implements Runnable {
     private final TestFilter testFilter;
     private final MalformedRequestTestFilter malformedRequestTestFilter;
     private final GenerationContext context;
+    private final ServiceClosure closure;
 
     private TypeScriptWriter writer;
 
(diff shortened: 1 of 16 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -142,15 +145,27 @@ private void generateClientOperationTests(OperationShape operation, OperationInd
                 .getTrait(HttpRequestTestsTrait.class)
                 .ifPresent(trait -> {
                     for (HttpRequestTestCase testCase : trait.getTestCasesFor(AppliesTo.CLIENT)) {
-                        onlyIfProtocolMatches(testCase, () -> generateClientRequestTest(operation, testCase));
+                        onlyIfProtocolMatches(testCase, () -> {
+                            if (testCase.hasTag(SERDE_BENCHMARK_TAG)) {
+                                generateClientRequestBenchmark(operation, testCase);
```
siddsriv, line 185: nit: should we rename to `generateRequestBenchmark` to keep uniformity between this and `generateResponseBenchmark`?

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
```diff
@@ -81,6 +81,9 @@ public final class HttpProtocolTestGenerator implements Runnable {
 
     private static final Logger LOGGER = Logger.getLogger(HttpProtocolTestGenerator.class.getName());
     private static final String TEST_CASE_FILE_TEMPLATE = "test/functional/%s.spec.ts";
+    private static final String SERDE_BENCHMARK_TAG = "serde-benchmark";
+    private static final String WARMUP_ITERATIONS = "10000";
+    private static final String BENCHMARK_ITERATIONS = "10000";
```
siddsriv: should these be made configurable params when running the test? 

are these static so that we don't pressure memory (or GC pressure) with too many samples? even then, maybe something like reservoir sampling would be good

siddsriv: to note, 10k samples should run ok and should be statistically significant, so this isn't a blocker

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
kuhe: these mirror the asymmetrically named methods: 
- generateClientRequestBenchmark
- generateResponseBenchmark

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java
kuhe: I'll make these values into a const at the start of the test so it can be edited more easily when run.
It can then accept an env var with slight further modification.

imo GC isn't something to be avoided because under real load GC must run anyway.

PR #1866, opened 2026-02-07
fix(node-http-handler): avoid calling httpRequest.end() with empty Buffer

kuhe:
- ~~deferred load for node:http (https://github.com/aws/aws-sdk-js-v3/issues/6914)~~
  - this change was backed out of this PR
- avoid empty buffers to httpRequest.end() (https://github.com/aws/aws-sdk-js-v3/issues/6426)
- rewrite some super sketchy tests from https://github.com/smithy-lang/smithy-typescript/pull/1636

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -1,6 +1,6 @@
 import { HttpRequest } from "@smithy/protocol-http";
-import http, { request as hRequest } from "http";
-import https, { request as hsRequest } from "https";
+import http, { request as hRequest } from "node:http";
+import https, { request as hsRequest } from "node:https";
 import { afterEach, beforeEach, describe, expect, test as it, vi } from "vitest";
 
 import { NodeHttpHandler } from "./node-http-handler";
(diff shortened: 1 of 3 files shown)


packages/node-http-handler/src/node-http-handler.spec.ts
```diff
@@ -412,51 +450,4 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
...
-
-      // per-request timeout takes precedence
-      expect(testTimeout(5000, 100)).toBe(100);
-
-      // fallback to handler config
-      expect(testTimeout(200, undefined)).toBe(200);
-      expect(testTimeout(200)).toBe(200);
-    });
-
-    it("should pass correct timeout values to internal functions", async () => {
```
kuhe, line 435: 😐

packages/node-http-handler/src/node-http-handler.spec.ts
```diff
@@ -412,51 +450,4 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
       );
     });
   });
-
-  describe("per-request requestTimeout", () => {
-    it("should use per-request timeout over handler config timeout", () => {
```
kuhe, line 417: 😐

PR #1871, opened 2026-02-16
fix(core): skip assigning undefined for absent response body fields

smilkuri:
*Issue #, if available:*
P383491202

*Description of changes:*
After schema-serde was enabled in [v3.930.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.930.0), fields absent from the HTTP response body are being set to undefined on the output object. The root cause is the copy loop in `deserializeResponse` was iterating through all schema-defined `nonHttpBindingMembers` and assigning undefined for the fields that were absent.

---

### DIFF
--- a/packages/core/src/submodules/protocols/HttpBindingProtocol.spec.ts
+++ b/packages/core/src/submodules/protocols/HttpBindingProtocol.spec.ts
@@ -11,6 +11,7 @@ import type {
   ListSchemaModifier,
   MapSchemaModifier,
   MetadataBearer,
+  NumericSchema,
   OperationSchema,
   ResponseMetadata,
   SerdeFunctions,
@@ -197,7 +198,6 @@ describe(HttpBindingProtocol.name, () => {
         "my-header": "header-value",
       },
     };
-
     const dataObject = {};
     await deserializeHttpMessage(
       [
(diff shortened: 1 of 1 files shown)


packages/core/src/submodules/protocols/HttpBindingProtocol.ts
```diff
@@ -238,7 +238,9 @@ export abstract class HttpBindingProtocol extends HttpProtocol {
       if (bytes.byteLength > 0) {
         const dataFromBody = await deserializer.read(ns, bytes);
         for (const member of nonHttpBindingMembers) {
-          dataObject[member] = dataFromBody[member];
+          if (member in dataFromBody){
```
kuhe:
```suggestion
          if (dataFromBody[member] != null){
```

packages/core/src/submodules/protocols/HttpBindingProtocol.spec.ts
```diff
@@ -298,6 +297,31 @@ describe(HttpBindingProtocol.name, () => {
...
+  it("should not create undefined fields when deserializing empty object with schema", async () => {
+    const protocol = new StringRestProtocol();
+    const response = new HttpResponse({
+      statusCode: 200,
+      headers: {},
+      body: Readable.from(JSON.stringify({})),
+    });
+
+    const output = (await protocol.deserializeResponse(
+      op("", "", 0, "unit", [3, "", "", 0, ["prop", "num"], [0, 9]]),
```
kuhe: 9 is not a schema indicator, use this to give the numbers meaning:

```js
0 satisfies StringSchema,
1 satisfies NumericSchema,
```

the types can be imported from `@smithy/types`.

packages/core/src/submodules/protocols/HttpBindingProtocol.spec.ts
```diff
@@ -298,6 +297,31 @@ describe(HttpBindingProtocol.name, () => {
     expect(streamProgress).toBe(100);
   });
 
+  it("should not create undefined fields when deserializing empty object with schema", async () => {
```
kuhe:
```suggestion
  it("should not create undefined fields when deserializing non-http-binding members of an output shape", async () => {
```

kuhe (approved):

PR #1880, opened 2026-02-21
fix(util-endpoints): reject non-ASCII input in substring per Smithy spec

### LINKED ISSUE #1097: util-endpoints: substring allows non-ASCII characters
The [`substring`](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/util-endpoints/src/lib/substring.ts) function should follow the [Smithy spec](https://smithy.io/2.0/additional-specs/rules-engine/standard-library.html?highlight=substring#substring-function):

> The function MUST return an empty optional when the input contains non-ASCII characters.

TrevorBurnham:
*Issue #, if available:*

Fixes #1097

*Description of changes:*

The `substring` function in `@smithy/util-endpoints` did not reject inputs containing non-ASCII characters, which is required by the [Smithy spec](https://smithy.io/2.0/additional-specs/rules-engine/standard-library.html#substring-function):

> The function MUST return an empty optional when the input contains non-ASCII characters.

The fix adds a `/[^\u0000-\u007f]/` check to the early-return guard, returning `null` for any input outside the ASCII range. Test cases added for `\u0080`, accented, and CJK characters.

---

### DIFF
--- a/.changeset/ten-lobsters-juggle.md
+++ b/.changeset/ten-lobsters-juggle.md
@@ -2,4 +2,4 @@
 "@smithy/util-endpoints": minor
 ---
 
-reject non-ASCII input in substring
+return empty when given non-ASCII input in substring


.changeset/ten-lobsters-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/util-endpoints": minor
+---
+
+reject non-ASCII input in substring
```
kuhe:
```suggestion
return empty when given non-ASCII input in substring
```

PR #1886, opened 2026-02-21
fix: reject aborted requests with AbortSignal.reason

### LINKED ISSUE #1360: Aborted requests should be rejected with AbortSignal.reason
The current behavior of aborted request cause the promise to reject with a simple `Error`. e.g.:

https://github.com/smithy-lang/smithy-typescript/blob/efcbbf42baf48ed19d2634555b2545acbd974321/packages/node-http-handler/src/node-http-handler.ts#L171-L173

https://github.com/smithy-lang/smithy-typescript/blob/efcbbf42baf48ed19d2634555b2545acbd974321/packages/node-http-handler/src/node-http-handler.ts#L249-L251

This makes error handling more difficult, as the caller does not know which AbortSignal e.g. when used with `AbortSignal.any` caused the request to be aborted.

I recommend changing the code to reject with `AbortSignal.reason` when an AbortSignal is aborted (at least for native `AbortSignal` implementations).

TrevorBurnham:
*Issue #, if available:*

Fixes #1360

*Description of changes:*

When a request is aborted, all three HTTP handlers (`NodeHttpHandler`, `NodeHttp2Handler`, `FetchHttpHandler`) reject with a generic `new Error("Request aborted")`. This makes it impossible for callers to determine which `AbortSignal` caused the abort, especially when using `AbortSignal.any()`.

The fix adds a `buildAbortError` helper to each handler that uses `AbortSignal.reason` when available:

- If `reason` is an `Error`, it is used directly (preserving the caller's error type and stack).
- If `reason` is a non-Error value (e.g. a string), it is wrapped in an `Error` with `name: "AbortError"`.
- If no `reason` is set (older runtimes or bare `{ aborted: true }` signals), the existing `"Request aborted"` behavior is preserved.

```ts
// Before — always a generic error
const abortError = new Error("Request aborted");
abortError.name = "AbortError";
reject(abortError);

// After — uses the signal's reason when available
const abortError = buildAbortError(abortSignal);
reject(abortError);
```

All six abort sites across the three handlers were updated (two per handler: the early-abort check and the `onAbort` listener).

Tests added to `node-http-handler.spec.ts`:
- Rejects with the custom `Error` reason directly
- Falls back to default `"Request aborted"` when no reason is set
- Wraps non-Error string reasons in an `Error` with `name: "AbortError"`

---

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -6,6 +6,7 @@ import { Agent as hAgent, request as hRequest } from "http";
 import type { RequestOptions } from "https";
 import { Agent as hsAgent, request as hsRequest } from "https";
 
+import { buildAbortError } from "./build-abort-error";
 import { NODEJS_TIMEOUT_ERROR_CODES } from "./constants";
 import { getTransformedHeaders } from "./get-transformed-headers";
 import { setConnectionTimeout } from "./set-connection-timeout";
@@ -196,8 +197,7 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
 
       // if the request was already aborted, prevent doing extra work
       if (abortSignal?.aborted) {
-        const abortError = new Error("Request aborted");
-        abortError.name = "AbortError";
+        const abortError = buildAbortError(abortSignal);
         reject(abortError);
         return;
       }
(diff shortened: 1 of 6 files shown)


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -354,3 +352,20 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
     return this.config ?? {};
   }
 }
+
+/**
+ * Builds an abort error, using the AbortSignal's reason if available.
+ */
+function buildAbortError(abortSignal?: { reason?: unknown }): Error {
```
kuhe: should appear only once in this package

TrevorBurnham: Updated, thanks!

PR #1888, opened 2026-02-21
feat(codegen): include @deprecated message and since in generated JSDoc

### LINKED ISSUE #1425: @deprecated `message` and `since` in JSDoc
The `@deprecated` trait goes into the JSDoc of the relevant field, but the `message:` and `since:` strings are not put anywhere. Usually deprecation has a "see other usage" or some sort of explanation that is helpful here, and it'd be nice for clients to see that in the documentation of that type/field.

TrevorBurnham:
*Issue #, if available:*

Fixes #1425

*Description of changes:*

The `@deprecated` trait in Smithy supports `message` and `since` fields, but the TypeScript codegen was only including `message` in the generated JSDoc and completely ignoring `since`. This meant useful context like "Use FooV2 instead" or "deprecated since 2024-01-01" was lost in the generated client code.

This PR adds a `buildDeprecationAnnotation` helper to `TypeScriptWriter` that renders both fields into the JSDoc:

```
message only:    @deprecated Use FooV2 instead
since only:      @deprecated
                 @since 2024-01-01
both:            @deprecated Use FooV2 instead
                 @since 2024-01-01
neither:         @deprecated deprecated    (existing fallback behavior)
```

Updated all three sites that generate `@deprecated` annotations:
- `TypeScriptWriter.writeShapeDocs` — shapes with a `DocumentationTrait`
- `TypeScriptWriter.writeMemberDocs` — member shapes (including inherited deprecation from targets)
- `CommandGenerator` — operations without a `DocumentationTrait`

Unit tests added for all four field combinations (message only, since only, both, neither).

---

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
@@ -282,8 +282,7 @@ public final class TypeScriptWriter extends SymbolWriter<TypeScriptWriter, Impor
                 docs = docs.replace("{", "\\{").replace("}", "\\}");
                 if (shape.getTrait(DeprecatedTrait.class).isPresent()) {
                     DeprecatedTrait deprecatedTrait = shape.expectTrait(DeprecatedTrait.class);
-                    String deprecationMessage = deprecatedTrait.getMessage().orElse("deprecated");
-                    String deprecationAnnotation = "@deprecated " + deprecationMessage;
+                    String deprecationAnnotation = buildDeprecationAnnotation(deprecatedTrait);
                     docs = docs + "\n\n" + deprecationAnnotation;
                 }
                 docs = preprocessor.apply(docs);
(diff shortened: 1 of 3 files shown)


kuhe:
`@since` is not a tsdoc annotation and would have to be synthesized into the deprecation message

TrevorBurnham:
> @since is not a tsdoc annotation and would have to be synthesized into the deprecation message

Updated, thanks!

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -344,6 +342,28 @@ private boolean isTargetDeprecated(Model model, MemberShape member) {
...
+        StringBuilder annotation = new StringBuilder("@deprecated");
+        String message = trait.getMessage().orElse(null);
+        String since = trait.getSince().orElse(null);
+
+        if (message != null && since != null) {
+            annotation.append(" ").append(message).append(" (since ").append(since).append(")");
+        } else if (message != null) {
+            annotation.append(" ").append(message);
+        } else if (since != null) {
+            annotation.append(" Deprecated since ").append(since).append(".");
```
kuhe:
```suggestion
            annotation.append(" since ").append(since);
```

smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptWriter.java
```diff
@@ -344,6 +342,28 @@ private boolean isTargetDeprecated(Model model, MemberShape member) {
...
+        } else {
+            annotation.append(" deprecated");
+        }
+
+        return annotation.toString();
```
kuhe, line 367:
```suggestion
        if (!annotation.endsWith(".")) {
          annotation.append(".");
        }
        return annotation.toString();
```

PR #1892, opened 2026-02-23
fix: Filter out http-bound members from schema passed to payload serializer

haydenbaker:
*Context:*
- When a member has both `@httpHeader` and `@idempotencyToken`, the header is serialized correctly but the body also had that same member populated (with a different value)
- The payload serializer (e.g. `JsonShapeSerializer`) iterates schema members (not input keys), finds the member is undefined, sees `@IdempotencyToken`, and generates a fresh UUID bound to the request body (not the header)
  - If the server rejects unknown fields in a request, then the request will be rejected and will never be processable
- The root cause is that `HttpBindingProtocol.serializeRequest()` passes the full schema to the payload serializer including http-bound members (headers, params, etc)
  - The `delete input[memberName]` trick mostly works, but not for fields that get special treatment in the payload serializer
    - In this case, `@idempotencyToken` breaks it because the serializer treats `undefined` + `@idempotencyToken` as "generate a value."

*Description of changes:*
- This adds `filterMembers(cacheKey, predicate)` to `NormalizedSchema` as a wrapper that overrides `structIterator()` so that a "view" of the schema can be passed (without explicit copying)
  - It's also cached, this is just an optimization to not recompute the view every time
- This method is used in `HttpBindingProtocol` to give the payload serializer a view of the schema that excludes HTTP-bound members, so that it doesn't try to consume already bound members
- There may be a better way to do this, but I'll leave that to the team to figure out if this isn't sufficient (I didn't want to make a super disruptive change that could have higher surface area) 

---

### DIFF
--- a/packages/core/src/submodules/protocols/HttpBindingProtocol.ts
+++ b/packages/core/src/submodules/protocols/HttpBindingProtocol.ts
@@ -21,6 +21,12 @@ import { extendedEncodeURIComponent } from "./extended-encode-uri-component";
 import { HttpProtocol } from "./HttpProtocol";
 
 /**
+ * Cache key for the body-only schema view that excludes HTTP-bound members.
+ * @internal
+ */
+const HTTP_BODY_MEMBERS = Symbol.for("@smithy/http-body-members");
+
+/**
  * Base for HTTP-binding protocols. Downstream examples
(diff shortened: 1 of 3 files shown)


kuhe:
Thanks, I'll take a look soon, and add some changes & tests as needed.

PR #1912, opened 2026-03-08
feat(schema): add caching to NormalizedSchema.of() and translateTraits()

TrevorBurnham:
This PR adds caching to `NormalizedSchema.of()` and `translateTraits()` to eliminate redundant object construction on hot serialization/deserialization paths. Operation schemas are static and reused across every request, so the same `SchemaRef` objects are passed to `NormalizedSchema.of()` repeatedly — each time triggering `deref()`, `translateTraits()`, and constructor work that produces an identical result.

### Changes

**`NormalizedSchema.of()` — WeakMap cache**

A module-level `WeakMap<object, NormalizedSchema>` caches constructed instances keyed by the dereferenced object-type schema. On subsequent calls with the same schema reference, the cached instance is returned directly. The existing identity pass-through (`sc instanceof NormalizedSchema`) and member schema branches execute before any cache interaction and are unchanged. Primitive schema refs (numbers, strings) bypass the cache since they cannot be WeakMap keys and are trivially cheap to construct.

**`translateTraits()` — Map cache with freeze**

A module-level `Map<number, SchemaTraitsObject>` caches translated trait objects keyed by numeric bitmask. Since there are only ~128 possible bitmask values, the Map is bounded. Cached objects are frozen with `Object.freeze()` before storage to prevent accidental mutation of shared entries. Object-type indicators (already a `SchemaTraitsObject`) continue to be returned as-is without caching.

### Benchmarks

| Function | Before (no cache) | After (cached) | Speedup |
|---|---|---|---|
| `NormalizedSchema.of()` | 1.2M ops/sec | 14.0M ops/sec | **11.5x** |
| `translateTraits()` | 1.8M ops/sec | 18.7M ops/sec | **10.6x** |

Both functions sit on the hot path for every serialize/deserialize call. Operation schemas are static and reused across requests, so after the first call the cache is always warm.

---

### DIFF
--- a/packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
+++ b/packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
@@ -42,21 +42,26 @@ import { translateTraits } from "./translateTraits";
 const anno = {
   // reference to structure's member iterator.
   it: Symbol.for("@smithy/nor-struct-it"),
+
+  /**
+   * Key for a cached NormalizedSchema on a $SchemaRef object.
+   * For non-object refs, we use `simpleSchemaCache(N|S)`.
+   */
+  ns: Symbol.for("@smithy/ns"),
 };
 
 /**
- * Module-level cache for NormalizedSchema.of().
- * Uses WeakMap keyed on object-type schemas (arrays, objects) so entries
- * are garbage-collected when the schema is no longer referenced.
+ * Cache for numeric schemaRef. Having separate cache objects for number/string improves
+ * lookup performance.
+ * @internal
  */
-const schemaCache = new WeakMap<object, NormalizedSchema>();
+export const simpleSchemaCacheN: Array<NormalizedSchema | undefined> = [];
 
 /**
- * Module-level cache for primitive (number/string) schema refs.
- * The set of numeric schema sentinels is small and bounded (~20 values),
- * so a regular Map is appropriate.
+ * Cache for string schemaRef.
+ * @internal
  */
-const primitiveSchemaCache = new Map<number | string, NormalizedSchema>();
+export const simpleSchemaCacheS: Record<string, NormalizedSchema> = {};
 
 /**
  * Wraps both class instances, numeric sentinel values, and member schema pairs.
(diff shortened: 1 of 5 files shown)


packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
```diff
@@ -150,6 +157,19 @@ export class NormalizedSchema implements INormalizedSchema {
...
+      const cached = schemaCache.get(sc);
+      if (cached !== undefined) {
+        return cached;
+      }
+      const ns = new NormalizedSchema(sc as $SchemaRef);
+      schemaCache.set(sc, ns);
+      return ns;
+    }
+
+    // Primitive schemas (numbers, strings) bypass cache
```
kuhe: numbers can be assigned objects or use a non-weak map

packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
```diff
@@ -44,6 +44,13 @@ const anno = {
   it: Symbol.for("@smithy/nor-struct-it"),
 };
 
+/**
+ * Module-level cache for NormalizedSchema.of().
+ * Uses WeakMap keyed on object-type schemas (arrays, objects) so entries
+ * are garbage-collected when the schema is no longer referenced.
+ */
+const schemaCache = new WeakMap<object, NormalizedSchema>();
```
kuhe: Can you commit the benchmark code, and flamegraph of time spent in the NormalizedSchema constructor with and without this change?

I remember trying this and discarding it after profiling.

packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
TrevorBurnham: Updated!

packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
TrevorBurnham: I've pushed the benchmark code and some flamegraphs to this branch: https://github.com/TrevorBurnham/smithy-typescript/tree/feat/normalized-schema-caching-benchmarks The flamegraphs look pretty favorable to me; in particular, there's a significant reduction in GC pressure.

kuhe: Once I finish some other work I'll take a look at performance related PRs such as this one and https://github.com/smithy-lang/smithy-typescript/pull/1883.

Can you give me your preferences as to which of your PRs to review first?

kuhe: Also, can you describe your application call-style, if there was one that led to these investigations?

packages/core/src/submodules/schema/schemas/translateTraits.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, test as it } from "vitest";
+
+import { translateTraits } from "./translateTraits";
+
+describe("translateTraits() caching", () => {
+  /**
+   * Validates: Requirements 2.1, 2.2 (Property 3)
```
kuhe: what requirements are being referred to? 

either remove the comments or add the requirements doc

PR #1913, opened 2026-03-08
perf: remove unnecessary input spread and delete in serializeRequest

TrevorBurnham:
## Problem

`HttpBindingProtocol.serializeRequest` creates a shallow copy of the entire input object on every request:

```ts
const input: any = { ...(_input ?? {}) };
```

This O(n) copy exists solely to allow `delete input[memberName]` during HTTP binding extraction (httpPayload, httpLabel, httpHeader, httpPrefixHeaders, httpQuery/httpQueryParams). The intent was to strip HTTP-bound members before passing the remaining input to the body serializer, preventing double-serialization.

`RpcProtocol.serializeRequest` has a similar defensive spread, though it never deletes any members.

## Why the copy and deletes are unnecessary

The body serializer is schema-driven. When `serializer.write(payloadSchema, input)` is called, the codec (e.g. `CborShapeSerializer.serialize`) iterates `ns.structIterator()` with the schema's member list, not the object's own keys:

```ts
} else if (ns.isStructSchema()) {
  for (const [key, memberSchema] of ns.structIterator()) {
    const value = this.serialize(memberSchema, sourceObject[key]);
    // ...
  }
}
```

The `payloadSchema` passed to the serializer is already constructed with only `payloadMemberNames` and `payloadMemberSchemas` — the non-HTTP-binding members. HTTP-bound members are never listed in this schema, so they are never read from the input object regardless of whether they've been deleted.

## Benchmark: `serializeRequest` input spread+delete overhead

Measures the cost of the shallow copy (`{ ...input }`) and `delete` operations
that were previously performed in `HttpBindingProtocol.serializeRequest` and
`RpcProtocol.serializeRequest`.

```bash
cd packages/core
node scripts/serialize-request-perf.mjs
```

500,000 iterations per scenario, 50,000 warmup. 30% of members are HTTP-bound.

| Input size           | Members | HTTP-bound | Old (ms)  | New (ms) | Speedup | Saved/op (µs) |
|----------------------|---------|------------|-----------|----------|---------|----------------|
| small (5 members)    | 5       | 1          | 139.3     | 28.6     | 4.87x   | 0.2            |
| medium (20 members)  | 20      | 6          | 2,357.5   | 104.6    | 22.54x  | 4.5            |
| large (50 members)   | 50      | 15         | 5,648.3   | 271.0    | 20.85x  | 10.8           |
| xlarge (100 members) | 100     | 30         | 14,216.1  | 844.4    | 16.84x  | 26.7           |

- **Old**: shallow copy via spread + delete of HTTP-bound keys (previous code)
- **New**: direct property access, no copy or delete (current code)

---


packages/core/src/submodules/protocols/HttpBindingProtocol.ts
```diff
@@ -35,12 +35,9 @@ export abstract class HttpBindingProtocol extends HttpProtocol {
 
   public async serializeRequest<Input extends object>(
     operationSchema: OperationSchema,
-    _input: Input,
+    input: Input,
     context: HandlerExecutionContext & SerdeFunctions & EndpointBearer
   ): Promise<IHttpRequest> {
-    const input: any = {
-      ...(_input ?? {}),
-    };
```
kuhe, line 43: consider what this transform does other than copy the input

TrevorBurnham: Good catch! Updated.

packages/core/src/submodules/protocols/RpcProtocol.ts
```diff
@@ -57,31 +57,27 @@ export abstract class RpcProtocol extends HttpProtocol {
       this.setHostPrefix(request, operationSchema, input);
     }
 
-    const _input: any = {
-      ...input,
-    };
```
kuhe, line 62: The same applies here. 

I'd like to see tests and benchmarks.

packages/core/src/submodules/protocols/HttpBindingProtocol.ts
```diff
@@ -35,12 +35,10 @@ export abstract class HttpBindingProtocol extends HttpProtocol {
...
+    input = (input ?? {}) as Input;
```
kuhe: this needs to be 

```js
const input: any = _input && typeof _input === "object" ? _input : {};
```

because non object truthy vlues coalesce to themselves

packages/core/src/submodules/protocols/HttpBindingProtocol.spec.ts
```diff
@@ -483,6 +483,194 @@ describe(HttpBindingProtocol.name, () => {
     expect(keys).toEqual(["$metadata"]);
   });
 
+  describe("input not mutated by serializeRequest", () => {
+    it("considers non-object inputs to be equivalent to empty objects", async () => {
```
kuhe, line 487: added this test case

packages/core/src/submodules/protocols/RpcProtocol.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    const request = await protocol.serializeRequest(
+      op("", "", 0, "unit", "unit"),
+      null as any,
+      {
+        endpoint: async () => parseUrl("https://localhost"),
+      } as any
+    );
+
+    expect(request.body).toEqual("{}");
```
kuhe, line 150: This was formerly undefined, I've changed it to `{}`.

This was a core reason of why there was a spread operation.

Consider the concept of RPC and that Smithy inputs are always structural. With no http bindings coming from the input shape's members, the body of an RPC request is always a structure.

trivikr (approved):

PR #1920, opened 2026-03-11
fix: do not return caller's Error directly from buildAbortError

TrevorBurnham:
I received a report that #1886 introduced a breaking change: If the caller passes a frozen or sealed object as the abort reason, the retry middleware throws a `TypeError` when it tries to set `$metadata` on the error:

```
TypeError: Cannot add property $metadata, object is not extensible
```

### Root cause

`buildAbortError` returns the user's `Error` object by reference. The retry middleware (`retryMiddleware.ts`, `StandardRetryStrategy.ts`) assumes it can mutate caught errors to attach `$metadata.attempts` and `$metadata.totalRetryDelay`. Frozen/sealed errors violate that assumption.

Before #1886, abort errors were always freshly constructed (`new Error("Request aborted")`), so this was never an issue.

### Fix

`buildAbortError` now always creates a new `Error` with the standard `name: "AbortError"` and `message: "Request aborted"`. When `reason` is an `Error`, the original is preserved via the standard `.cause` property.

This restores the pre-#1886 contract that abort errors are always fresh, mutable `Error` instances with a consistent shape, while still making the caller's reason accessible through `.cause`.

Both copies of `buildAbortError` are updated:
- `packages/node-http-handler/src/build-abort-error.ts`
- `packages/fetch-http-handler/src/fetch-http-handler.ts`

---

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -227,7 +227,7 @@ function buildAbortError(abortSignal?: unknown): Error {
     if (reason instanceof Error) {
       const abortError = new Error("Request aborted");
       abortError.name = "AbortError";
-      abortError.cause = reason;
+      (abortError as { cause?: unknown }).cause = reason;
       return abortError;
     }
     const abortError = new Error(String(reason));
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "downlevelIteration": true,
     "esModuleInterop": true,
     "incremental": true,
-    "lib": ["es2015", "dom"],
+    "lib": ["es2023", "dom"],
```
kuhe: this is compatible with node 18/20

kuhe: nevermind

PR #1922, opened 2026-03-16
chore(util-retry): specification updates for retry 2.1

kuhe:
*Issue #, if available:*
pending

*Description of changes:*

Adds new default retry behaviors. In summary:

- new behavior is gated being a process.env variable to be defaulted to true later (and removed)
	- slightly less time between default retries
	- slightly more time between default retries when throttled by server
	- lower retry ticket consumption for throttling errors
	- better parsing for `retry-after` and `x-amz-retry-after` headers

### work checklist
- [x] customize attempt count for special services (2 in AWS)
  - This is a code generation change for those services in which the generated runtimeConfig init object will include a StandardRetryStrategy initialized with a different maxAttempts count.
- [x] customize long polling behavior for certain operations (4 in AWS)
  - This is also a code generation change in which the specific operations in question will make use of the long poll middleware from this PR.
- [x] integration test 
  - [x] jitter normalized
  - [x] env var on/off

### DIFF
--- a/.changeset/proud-pumpkins-relax.md
+++ b/.changeset/proud-pumpkins-relax.md
@@ -4,4 +4,5 @@
 "@smithy/types": minor
 ---
 
-temporary configuration for new 2026 retry behaviors
+Introduce default retry behavior modifications slated for 2026. They are:
+less time between server error retries, but slightly more time between throttling errors. Lower retry capacity consumption for throttling, and improved parsing of the retry-after and x-amz-retry-after headers.
--- a/packages/snapshot-testing/integ-snapshots/res-err/ComplexError.txt
+++ b/packages/snapshot-testing/integ-snapshots/res-err/ComplexError.txt
@@ -44,13 +44,18 @@ content-type: application/cbor
 
 [Uint8Array (cbor object view)]
 {
+  "TopLevel": "__TopLevel__",
+  "Nested": {
+    "Foo": "__Foo__"
+  },
   "__type": "smithy.protocoltests.rpcv2Cbor#ComplexError"
 }
 
 [actual bytes]
-161, 102, 95, 95, 116, 121, 112, 101, 120, 43, 115, 109, 105, 116, 104, 121, 46, 112, 114, 111, 116, 111, 99, 111,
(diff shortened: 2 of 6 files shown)


packages/middleware-retry/src/parseRetryAfterHeader.ts
```diff
@@ -0,0 +1,61 @@
...
+        retryAfterSeconds = Number(retryAfter.match(/ GMT, ([\d.]+)$/)?.[1]);
+      } else if (retryAfter.match(/^((\d+)|(\d+\.\d+))$/)) {
+        retryAfterSeconds = Number(retryAfter);
+      } else if (Date.parse(retryAfter) >= Date.now()) {
+        // non-standard header value, attempt to parse as date.
+        retryAfterSeconds = (Date.parse(retryAfter) - Date.now()) / 1000;
+      }
+
+      if (isNaN(retryAfterSeconds)) {
+        return undefined;
```
kuhe:
```suggestion
        return;
```

.changeset/proud-pumpkins-relax.md
```diff
@@ -0,0 +1,7 @@
+---
+"@smithy/middleware-retry": minor
+"@smithy/util-retry": minor
+"@smithy/types": minor
+---
+
+temporary configuration for new 2026 retry behaviors
```
kuhe:
```suggestion
Introduce default retry behavior modifications slated for 2026. They are:
less time between server error retries, but slightly more time between throttling errors. Lower retry capacity consumption for throttling, and improved parsing of the retry-after and x-amz-retry-after headers.
```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,38 +1,67 @@
...
+import { DefaultRetryBackoffStrategy } from "./DefaultRetryBackoffStrategy";
+import { DefaultRetryToken } from "./DefaultRetryToken";
+import { Retry } from "./retries-2026-config";
+
+/**
+ * @public
+ */
+export type StandardRetryStrategyOptions = {
+  /**
+   * Maximum number of attempts. If set to 1, no retries will be made.
```
siddsriv, line 21: for 0? is that validated?

kuhe: having n <= 1 only prevents a retry from happening, and does not prevent the initial request

PR #1950, opened 2026-04-09
fix(util-retry): restore attempt count private APIs

kuhe:
*Issue #, if available:*
n/a

*Description of changes:*

This establishes backwards compatible public APIs for reporting attempt counts in retry strategies due to an integration point that accessed the private properties at runtime.

### DIFF
--- a/packages/util-retry/src/AdaptiveRetryStrategy.ts
+++ b/packages/util-retry/src/AdaptiveRetryStrategy.ts
@@ -62,4 +62,12 @@ export class AdaptiveRetryStrategy implements RetryStrategyV2 {
     this.rateLimiter.updateClientSendingRate({});
     this.standardRetryStrategy.recordSuccess(token);
   }
+
+  /**
+   * There is an existing integration which accesses this field.
+   * @deprecated
+   */
+  public async maxAttemptsProvider(): Promise<number> {
+    return this.standardRetryStrategy.maxAttempts();
+  }
 }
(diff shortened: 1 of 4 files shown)


trivikr (approved):

packages/util-retry/src/AdaptiveRetryStrategy.ts
```diff
@@ -62,4 +62,12 @@ export class AdaptiveRetryStrategy implements RetryStrategyV2 {
...
     this.standardRetryStrategy.recordSuccess(token);
   }
+
+  /**
+   * There is an existing integration which accesses this field.
+   * @deprecated
+   */
+  public async maxAttemptsProvider(): Promise<number> {
+    return this.standardRetryStrategy.maxAttempts();
+  }
```
trivikr, line 72: Should this be `maxAttempts`?
```suggestion
  public async maxAttempts(): Promise<number> {
    return this.standardRetryStrategy.maxAttempts();
  }
```

trivikr (changes requested):

packages/util-retry/src/AdaptiveRetryStrategy.ts
trivikr: Nope. The existing integration is for `maxAttemptsProvider` in `AdaptiveRetryStrategy`

trivikr (approved):

PR #1957, opened 2026-04-14
perf(util-endpoints): reduce temporary object allocations

trivikr:
### Issue
Internal JS-6645

### Description

Remove rest spread destructuring that created a new object on every call, and replace conditional spread in the return with an early return branch to avoid temporary object allocation.

---

### DIFF
--- a/packages/util-endpoints/src/types/shared.ts
+++ b/packages/util-endpoints/src/types/shared.ts
@@ -1,4 +1,4 @@
-import type { EndpointARN, EndpointPartition, Logger } from "@smithy/types";
+import type { EndpointARN, EndpointPartition, EndpointURL, Logger } from "@smithy/types";
 
 export type ReferenceObject = { ref: string };
 
@@ -10,8 +10,11 @@ export type FunctionReturn =
   | number
   | EndpointARN
   | EndpointPartition
+  | EndpointURL
   | { [key: string]: FunctionReturn }
-  | null;
+  | Array<FunctionReturn>
+  | null
+  | undefined;
 
 export type ConditionObject = FunctionObject & { assign?: string };
 
--- a/packages/util-endpoints/src/types/shared.ts
+++ b/packages/util-endpoints/src/types/shared.ts
@@ -13,8 +13,7 @@ export type FunctionReturn =
   | EndpointURL
   | { [key: string]: FunctionReturn }
   | Array<FunctionReturn>
-  | null
-  | undefined;
+  | null;
 
 export type ConditionObject = FunctionObject & { assign?: string };
 
--- a/packages/util-endpoints/src/utils/evaluateCondition.ts
+++ b/packages/util-endpoints/src/utils/evaluateCondition.ts
@@ -3,16 +3,20 @@ import type { ConditionObject, EvaluateOptions } from "../types";
 import { EndpointError } from "../types";
 import { callFunction } from "./callFunction";
 
-export const evaluateCondition = ({ assign, ...fnArgs }: ConditionObject, options: EvaluateOptions) => {
+export const evaluateCondition = (condition: ConditionObject, options: EvaluateOptions) => {
+  const { assign } = condition;
+
   if (assign && assign in options.referenceRecord) {
     throw new EndpointError(`'${assign}' is already defined in Reference Record.`);
   }
-  const value = callFunction(fnArgs, options);
+  const value = callFunction(condition, options);
 
-  options.logger?.debug?.(`${debugId} evaluateCondition: ${toDebugString(fnArgs)} = ${toDebugString(value)}`);
+  options.logger?.debug?.(`${debugId} evaluateCondition: ${toDebugString(condition)} = ${toDebugString(value)}`);
 
-  return {
-    result: value === "" ? true : !!value,
-    ...(assign != null && { toAssign: { name: assign, value } }),
-  };
+  const result = value === "" ? true : !!value;
+
+  if (assign != null) {
+    return { result, toAssign: { name: assign, value } };
+  }
+  return { result };
 };
--- a/packages/util-endpoints/src/utils/evaluateConditions.ts
+++ b/packages/util-endpoints/src/utils/evaluateConditions.ts
@@ -23,5 +23,9 @@ export const evaluateConditions = (conditions: ConditionObject[] = [], options: 
     }
   }
 
+  if (Object.keys(conditionsReferenceRecord).length === 0) {
+    return { result: true };
+  }
+
   return { result: true, referenceRecord: conditionsReferenceRecord };
 };
--- a/packages/util-endpoints/src/utils/evaluateConditions.ts
+++ b/packages/util-endpoints/src/utils/evaluateConditions.ts
@@ -8,6 +8,7 @@ export const evaluateConditions = (conditions: ConditionObject[] = [], options: 
     ...options,
     referenceRecord: { ...options.referenceRecord },
   };
+  let didAssign = false;
 
   for (const condition of conditions) {
     const { result, toAssign } = evaluateCondition(condition, conditionOptions);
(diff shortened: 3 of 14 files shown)


kuhe (approved):

packages/util-endpoints/src/utils/evaluateConditions.ts
```diff
@@ -4,22 +4,21 @@ import { evaluateCondition } from "./evaluateCondition";
...
+  };
 
   for (const condition of conditions) {
-    const { result, toAssign } = evaluateCondition(condition, {
-      ...options,
-      referenceRecord: {
-        ...options.referenceRecord,
-        ...conditionsReferenceRecord,
-      },
-    });
```
kuhe, line 15: how can we confirm no edge case relied on this? the diff seems to result in a potentially different set of objects

if it can be proven this is all part of the same evaluation's lifecycle, maybe even the initial copy isn't necessary.

kuhe: one thing we could do is run the full AWS endpoint test suite

trivikr: We're not changing the behavior. The `conditionOptions` object was additive even before. We're just not creating new object for every call.

The test needs to check the first call received an empty referenceRecord at call time, but the spy only holds a reference that's since been mutated. That's why tests had to be updated.

packages/util-endpoints/src/utils/evaluateConditions.ts
```diff
@@ -4,25 +4,28 @@ import { evaluateCondition } from "./evaluateCondition";
 
 export const evaluateConditions = (conditions: ConditionObject[] = [], options: EvaluateOptions) => {
   const conditionsReferenceRecord: Record<string, FunctionReturn> = {};
+  const conditionOptions: EvaluateOptions = {
+    ...options,
+    referenceRecord: { ...options.referenceRecord },
+  };
```
kuhe, line 10: is it safe to use `options` all the way through with no copying?

Why I think it might be safe: options object lifecycle is one endpoint resolution, and this is all synchronous code...

packages/util-endpoints/src/utils/evaluateConditions.ts
trivikr: Verified that endpoint tests are successful

```console
$ aws-sdk-js-v3> make test-endpoints
...
 Test Files  1 passed (1)
      Tests  13946 passed | 5 skipped (13951)
   Start at  16:59:34
   Duration  8.39s (transform 38ms, setup 0ms, import 5.43s, tests 2.63s, environment 0ms)
```

They were run by copying `packages/util-endpoints/dist-cjs/index.js` from `smithy-typescript` to `node_modules/@smithy/util-endpoints/dist-cjs/index.js` in `aws-sdk-js-v3`

packages/util-endpoints/src/utils/evaluateConditions.ts
trivikr: I'll avoid modifying options, as it might create issues in future. It can be tested in a separate PR.

packages/util-endpoints/src/types/shared.ts
```diff
@@ -10,8 +10,11 @@ export type FunctionReturn =
   | number
   | EndpointARN
   | EndpointPartition
+  | EndpointURL
   | { [key: string]: FunctionReturn }
-  | null;
+  | Array<FunctionReturn>
+  | null
+  | undefined;
```
kuhe: should the functions instead be corrected to return null, instead of expanding the type to include undefined?

packages/util-endpoints/src/utils/evaluateCondition.ts
```diff
@@ -3,16 +3,20 @@ import type { ConditionObject, EvaluateOptions } from "../types";
...
 
-export const evaluateCondition = ({ assign, ...fnArgs }: ConditionObject, options: EvaluateOptions) => {
+export const evaluateCondition = (condition: ConditionObject, options: EvaluateOptions) => {
+  const { assign } = condition;
+
   if (assign && assign in options.referenceRecord) {
     throw new EndpointError(`'${assign}' is already defined in Reference Record.`);
   }
-  const value = callFunction(fnArgs, options);
+  const value = callFunction(condition, options);
```
kuhe, line 12: now condition has "assign" in it, compared to before. That's not a problem, right?

packages/util-endpoints/src/utils/evaluateConditions.ts
```diff
@@ -4,25 +4,28 @@ import { evaluateCondition } from "./evaluateCondition";
...
     }
 
     if (toAssign) {
       conditionsReferenceRecord[toAssign.name] = toAssign.value;
+      conditionOptions.referenceRecord[toAssign.name] = toAssign.value;
       options.logger?.debug?.(`${debugId} assign: ${toAssign.name} := ${toDebugString(toAssign.value)}`);
     }
   }
 
+  if (Object.keys(conditionsReferenceRecord).length === 0) {
```
kuhe: object keys theoretically performs a full object iteration and allocates an array. 

this could be avoided by flipping a boolean `didAssign = true` when assigning to `conditionsReferenceRecord`

not a blocker for this PR

packages/util-endpoints/src/utils/evaluateEndpointRule.ts
```diff
@@ -18,22 +18,24 @@ export const evaluateEndpointRule = (
...
-    ...(headers != undefined && {
-      headers: getEndpointHeaders(headers, endpointRuleOptions),
-    }),
-    ...(properties != undefined && {
-      properties: getEndpointProperties(properties, endpointRuleOptions),
-    }),
-    url: getEndpointUrl(url, endpointRuleOptions),
-  };
+  const endpointToReturn: EndpointV2 = { url: getEndpointUrl(url, endpointRuleOptions) };
+  if (headers != undefined) {
```
kuhe: we should use `null` when comparing with `==` an `!=`, since I think that's more common in the codebase overall

packages/util-endpoints/src/utils/evaluateEndpointRule.ts
```diff
@@ -18,22 +18,24 @@ export const evaluateEndpointRule = (
...
+    endpointToReturn.headers = getEndpointHeaders(headers, endpointRuleOptions);
+  }
+  if (properties != undefined) {
+    endpointToReturn.properties = getEndpointProperties(properties, endpointRuleOptions);
+  }
+
+  return endpointToReturn;
```
kuhe, line 40: unlikely, but does the call order matter for exception handling?

the old code called getEndpointUrl last.

packages/util-endpoints/src/utils/evaluateRules.ts
```diff
@@ -35,10 +35,11 @@ export const evaluateTreeRule = (treeRule: TreeRuleObject, options: EvaluateOpti
     return;
   }
 
-  return group.evaluateRules(rules, {
-    ...options,
-    referenceRecord: { ...options.referenceRecord, ...referenceRecord },
-  });
+  const treeRuleOptions = referenceRecord
```
kuhe, line 38: the more code I (re-)read, the more I think `options` is safe to mutate

kuhe (approved):

packages/util-endpoints/src/utils/evaluateCondition.ts
trivikr: No. The `callFunction` doesn't destructure `assign`

https://github.com/smithy-lang/smithy-typescript/blob/af3363fd5e03e63e6b737ff9a361454de79d2e3c/packages/util-endpoints/src/utils/evaluateExpression.ts#L19

packages/util-endpoints/src/utils/evaluateEndpointRule.ts
trivikr: Actually, the previous code would have thrown exceptions from getEndpointHeaders and getEndpointProperties before those from getEndpointUrl.

It should be okay though.

packages/util-endpoints/src/utils/evaluateConditions.ts
trivikr: Fixed in https://github.com/smithy-lang/smithy-typescript/pull/1957/changes/9a97c4980922c87a8fefc61e1fdae6037ab7641a

packages/util-endpoints/src/types/shared.ts
trivikr: Good catch! No function was returning undefined, so it was removed in https://github.com/smithy-lang/smithy-typescript/pull/1957/changes/aa81904cf899b3de1fe6362142fcae41b90a5d06

packages/util-endpoints/src/utils/evaluateEndpointRule.ts
trivikr: Done in https://github.com/smithy-lang/smithy-typescript/pull/1957/changes/4d4929225e186686049adc5855035159c221a415

PR #1961, opened 2026-04-16
perf(core): reduce intermediate allocations in hot paths

trivikr:
*Issue #, if available:*
Internal JS-6645

*Description of changes:*
Reduces intermediate allocations in hot paths

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1963, opened 2026-04-17
chore(core): replace Object.entries with single-pass header scan

trivikr:
*Issue #, if available:*
Internal JS-6645

*Description of changes:*
* Replaces Object.entries with single-pass header scan
* This was missed in https://github.com/smithy-lang/smithy-typescript/pull/1961 as it wasn't in hot path

---


trivikr:
Draft PR was posted for reference. It's not worth as it's not on hot path.

PR #1965, opened 2026-04-21
fix(service-error-classification): add check for clock skew based errors to retry

siddsriv:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/7605

*Description of changes:*
The SDK was not attempting retries upon receiving `InvalidSignatureException` with "Signature expired" (due to clock skew from Lambda freeze/thaw or ECS clock drift). The errorHandler ([code](https://github.com/aws/aws-sdk-js-v3/blob/f83dd4822559eca40d1fcd13bf330259c588bc28/packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts#L127)) attempts to detect clock skew by comparing server time to `Date.now()`. In freeze/thaw scenarios, the system clock could have already self-corrected by the time the error response arrives, so ultimately the RetryStrategy in the SDK classifies the error as a non-retryable error.

This change intends to make the error retryable since a subsequent request in this case can succeed.

---

---

### DIFF
--- a/packages/service-error-classification/src/index.ts
+++ b/packages/service-error-classification/src/index.ts
@@ -57,6 +57,7 @@ export const isThrottlingError = (error: SdkError) =>
 export const isTransientError = (error: SdkError, depth = 0): boolean =>
   isRetryableByTrait(error) ||
   isClockSkewCorrectedError(error) ||
+  (error.name === "InvalidSignatureException" && error.message?.includes("Signature expired")) ||
   TRANSIENT_ERROR_CODES.includes(error.name) ||
   NODEJS_TIMEOUT_ERROR_CODES.includes((error as { code?: string })?.code || "") ||
   NODEJS_NETWORK_ERROR_CODES.includes((error as { code?: string })?.code || "") ||
--- a/.changeset/neat-ravens-clean.md
+++ b/.changeset/neat-ravens-clean.md
@@ -1,0 +1,5 @@
+---
+"@smithy/service-error-classification": patch
+---
+
+retry clock skew errors for freeze/thaw/drift situations


packages/service-error-classification/src/index.ts
```diff
@@ -57,6 +57,7 @@ export const isThrottlingError = (error: SdkError) =>
 export const isTransientError = (error: SdkError, depth = 0): boolean =>
   isRetryableByTrait(error) ||
   isClockSkewCorrectedError(error) ||
+  (error.name === "InvalidSignatureException" && error.message?.includes("Signature expired")) ||
```
trivikr, line 60: Blocking as this will attempt retries for genuine InvalidSignatureException

trivikr (changes requested):

packages/service-error-classification/src/index.ts
siddsriv: what is the genuine `InvalidSignatureException` in this case? Anything other than a clock skew issue (like bad creds?) I tried to narrow it down, we predicate on both the name and message and that should be a clock skew issue with high confidence.

trivikr: On 4/22, we discussed that this fix will not impact the genuine InvalidSignatureException.
However, we're discussing a solution with rest of the SDKs before implementing a fix.

trivikr: On 4/23, we decided to merge this fix as we have check for "Signature expired" in message contents in addition to error name being "InvalidSignatureException".
It can be replaced with a long term fix when it's decided with the rest of the SDKs.

trivikr (approved):

PR #1970, opened 2026-04-24
feat(util-waiter): emit warning from waiter for sustained 403s

kuhe:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/3081
https://github.com/aws/aws-sdk-js-v3/issues/7818

*Description of changes:*

(waitUntil) Waiters, which throw on any non-success result, currently include the encountered errors in the thrown Error message.

For example:
```
Error [TimeoutError]: {"state":"TIMEOUT","observedResponses":{"403: UnknownError":3},"reason":"Waiter has timed out"}
```

This PR addresses long duration waiters, because it may take too long for them to surface the final thrown error to the user.

If 1 minute elapses and there are at least 3 `403` errors, or the latest error is 403, a warning will be pushed to a logger. The logger is selected from the client performing the polling or console.

This warning does not terminate the waiter. The warning will only be emitted once per waiter. The warning text is:

```
@smithy/util-waiter WARN - 403 status code encountered during waiter polling.
```

### DIFF
--- a/packages/util-waiter/src/poller.ts
+++ b/packages/util-waiter/src/poller.ts
@@ -4,20 +4,9 @@ import type { WaiterOptions, WaiterResult } from "./waiter";
 import { WaiterState } from "./waiter";
 
 /**
- * Reference: https://smithy.io/2.0/additional-specs/waiters.html#waiter-retries
- * @internal
- */
-const exponentialBackoffWithJitter = (minDelay: number, maxDelay: number, attemptCeiling: number, attempt: number) => {
-  if (attempt > attemptCeiling) return maxDelay;
-  const delay = minDelay * 2 ** (attempt - 1);
-  return randomInRange(minDelay, delay);
-};
-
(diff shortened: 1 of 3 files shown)


packages/util-waiter/src/poller.ts
```diff
@@ -83,19 +105,46 @@ export const runPolling = async <Client, Input, Reason = any>(
...
+    return maxDelayMs;
+  }
+
+  const delay = minDelayMs * 2 ** (attempt - 1);
+  const capped = Math.min(delay, maxDelayMs);
+  const waitFor = randomInRange(minDelayMs, capped);
+
+  if (Date.now() + waitFor > waitUntil) {
+    const timeRemaining = waitUntil - Date.now();
+    // fire the last request 500ms before the waiter would time out.
```
kuhe, line 141: we were missing the expedited last poll from the spec: https://smithy.io/2.0/additional-specs/waiters.html#waiter-retries

trivikr (approved):

PR #1974, opened 2026-04-27
fix(util-retry): missing call to getSendToken in Adaptive mode retries

kuhe:
*Issue #, if available:*
n/a

*Description of changes:*

adds missing sendToken acquisition in adaptive mode retry path

### DIFF
--- a/packages/util-retry/src/AdaptiveRetryStrategy.ts
+++ b/packages/util-retry/src/AdaptiveRetryStrategy.ts
@@ -46,8 +46,9 @@ export class AdaptiveRetryStrategy implements RetryStrategyV2 {
   }
 
   public async acquireInitialRetryToken(retryTokenScope: string): Promise<RetryToken> {
+    const token = this.standardRetryStrategy.acquireInitialRetryToken(retryTokenScope);
     await this.rateLimiter.getSendToken();
-    return this.standardRetryStrategy.acquireInitialRetryToken(retryTokenScope);
+    return token;
   }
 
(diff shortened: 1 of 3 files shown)


packages/util-retry/src/AdaptiveRetryStrategy.ts
```diff
@@ -54,6 +54,7 @@ export class AdaptiveRetryStrategy implements RetryStrategyV2 {
     tokenToRenew: StandardRetryToken,
     errorInfo: RetryErrorInfo
   ): Promise<RetryToken> {
+    await this.rateLimiter.getSendToken();
     this.rateLimiter.updateClientSendingRate(errorInfo);
     return this.standardRetryStrategy.refreshRetryTokenForRetry(tokenToRenew, errorInfo);
```
millems: I think getSendToken should be after updateClientSendingRate (the response will dictate when to retry) and after we do the standard retry mode checks and determine we will indeed retry, otherwise we may be waiting for a chance to send a message that never gets sent.

Something equivalent to this? (Not a TS expert, so this could be wrong)
```ts
    this.rateLimiter.updateClientSendingRate(errorInfo);
    const token = await this.standardRetryStrategy.refreshRetryTokenForRetry(tokenToRenew, errorInfo);
    await this.rateLimiter.getSendToken(); // Skipped if refreshRetryTokenForRetry throws
    return token;
```

siddsriv (approved):

PR #1984, opened 2026-04-29
perf(node-http-handler): skip await on settled writeRequestBodyPromise

trivikr:
*Issue #, if available:*

*Description of changes:*

Track writeRequestBodyPromise settlement with a boolean flag and only await when the promise is still pending.

In the common case where the request body has already been written before the response callback fires, this avoids scheduling an unnecessary microtask on every resolve and reject call.

---


trivikr:
Cancelled as performance improvement is much lesser than impact it has on readability.

Issue #1980, opened 2026-04-29
Core package consolidation


kuhe:
We are planning to consolidate the various NPM runtime packages into a smaller set within `@smithy/core`, using the submodule system enabled by the package.json `exports` field.

Reducing the number of packages should lower the chance of entering an invalid package-lock state, and simplify the patching process. It may also make things simpler for bundling and other build operations against generated clients.

See https://github.com/smithy-lang/smithy-typescript/blob/main/packages/core/planning/consolidation.md

PoC:
https://github.com/smithy-lang/smithy-typescript/pull/1978

PRs (list may be incomplete):
#1977
#1979
https://github.com/smithy-lang/smithy-typescript/pull/1981

core/serde https://github.com/smithy-lang/smithy-typescript/pull/1985
core/client https://github.com/smithy-lang/smithy-typescript/pull/1991
core/config https://github.com/smithy-lang/smithy-typescript/pull/1992
core/protocols https://github.com/smithy-lang/smithy-typescript/pull/1998
core/endpoints https://github.com/smithy-lang/smithy-typescript/pull/2001
core/event-streams https://github.com/smithy-lang/smithy-typescript/pull/2003

PR #1985, opened 2026-04-29
chore(core/serde): consolidate serde packages

kuhe:
https://github.com/smithy-lang/smithy-typescript/issues/1980

💀 I tried to make a small PR but there were too many dependency cycles.

sample client diff: https://github.com/aws/aws-sdk-js-v3/pull/7979


api-snapshot/api.json
```diff
@@ -4,12 +4,6 @@
     "AbortController": "function, since <=4.2.5",
     "AbortSignal": "function, since <=4.2.5"
   },
-  "@smithy/chunked-blob-reader": {
-    "blobReader": "function, since <=5.2.0"
-  },
-  "@smithy/chunked-blob-reader-native": {
-    "blobReader": "function, since <=4.2.1"
-  },
```
kuhe, line 12: Why deleting the migrated packages is good:

- prevents code-drift between the migrated and original source code files
- preserved git history on the new file location
- prevents accidental publishing of a new version of a deprecated package
  - the original package will not be deprecated on NPM. Clients use it.

package tombstone exists in the form of the banned import list in eslint

packages/core/src/submodules/serde/index.ts
```diff
@@ -1,10 +1,110 @@
-export * from "./copyDocumentWithTransform";
```
kuhe, line 1: why `export *` is dangerous:

- because it isn't explicit, we may end up removing or adding things to the exported API surface unintentionally. It has happened before. I propose that we ban the use of `export *`.

packages/core/package.json
```diff
@@ -141,5 +136,50 @@
   },
   "publishConfig": {
     "directory": ".release/package"
+  },
+  "browser": {
```
kuhe, line 140: in a future PR I plan to further consolidate the browser variants so that fewer variant files are required

api-snapshot/api.json
kuhe: review without deleted files:

https://github.com/smithy-lang/smithy-typescript/pull/1985/changes?show-deleted-files=false

kuhe: changed direction on this: I'm making the migrated packages stubs, so when older clients upgrade core, they don't receive a redundant set of code from two packages.

trivikr (approved):

PR #1986, opened 2026-04-29
perf(node-http-handler): cache 100-continue agents at module level

trivikr:
*Issue #, if available:*

*Description of changes:*

Replaces per-request Agent creation for 100-continue requests with lazily initialized module-level singletons.

Workloads that send many 100-continue requests (e.g., S3 PutObject) were creating a new Agent per request, causing significant GC pressure. A single cached agent per protocol (HTTP/HTTPS) is now reused across all handler instances.

---

### DIFF
--- a/.changeset/gentle-brooms-compete.md
+++ b/.changeset/gentle-brooms-compete.md
@@ -1,0 +1,5 @@
+---
+"@smithy/node-http-handler": patch
+---
+
+Cache 100-continue agents at module level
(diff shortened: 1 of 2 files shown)


kuhe (changes requested):
reusing an agent could resurface the bug that originally necessitated creating a new agent.

We should prove that the original issue (internal repro is available) is fixed before merging this change. 

If it is fixed, creating a separate agent may be unnecessary to begin with.

In any case this is partially mitigated by the byte length check before enabling 100-continue in S3.

PR #1987, opened 2026-04-29
chore(node-http-handler): replace timeouts array with individual variables

trivikr:
*Issue #, if available:*

*Description of changes:*

Avoid per-request array allocation, push calls, and forEach iteration on the hot path by using fixed individual timeout variables and a clearTimeouts helper.

The number of timeouts is bounded (max 5), so individual variables are a better fit than a dynamic array.

---


kuhe (approved):
approved for readability, but I don't think one array alloc should matter vs what is now a function + closure alloc

Issue #1993, opened 2026-05-02
AdaptiveRetryStrategy fails with unhandledRejection in node


diego-santacruz:
Following commit 2e5142c the `AdaptiveRetryStrategy` fails with an `unhandledRejection` and terminates the process when the client request fails with a non-retriable error (e.g., `ResourceNotFoundException`) but `getSendToken()` sleeps.

The problem is both in `refreshRetryTokenForRetry()` and in `acquireInitialRetryToken()`, although much more likely to occur in `refreshRetryTokenForRetry()`.

In
```ts
  public async refreshRetryTokenForRetry(
    tokenToRenew: StandardRetryToken,
    errorInfo: RetryErrorInfo
  ): Promise<RetryToken> {
    this.rateLimiter.updateClientSendingRate(errorInfo);
    const token = this.standardRetryStrategy.refreshRetryTokenForRetry(tokenToRenew, errorInfo);
    // called prior to return in case the token refresh throws (no need to wait in that case).
    await this.rateLimiter.getSendToken();
    return token;
  }
```

The call to `this.standardRetryStrategy.refreshRetryTokenForRetry()` returns a pending promise, stored in `token`, which is not yet awaited. If the call to `await this.rateLimiter.getSendToken();` sleeps (with `setTimeout()`) then execution of `this.standardRetryStrategy.refreshRetryTokenForRetry()` will resume and if it throws before `await this.rateLimiter.getSendToken();` fulfills then node.js treats that as an unhandled rejection by the runtime (https://nodejs.org/api/process.html#event-unhandledrejection). That is because no handler is attached at the time it throws. The unhandled rejection will not be caught by any enclosing `try / catch` blocks, but will instead terminate the runtime execution.

The `refreshRetryTokenForRetry()` method of `StandardRetryStrategy` will throw a `No retry token available` error if the request failure is not one that should be retried (e.g., `ResourceNotFoundException`).

I encountered this issue when using `AdaptiveRetryStrategy` with a rate limiter with very low rate (10 per second) when making calls wtih the AWS SDK that return `ResourceNotFoundException` (IoT Core `ListThingPrincipals` API).

The same problem exists in `acquireInitialRetryToken()` because it uses the exact same pattern.

The fix is to attach a handler with `.catch()` before `await this.rateLimiter.getSendToken();`. I see two possible variants.

[...]

PR #1994, opened 2026-05-02
fix(util-retry): add missing await to async calls in Adaptive Retry

kuhe:
*Issue #, if available:*

https://github.com/smithy-lang/smithy-typescript/issues/1993

*Description of changes:*

Adds missing `await` to AdaptiveRetryStrategy.


.changeset/wet-pianos-scream.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
kuhe, line 2: empty changeset because this patch was published directly

smilkuri (approved):

PR #2004, opened 2026-05-05
chore(scripts): add type symbols to api snapshot

kuhe:
*Issue #, if available:*

https://github.com/smithy-lang/smithy-typescript/issues/1980

*Description of changes:*

This PR updates the api snapshot script to also use TS to snapshot the exported type symbols. As a consequence of this, the package stubs that were consolidated were patched to export the same types as the baseline commit 42631157aca4f3bd7313df053d58b58f3f4afd1c.


api-snapshot/api.json
```diff
@@ -1,830 +1,1454 @@
 {
```
kuhe, line 1: lots of thrash in this file, but things should improve going forward:

ordering is now alphabetical to assist visual diffing rather than in the export order of the package.

kuhe: the 'since version' wasn't helping much so I removed it. It tended to be the wrong version anyway, because not all PRs perform a full build and run the api snapshot.

.github/workflows/ci.yml
```diff
@@ -117,8 +117,8 @@ jobs:
         run: |
           yarn
           yarn turbo telemetry disable
-      - name: Build packages
-        run: node ./scripts/retry -- yarn build
+      - name: Build and validate API snapshot
+        run: node ./scripts/retry -- make api-snapshot
```
kuhe, line 121: CI will now verify api snapshot has no diff after a clean build before unit testing

siddsriv (approved):

PR #2027, opened 2026-05-14
fix(types): exclude SharedArrayBuffer from recursive type transform

smilkuri:
*Issue #, if available:*
Internal P428473469

*Description of changes:*
Exclude SharedArrayBuffer from recursive type transformation to prevent circular type references caused by its self-referential [Symbol.species] property.

---

### DIFF
--- a/packages/types/src/transform/type-transform.spec.ts
+++ b/packages/types/src/transform/type-transform.spec.ts
@@ -21,6 +21,21 @@ type A = {
   const assert3: Exact<T["nested"]["nested"]["nested"]["b"], "enum"> = true as const;
 }
 
+// It should not recurse into SharedArrayBuffer.
+type B = {
+  typed: SharedArrayBuffer;
+  untyped: {
+    byteLength: number;
+  };
+};
+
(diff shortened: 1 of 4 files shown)


kuhe (approved):

packages/types/src/transform/type-transform.spec.ts
```diff
@@ -21,6 +21,19 @@ type A = {
...
 }
 
+{
+  // It should not recurse into SharedArrayBuffer.
+  type B = {
+    data: SharedArrayBuffer;
+    name: number | string;
+  };
+
+  type T = Transform<B, number | string, "enum">;
```
kuhe: also try transforming SharedArrayBuffer into something else and confirm that it doesn't?

smilkuri: the check we added only prevents being recursed into sharedArrayBuffer right doesn't prevent from transforming if we specify a target.

kuhe: you can attempt to transform a sub-property type of SharedArrayBuffer and then assert that it does not change

PR #2029, opened 2026-05-14
feat(undici-http-handler): add HttpHandler backed by Node.js undici

trivikr:
*Issue #, if available:*

Migrating code from test package in https://github.com/trivikr/trivikr-test-undici-http-handler

*Description of changes:*

Adds a new `@smithy/undici-http-handler` package that provides a Smithy-compatible HTTP handler backed by [undici](https://undici.nodejs.org/) instead of Node.js native http/https modules.

undici offers improved HTTP performance over Node.js built-in modules. Benchmarks show **20%-30%** less time spent in request handling compared to `NodeHttpHandler` from `@smithy/node-http-handler`, particularly under concurrent load.

Benchmarks (in several runs, it's close to 1.20x and 1.30x)
```console
$ undici-http-handler> yarn test:bench
...
 ✓ src/undici-http-handler.bench.ts > 10 sequential GETs 1213ms
     name                   hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    413.26  2.0419  5.2733  2.4198  2.3764  3.9081  5.0611  5.2733  ±2.49%      207
   · UndiciHttpHandler  476.21  1.8400  4.0838  2.0999  2.1102  2.9196  3.0122  4.0838  ±1.55%      239

 ✓ src/undici-http-handler.bench.ts > 50 concurrent GETs 1267ms
     name                   hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    208.81  4.2706  7.0158  4.7890  4.8566  6.7310  7.0158  7.0158  ±2.30%      105
   · UndiciHttpHandler  274.86  3.3944  4.3799  3.6383  3.6770  4.2547  4.3799  4.3799  ±0.87%      138

 BENCH  Summary

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 10 sequential GETs
    1.15x faster than NodeHttpHandler

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 50 concurrent GETs
    1.32x faster than NodeHttpHandler
```

The E2E tests were run on test handler in https://github.com/aws/aws-sdk-js-v3/pull/8014
New E2E tests will be added when `@smithy/undici-http-handler` is published.

---


packages/undici-http-handler/src/undici-http-handler.spec.ts
```diff
@@ -0,0 +1,570 @@
...
+beforeAll(async () => {
+  server = createServer((req: IncomingMessage, res: ServerResponse) => {
+    const url = new URL(req.url!, `http://localhost`);
+
+    if (url.pathname === "/delay") {
+      const ms = parseInt(url.searchParams.get("ms") ?? "1000", 10);
+      setTimeout(() => {
+        res.writeHead(200, { "content-type": "text/plain" });
+        res.end("delayed");
+      }, ms);
```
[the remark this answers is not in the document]
trivikr: This is a test code and is not exposed to end users, but I accepted the edits.

trivikr:
There are a lot of issues with monorepo setup which blocks us from using private fields in UndiciHttpHandler.
I'll switch to using TypeScript private member. We can switch to private fields in a major version bump when dropping support for Node.js version in future.

packages/undici-http-handler/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "@smithy/abort-controller": "workspace:^",
+    "@smithy/node-http-handler": "workspace:^",
+    "@types/node": "^18.11.9",
+    "concurrently": "7.0.0",
+    "downlevel-dts": "0.10.1",
+    "premove": "4.0.0",
+    "typedoc": "0.23.23"
+  },
+  "peerDependencies": {
+    "undici": "^6.0.0"
```
kuhe: This should only appear in one of deps and peerDeps, right?

packages/undici-http-handler/src/undici-http-handler.bench.ts
```diff
@@ -0,0 +1,129 @@
+import { once } from "node:events";
```
kuhe, line 1: does this file get compiled into dist-es and types?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+  private externalDispatcher = false;
+
+  constructor(options?: UndiciHttpHandlerOptions) {
+    this.config = { ...options };
+    if (this.config.dispatcher) {
+      this.externalDispatcher = true;
+    }
+  }
+
+  private getOrCreateDispatcher(): Dispatcher {
```
kuhe: private code should go towards the bottom

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    if (this.config.dispatcher) {
+      return this.config.dispatcher;
+    }
+
+    this.config.dispatcher = new Agent({ allowH2: true });
+
+    return this.config.dispatcher;
```
kuhe: you could reuse `this.config` or `this.config.dispatcher`.

```ts
const { config } = this;
const { dispatcher } = config;

if (dispatcher) {
  return dispatcher;
}

return (config.dispatcher = new Agent({ ... }));
```

P.S. how do you feel about using assignments as expressions? 😅

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  public destroy(): void {
+    if (this.config.dispatcher && !this.externalDispatcher) {
+      this.config.dispatcher.destroy();
+      this.config.dispatcher = undefined;
```
kuhe: what if the user intent/expectation is that the destroy method destroys their dispatcher too?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  public async handle(
+    request: HttpRequest,
+    { abortSignal, requestTimeout }: HttpHandlerOptions = {}
```
kuhe: since this is an H2 handler as well, check the handle method on the `node-http2-handler`. There is an additional bit of request context: `isEventStream: boolean`.

When that flag is true, the H2 session must be isolated and not share concurrency with other requests. 

Also, the http2 connect/timeout parameters that are supported by node-http2-handler should also be supported here, if indeed undici uses node:http2 under the hood.

if all that seems too much, perhaps H2 or event stream requests should be rejected by this handler?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
+import type { Readable } from "node:stream";
+import { HttpResponse, buildQueryString, type HttpHandler, type HttpRequest } from "@smithy/core/protocols";
+import type { HttpHandlerOptions, Logger } from "@smithy/types";
+import { Agent, Dispatcher } from "undici";
+
+/**
+ * Options for the UndiciHttpHandler.
```
kuhe, line 15: needs public/alpha/beta annotation

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * Optional logger.
+   */
+  logger?: Logger;
+}
+
+/**
+ * An HTTP handler that uses undici instead of Node.js native http/https modules.
+ * Smithy-compatible request handler backed by undici.
```
kuhe, line 47: needs public/alpha/beta annotation

packages/undici-http-handler/src/undici-http-handler.ts
kuhe: I know you wanted something in H2 that at least worked, but making it available will invite feature requests to bring it up to the same configurability as node-http2-handler.

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+      origin = `${request.protocol}//${username}:${password}@${request.hostname}${port}`;
+    } else {
+      origin = `${request.protocol}//${request.hostname}${port}`;
+    }
+
+    // Strip the Expect header — undici does not support 100-continue and
+    // sends the body immediately, so the header is unnecessary.
+    const headers = request.headers;
+    if ("Expect" in headers) delete headers["Expect"];
+    if ("expect" in headers) delete headers["expect"];
```
kuhe: although there are no other known values at this time, only delete the header if the value is `100-continue`.

kuhe: should this warn, via the input logger, that the request instruction is being ignored?

I think **no**: the SDK user is not really concerned about waiting for 100-continue. It's an internal optimization agreement between the SDK and e.g. S3.

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    // Strip transfer-encoding header for streaming bodies — undici manages
+    // chunked encoding internally for streams, so the explicit header is not
+    // needed and causes issues with content-length negotiation.
+    // Uses the same duck-typing check as undici's isStream (pipe + on).
+    const body = request.body as Readable | undefined;
+    if (body && typeof body.pipe === "function" && typeof body.on === "function") {
+      if ("transfer-encoding" in headers) delete headers["transfer-encoding"];
+      if ("Transfer-Encoding" in headers) delete headers["Transfer-Encoding"];
```
kuhe: according to your comments in https://github.com/trivikr/trivikr-test-undici-http-handler/pull/26, does undici fail for _all_ values of this header?

If there are any values at all that would succeed, they should not be deleted.
Either delete only the `chunked` header because that is the only value set by the expected internal middlewares, or... throw an error otherwise?

If the only thing that can set a different transfer-encoding header is a custom _user_ middleware, then perhaps we should tell them that it is incompatible with undici and make them rewrite it rather than silently deleting it.

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    const headersTimeout = requestTimeout !== undefined ? requestTimeout || undefined : undefined;
+    const bodyTimeout = requestTimeout !== undefined ? requestTimeout || undefined : undefined;
```
kuhe: a little bit hard to read - is the intent this?

```ts
// greater than 0 number or undefined.
const timeout: number | undefined = requestTimeout > 0 ? val : undefined;

const headersTimeout = timeout;
const bodyTimeout = timeout;
```

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    // chunked encoding internally for streams, so the explicit header is not
+    // needed and causes issues with content-length negotiation.
+    // Uses the same duck-typing check as undici's isStream (pipe + on).
+    const body = request.body as Readable | undefined;
+    if (body && typeof body.pipe === "function" && typeof body.on === "function") {
+      if ("transfer-encoding" in headers) delete headers["transfer-encoding"];
+      if ("Transfer-Encoding" in headers) delete headers["Transfer-Encoding"];
+    }
+
+    const headersTimeout = requestTimeout !== undefined ? requestTimeout || undefined : undefined;
```
kuhe: if there are request level timeouts there should be timeouts configurable on the requestHandler class options

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+      const {
+        statusCode,
+        headers: responseHeaders,
+        body: responseBody,
+      } = await dispatcher.request({
+        origin,
+        path,
+        method: request.method as Dispatcher.HttpMethod,
+        headers,
+        body: request.body ?? null,
```
kuhe, line 139: is there a need to convert undefined to null here?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+      if (err?.code === "UND_ERR_SOCKET") {
+        throw Object.assign(err, { name: "RequestTimeout" });
+      }
+      throw err;
+    }
+  }
+
+  public updateHttpClientConfig<K extends keyof UndiciHttpHandlerOptions>(
+    key: K,
+    value: NonNullable<UndiciHttpHandlerOptions[K]>
```
kuhe: what if this is an attempt to unset something?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    if (value === this.config.dispatcher) {
+      return;
+    }
+
+    // Capture the previous dispatcher before assignment.
+    const previousDispatcher = this.config.dispatcher;
+
+    // Destroy the previous dispatcher only if it was internally created.
+    if (previousDispatcher && !this.externalDispatcher) {
+      previousDispatcher.destroy();
```
kuhe: is there a soft destroy/close method to allow requests in flight to complete?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    // Assign the new value and update externalDispatcher based on it.
+    this.config.dispatcher = value as Dispatcher;
+    this.externalDispatcher = true;
+  }
+
+  public httpHandlerConfigs(): UndiciHttpHandlerOptions {
+    return { ...this.config };
```
kuhe, line 248: why copy here? in node http2 I see `return this.config ?? {}`

packages/undici-http-handler/README.md
```diff
@@ -0,0 +1,68 @@
...
+  connections: 50,
+  headersTimeout: 3000,
+  bodyTimeout: 3000,
+  connect: {
+    timeout: 3000,
+  },
+});
+
+const client = new S3({
+  requestHandler: new UndiciHttpHandler({ dispatcher }),
```
kuhe, line 70: Constructor pass-through should be supported like in NodeHttpHandler:

Although top level ctor passthrough is not support-able unless 
undici request handler becomes the default for some client,
```ts
const client = new Client({
  requestHandler: {
    dispatcher: {}
  }
});
```

We can still support ctor passthrough one level deeper:

```ts
const client = new S3({
  requestHandler: new UndiciHttpHandler({ 
    dispatcher: {
      connections: 50,
    }
  }),
});
```

Can we do this by assuming the default desired dispatcher is `Agent`?

This lets users avoid a direct dependency on undici. In AWS Lambda, if customers continue using the Lambda-bundled SDK, they might not have import access to undici.

packages/undici-http-handler/src/undici-http-handler.bench.ts
kuhe: answered below: no, excluded by tsconfigs

perhaps this should sit in `/test` rather than `/src`? We already have a convention of unit spec files being adjacent to source files, but integration and other types of tests can be somewhat separated.

doing so would avoid the custom tsconfig and the customization in the src linter.

scripts/check-dependencies.js
```diff
@@ -50,7 +50,7 @@ const node_libraries = [
     for await (const file of walk(srcPath, ["node_modules"])) {
       const contents = fs.readFileSync(file);
 
-      if (file.endsWith(".spec.ts")) {
+      if (file.endsWith(".spec.ts") || file.endsWith(".bench.ts")) {
```
kuhe, line 53: if we move the bench to `/tests` it wouldn't need this exception, as only src is walked.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: > how do you feel about using assignments as expressions? 😅

I initially didn't like it because it was different, but now I'm used to it.
The AI tools started using it sometime in Jan 2026 from my experience.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: By default, we should not destroy the dispatcher as it's external.
But I can also see that most of the customers would likely create dispatcher specifically for the client.

I would like to keep this behavior, and make a change in major version bump is customers request it.
Their existing code, which destroys the custom dispatcher would be backward compatible.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: It should not warn, as it's the contract between handler and undici is to not have expect header - similar to transfer-encoding header. User need not be aware of this.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Yes. Undici throws error for the presence of transfer-encoding header
Source: https://github.com/nodejs/undici/blob/314ba6a2cdd83a5b1e7fcca409cd3a75691a8c00/lib/core/request.js#L509-L511

In the test PR, I went through values accepted in [Transfer-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) header - but couldn't find a use case for it's value being compress/deflate/gzip in SDKs. One of the values - `compress` - is very old and disappeared from most UNIX distributions because of a patent issue.

Irrespective of what path we choose, customer could get an error when testing those values. The error from service backend would be more valuable to understand their use case, over `invalid transfer-encoding header` which undici throws.

Although very less likely, there's a non-zero possibility of service call being successful and invalid data being stored on the server. I think we should play safe, and remove the value if it's only `chunked`. The check can always be made generic in future.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: If we don't copy, the callee can mutate `dispatcher`/`logger` directly and can bypass the intended runtime encapsulation. We should update node http2 handler implementation.

scripts/check-dependencies.js
trivikr: This should be okay, since `.bench.ts` is a recommended standard for naming benchmark files in vitest.
And we're likely going to use the recommendation for future benchmarks.

Historically, we've use `test` folder only for clients as we overwrite the src folder which is code generated. For non-clients modules, we tend to keep tests as close to the source as possible.

I'm open to naming benchmark files as `*.bench.spec.ts`, but that would be against the recommendation from vitest. Do you have a preference?

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Moved in [`1730a56` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/1730a56efb7d53a06f5c2b3897d6fb70d8f54da0)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Updated in [`ef5fd38` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/ef5fd38844c63186e521902315519e6522c21b75)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Added in [`7d28db6` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/7d28db690c9601d726a3423f26f9df248323fba2)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Added in [`7d28db6` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/7d28db690c9601d726a3423f26f9df248323fba2)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Yes. Updated in [`aacd105` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/aacd105aa8783742909e8cff5d02e26dcb6f1f5a)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: The `100-continue` check was added in [`0035a7c` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/0035a7cb8a44dcfd445a8cce673c992725b2a2e3)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: The `chunked` check was added in [`d77b94b` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/d77b94b06b8f08bc98cc06149b89fc4dc1b4e546)

scripts/check-dependencies.js
kuhe: leaving `.bench.ts` here is fine, is there some way to globally exclude the pattern from compilation into dist-cjs/es/types?

packages/undici-http-handler/package.json
trivikr: A module, when it appears in both `dependencies` and `peerDependencies` within a package.json file, it usually means a default fallback for the dependency.

When we specify `undici@^6.0.0` in dependencies, that mean it's the default. When we specify `undici@^6.0.0` in peerDepedencies, it means customers can use those versions when passing their custom dispatcher.

The handler as it's currently designed will accept `7.x` and `8.x` versions of undici too, if customers on newer versions of Node.js and want to avail performance benefits.

trivikr: I considered removing `undici` from dependencies and asking customers to provide a dispatcher.
But customers will have to import both the handler and undici, which won't be a good minimal experience.

Existing minimal code
```js
import { UndiciHttpHandler } from "@smithy/undici-http-handler";

const client = new S3({
  requestHandler: new UndiciHttpHandler(),
});
```

Minimal code if we require customer to pass dispatcher
```js
import { UndiciHttpHandler } from "@smithy/undici-http-handler";
import { Agent } from "undici";

const client = new S3({
  requestHandler: new UndiciHttpHandler({ dispatcher: new Agent() }),
});
```

I think the former is better.

packages/undici-http-handler/README.md
trivikr: Added in [`ebf2fab` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/ebf2fabda8dc64c9720eed636036a49a0df86204)

trivikr: I also updated READE in [`8c3a032` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/8c3a032766b37d19b666bb6fec18e92b7de9ed91) with example for undici `Client` which customers can pass to reduce overhead further

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Removal of null check might work, but is required as per undici types.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Removed `NonNullable` keeping dispatcher as no-op in [`d355253` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/d355253bf19d43defea0d3b8e59b275179516427)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Changed to close in [`f909f89` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/f909f89900f1c9ed2d87b2e38c00e306a4f71a4c)

packages/undici-http-handler/src/undici-http-handler.ts
kuhe: update: I realized that we destroy the user provided agent in node-http-handler, so it would make sense to also do so here

ref: https://github.com/smithy-lang/smithy-typescript/blob/396de9c6de6020bdc353e338d60afdbe196e6bc1/packages/node-http-handler/src/node-http-handler.ts#L131-L134

packages/undici-http-handler/src/undici-http-handler.ts
kuhe: should the return type be marked `Readonly<UndiciHttpHandlerOptions>`?

packages/undici-http-handler/package.json
kuhe: I see, so should the one in peerDep be stated as `>=6.0.0` then?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (headers["Expect"] === "100-continue") delete headers["Expect"];
+    if (headers["expect"] === "100-continue") delete headers["expect"];
+
+    // Strip transfer-encoding header for streaming bodies — undici manages
+    // chunked encoding internally for streams, so the explicit header is not
+    // needed and causes issues with content-length negotiation.
+    // Uses the same duck-typing check as undici's isStream (pipe + on).
+    const body = request.body as Readable | undefined;
+    if (body && typeof body.pipe === "function" && typeof body.on === "function") {
+      if (headers["transfer-encoding"] === "chunked") delete headers["transfer-encoding"];
```
siddsriv, line 120: http headers are case insensitive entirely, things like `EXPECT` and `Transfer-encoding` will slip through here, is it ok?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,230 @@
...
+    // Build path with query string — skip buildQueryString when query is undefined.
+    let path = request.path;
+    if (request.query) {
+      const queryString = buildQueryString(request.query);
+      if (queryString) {
+        path += `?${queryString}`;
+      }
+    }
+    if (request.fragment) {
+      path += `#${request.fragment}`;
```
siddsriv: does this need to be appended and sent over the wire? reading some RFCs around this here ([9110](https://www.rfc-editor.org/rfc/rfc9110#section-4.2.5), [3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.5), [9112](https://www.rfc-editor.org/rfc/rfc9112#section-3.2)) it seems like the server could reject it or ignore it. 

> Some protocol elements that refer to a URI allow inclusion of a fragment, while others do not.

> Note: The fragment identifier component is not part of the scheme definition for a URI scheme....

and it does not seem to be a part of the request-targets either (RFC 9112)

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,230 @@
...
+        bodyTimeout,
+        signal: abortSignal as AbortSignal | undefined,
+      });
+
+      // Transform undici headers (Record<string, string | string[]>) to HeaderBag (Record<string, string>)
+      const transformedHeaders: Record<string, string> = {};
+      for (const key in responseHeaders) {
+        const value = responseHeaders[key];
+        if (value !== undefined) {
+          transformedHeaders[key] = Array.isArray(value) ? value.join(", ") : value;
```
siddsriv: does this transformation need to be unconditional? we might have some perf gains here: most headers might not require this transformation, they should be strings. We might not need to make a new allocation and `x` iterations

we could check if the value is indeed an array otherwise skip this copying. i'm hoping the multi-value-header case is indeed rare in which case this could be optimized further

packages/undici-http-handler/package.json
trivikr: Yes. I've made the change in [`aa09bed` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/aa09bed75a7ce8f9583d39f5776236bf1c1ef79c)

trivikr: I also made the change to match engines.node to that of undici dependency, and added a test for it in [`cc1141d` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/cc1141d57f1ec04b06389f69dc9ebbdd1caa19c3)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Done in [`bdcf1d2` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/bdcf1d22315a51125bad23675ce138c50c1af051)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Since we allowed passing `Agent.Options` in UndiciHttpHandlerOptions in [`ebf2fab` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/ebf2fabda8dc64c9720eed636036a49a0df86204), it's no longer tied to `UndiciHttpHandlerOptions` and is a separate type.

If we want to mark it as `ReadOnly`, it needs to be done at interface level and in all other implementations. Since it's an internal API, it should be okay. Marking as ReadOnly might be confusing too, since it only prevents direct property assignment on the returned object but doesn't provide deep immutability

scripts/check-dependencies.js
trivikr: I'm just excluding it from undici-http-handler tsconfigs as of now. We should take it up when working on standardization of root tsconfigs (extending from tsconfig bases) in future.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Support added by using isolated client for event stream requests in [`da9ccf5` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/da9ccf5778dae804bc249b43a60327415bcd97d1)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: The request level timeouts are for the callee of the handle function, i.e. within the SDKs.
For RequestHandler class options, we allow end users to pass Dispatcher or Agent.Options instead.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Yes. The SDK uses lowercase http headers. We support PascalCase in case it's set by customer middleware.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Lazily transformed response headers in [`d0105e4` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/d0105e4f0d5d07571109a27999708237f63482ba)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Removed fragment in [`c92b978` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/c92b978211b505cc0b512fbdbb0b1ab35d152fb4)

trivikr:
E2E tests for the latest changes were run in https://github.com/aws/aws-sdk-js-v3/pull/8034

trivikr:
The latest benchmarks show 30%-40% improvement across several runs.
The max I noticed it 55% for serial and 35% for parallel. Tested on node 18 as well as 24.

```console
 DEV  v3.2.4 /local/home/trivikr/workspace/smithy-typescript/packages/undici-http-handler


 ✓ src/undici-http-handler.bench.ts > 10 sequential GETs 1208ms
     name                     hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    1,385.83  0.5811  1.4585  0.7216  0.7641  1.2529  1.2864  1.4585  ±1.35%      693
   · UndiciHttpHandler  2,061.69  0.4148  1.9062  0.4850  0.4856  1.1038  1.2146  1.2764  ±1.36%     1031

 ✓ src/undici-http-handler.bench.ts > 50 concurrent GETs 1250ms
     name                   hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    363.27  2.5150  4.1371  2.7528  2.7913  3.6991  4.1371  4.1371  ±1.40%      182
   · UndiciHttpHandler  473.83  1.9688  3.1645  2.1105  2.1024  2.8682  2.8725  3.1645  ±1.23%      237

 BENCH  Summary

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 10 sequential GETs
    1.49x faster than NodeHttpHandler

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 50 concurrent GETs
    1.30x faster than NodeHttpHandler
```

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,281 @@
...
+ * Options for the UndiciHttpHandler.
+ *
+ * @public
+ */
+export interface UndiciHttpHandlerOptions {
+  /**
+   * You can pass an existing undici Dispatcher (Agent, Pool, Client, etc.)
+   * or Agent.Options to have one created for you.
+   */
+  dispatcher?: Dispatcher | Agent.Options;
```
kuhe, line 24: top level options don't directly have timeouts. Are they in Agent.Options?

Secondly, if we ever hook this up to an AWS SDK client, the defaults mode timeout config resolution is async. That could lead to a big refactor of how config is resolved after init, but since we have major version increment control, it shouldn't be a problem for now.

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,281 @@
...
+      const httpResponse = new HttpResponse({
+        statusCode,
+        headers: (transformedHeaders ?? responseHeaders) as Record<string, string>,
+        body: responseBody,
+      });
+
+      // Destroy the isolated client once the response body stream closes.
+      if (isolatedClient) {
+        (responseBody as Readable).once("close", () => {
+          isolatedClient.destroy();
```
kuhe: should close be called here instead of destroy?

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -0,0 +1,281 @@
...
+        });
+      }
+
+      return { response: httpResponse };
+    } catch (err: any) {
+      // Ensure the isolated client is cleaned up on errors.
+      if (isolatedClient) {
+        isolatedClient.destroy();
```
kuhe, line 182: this is clearer than `isolatedClient?.destroy()`, any preference?

packages/undici-http-handler/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+### Configuring undici Dispatcher
+
+You can pass `Agent.Options` to configure transport behavior such as connection
+pooling and timeouts. The handler creates an `Agent` internally for you.
+
+```js
+import { S3 } from "@aws-sdk/client-s3";
+import { UndiciHttpHandler } from "@smithy/undici-http-handler";
```
kuhe, line 34: since we're using S3 in the example, want to add this handler to `@aws-sdk/config/requestHandler`? 

It currently only exports Fetch and NodeHttpHandlers.

kuhe (approved):

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Yes. Since UndiciHttpHandler is tightly coupled with undici, we request callee to provide Agent.Options.

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: The `if` check if more readable. Changed in [`a1215db` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/a1215dbd55ec73b80fe39c78a18493ef00df1974)

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Yes. Updated in [`a893a3d` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2029/changes/a893a3d176575866e036327c8f5a398babb4d691)

packages/undici-http-handler/README.md
trivikr: Since UndiciHttpHandler is not included by default in JS SDK client, I'll leave it from `@aws-sdk/config` with the first release.

Let's wait for usage and feedback post release, and add it in future if needed.

trivikr:
Our release script seems to have some issue with publishing a new package

```console
🦋  info Publishing "@smithy/undici-http-handler" at "1.0.0"
🦋  error an error occurred while publishing @smithy/undici-http-handler: E404 Not Found - PUT https://registry.npmjs.org/@smithy%2fundici-http-handler - Not found 
🦋  error '@smithy/undici-http-handler@1.0.0' is not in this registry.
🦋  error 
🦋  error Note that you can also install from a
🦋  error tarball, folder, http url, or git url.
🦋  error npm warn publish npm auto-corrected some errors in your package.json when publishing.  Please run "npm pkg fix" to address these errors.
🦋  error npm warn publish errors corrected:
🦋  error npm warn publish "repository.url" was normalized to "git+https://github.com/smithy-lang/smithy-typescript.git"
🦋  error npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
🦋  error npm error code E404
🦋  error npm error 404 Not Found - PUT https://registry.npmjs.org/@smithy%2fundici-http-handler - Not found
🦋  error npm error 404
🦋  error npm error 404  '@smithy/undici-http-handler@1.0.0' is not in this registry.
🦋  error npm error 404
🦋  error npm error 404 Note that you can also install from a
🦋  error npm error 404 tarball, folder, http url, or git url.
🦋  error npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2026-05-20T16_49_28_319Z-debug-0.log
🦋  error 
🦋  error packages failed to publish:
🦋  @smithy/undici-http-handler@1.0.0
```

https://github.com/smithy-lang/smithy-typescript/actions/runs/26176337520/job/77007309986

The package was published manually with a 2PR

In root
```console
$ smithy-typescript> git status
On branch main
Your branch is up to date with 'origin/main'.

$ smithy-typescript> git clean -dfx && yarn && yarn build
```

In package:
```
$ undici-http-handler> npm whoami
<npm-publish-account>

$ undici-http-handler> ls dist-*
dist-cjs:
index.js

dist-es:
index.js                undici-http-handler.spec.js
undici-http-handler.js  undici-http-handler.version.spec.js

dist-types:
index.d.ts                undici-http-handler.spec.d.ts
undici-http-handler.d.ts  undici-http-handler.version.spec.d.ts

$ undici-http-handler> npm publish --access public
...
npm notice name: @smithy/undici-http-handler
npm notice version: 1.0.0
npm notice filename: smithy-undici-http-handler-1.0.0.tgz
npm notice package size: 12.0 kB
npm notice unpacked size: 56.9 kB
npm notice shasum: 1b79fc801d4574fe756789fd659f925c4d12e8fa
npm notice integrity: sha512-wrYVYprqhmwCt[...]K7/uHaDHvQoEQ==
npm notice total files: 12
npm notice
npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
+ @smithy/undici-http-handler@1.0.0
```

The package is available on npm https://www.npmjs.com/package/@smithy/undici-http-handler

trivikr:
We missed updating `workspace:^` entries when doing manual publish leading to installation failures.

```console
$ npm install @smithy/undici-http-handler --save-exact
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:^
```

A dummy PR is posted at https://github.com/smithy-lang/smithy-typescript/pull/2035, and we'll perform automated release of `@smithy/undici-http-handler@1.0.1`

siddsriv:
Due to the npm token being expired, we were temporarily unable to publish packages earlier. Internal tracker for the team: P437384678 for more details.

PR #2031, opened 2026-05-14
feat(node-http-handler): passes client logger to request handler in codegen

siddsriv:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/6130

*Description of changes:*
`NodeHttpHandler` emits socket exhaustion and request timeout warnings via `console.warn`, even when the user sets a custom logger on the SDK client. The client's logger and the request handler are assembled independently during client construction.
- `NodeHttpHandler.create()` and the constructor now accept an optional `clientLogger` parameter. This is used as a fallback, if the user didn't explicitly set logger in the handler's own options, the client logger is used instead.
- The generated `runtimeConfig.ts` now passes `clientSharedValues.logger` as an argument to `RequestHandler.create()` 
- Precedence: handler-level logger > client-level logger > console (fallback)

*Testing*
- new unit tests
```console
RUN  v3.2.4 /../smithy-typescript
  
   ✓ packages/node-http-handler/src/node-http-handler.spec.ts (38 tests)
  61ms
  
   Test Files  1 passed (1)
        Tests  38 passed (38)
```

*Usage*
Before (warning goes to console.warn):
```
  const client = new S3Client({
    logger: myLogger,
  });
  // Socket exhaustion warning still prints to console.warn
```
  After (warning goes to myLogger.warn):
``` 
  const client = new S3Client({
    logger: myLogger,
  });
  // Socket exhaustion warning now routed through myLogger.warn
```
Output:
```
Output:
  
  [myLogger.warn] @smithy/node-http-handler:WARN - socket usage at
  capacity=50 and 100 additional requests are enqueued.
```

---

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -50,14 +50,15 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
    * or instantiates a new instance of this handler.
    */
   public static create(
-    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>
+    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>,
+    logger?: Logger
   ) {
     if (typeof (instanceOrOptions as any)?.handle === "function") {
       // is already an instance of HttpHandler.
       return instanceOrOptions as HttpHandler<any>;
     }
     // input is ctor options or undefined.
-    return new NodeHttpHandler(instanceOrOptions as NodeHttpHandlerOptions);
+    return new NodeHttpHandler(instanceOrOptions as NodeHttpHandlerOptions, logger);
   }
 
   /**
@@ -114,16 +115,16 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
     return socketWarningTimestamp;
   }
 
-  constructor(options?: NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>) {
+  constructor(options?: NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>, clientLogger?: Logger) {
     this.configProvider = new Promise((resolve, reject) => {
       if (typeof options === "function") {
         options()
           .then((_options) => {
-            resolve(this.resolveDefaultConfig(_options));
+            resolve(this.resolveDefaultConfig(_options, clientLogger));
           })
           .catch(reject);
       } else {
-        resolve(this.resolveDefaultConfig(options));
+        resolve(this.resolveDefaultConfig(options, clientLogger));
       }
     });
   }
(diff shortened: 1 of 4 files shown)


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -373,7 +377,7 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
         }
         return new hsAgent({ keepAlive, maxSockets, ...httpsAgent });
       })(),
-      logger,
+      logger: logger ?? clientLogger,
```
siddsriv, line 380: Handler-level logger (from `NodeHttpHandlerOptions`) takes precedence over the client-level logger.

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -50,14 +50,15 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
    * or instantiates a new instance of this handler.
    */
   public static create(
-    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>
+    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>,
+    logger?: Logger
   ) {
     if (typeof (instanceOrOptions as any)?.handle === "function") {
       // is already an instance of HttpHandler.
       return instanceOrOptions as HttpHandler<any>;
```
siddsriv, line 58: When the user passes a pre-constructed handler instance (e.g.  `requestHandler: new NodeHttpHandler({...})`), the client logger is not forwarded. The `create()` method returns the instance as-is. This is intentional, if you manually construct the handler, you own its configuration. Users who want the client logger in this case should pass logger explicitly in their `NodeHttpHandler` options.

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -50,14 +50,15 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
    * or instantiates a new instance of this handler.
    */
   public static create(
-    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>
+    instanceOrOptions?: HttpHandler<any> | NodeHttpHandlerOptions | Provider<NodeHttpHandlerOptions | void>,
+    logger?: Logger
```
kuhe, line 54: why isn't this in the options arg?

siddsriv: `instanceOrOptions` can be a Provider function, not just an object, so we can't inject into that i believe. A separate param would be cleaner(?), the handler resolves precedence internally as `options.logger ?? clientLogger`.

siddsriv:
Converting to draft to discuss some design details for this one.

packages/node-http-handler/src/node-http-handler.ts
kuhe: discussed elsewhere - 

I don't want to have a separate arg _just_ for the client logger. We have several request handlers and they should all receive loggers through their primary constructor argument object.

It could perhaps be done in one of the config resolvers, which at their point have reference to the partially initialized config containing both the request handler and a client-level logger.

If the request handler has an async-init config, it would look like this:

```ts
function resolveSomeConfig(previouslyResolved) {
  if (previouslyResolved.requestHandler.configProvider is a promise) {
    configProvider = configProvider.then(c => {
      c.logger ??= previouslyResolved.logger || new LogLevel("warn");
    })
  }
}
```

kuhe (changes requested):

PR #2044, opened 2026-05-21
fix(undici-http-handler): require close/destroy in isDispatcher duck-type check

trivikr:
*Issue #, if available:*

Refs: https://github.com/smithy-lang/smithy-typescript/pull/2043#discussion_r3283844782

*Description of changes:*

The isDispatcher check accepted any object with a `request` function, so a
caller-supplied dispatcher-like object without `close` could pass through
and later trigger a TypeError when updateHttpClientConfig invoked
previousDispatcher.close(). Tighten the check to also require `close` and
`destroy` — the three methods this handler actually invokes.

---

### DIFF
--- a/packages/undici-http-handler/src/undici-http-handler.ts
+++ b/packages/undici-http-handler/src/undici-http-handler.ts
@@ -11,8 +11,14 @@ import { Agent, Client, Dispatcher } from "undici";
  * and surface as a TypeError later when `close`/`destroy` is called.
  */
 const isDispatcher = (value: unknown): value is Dispatcher => {
-  if (value instanceof Dispatcher) return true;
-  if (typeof value !== "object" || value === null) return false;
+  if (value instanceof Dispatcher) {
+    return true;
+  }
+
+  if (typeof value !== "object" || value === null) {
+    return false;
+  }
+
   const candidate = value as Record<string, unknown>;
   return (
     typeof candidate.request === "function" &&
--- a/packages/undici-http-handler/src/undici-http-handler.ts
+++ b/packages/undici-http-handler/src/undici-http-handler.ts
@@ -5,11 +5,21 @@ import { Agent, Client, Dispatcher } from "undici";
 
 /**
  * Duck-type check: returns true if the value looks like a Dispatcher
- * (has a `request` method), as opposed to plain Agent.Options.
+ * (has `request`, `close`, and `destroy` methods), as opposed to plain
+ * Agent.Options. We require all three because the handler invokes each
+ * of them — checking only `request` would let a partial mock through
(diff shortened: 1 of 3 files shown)


packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -5,11 +5,21 @@ import { Agent, Client, Dispatcher } from "undici";
...
- * (has a `request` method), as opposed to plain Agent.Options.
+ * (has `request`, `close`, and `destroy` methods), as opposed to plain
+ * Agent.Options. We require all three because the handler invokes each
+ * of them — checking only `request` would let a partial mock through
+ * and surface as a TypeError later when `close`/`destroy` is called.
  */
-const isDispatcher = (value: unknown): value is Dispatcher =>
-  value instanceof Dispatcher ||
-  (typeof value === "object" && value !== null && typeof (value as any).request === "function");
+const isDispatcher = (value: unknown): value is Dispatcher => {
```
kuhe, line 13: If you're concerned about only calling verified methods, the return type should be a `Pick` of `Dispatcher`. 

That way if code tries to call another unchecked method on Dispatcher, the type error will point the future developer back here to update the duck check.

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -5,11 +5,21 @@ import { Agent, Client, Dispatcher } from "undici";
...
+  if (value instanceof Dispatcher) return true;
+  if (typeof value !== "object" || value === null) return false;
```
kuhe: minor: I prefer brace-wrapped control-flow blocks, perhaps we could lint it going forward

packages/undici-http-handler/src/undici-http-handler.ts
kuhe: It would have to be set throughout:
```ts
private config: { 
  dispatcher?: Pick<Dispatcher, 'request' | 'close' | 'destroy'>
}
```

kuhe (approved):

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: I'll leave it out of this PR, as it interferes with `httpHandlerConfigs` function

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: Done in [`560b783` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2044/changes/560b7836234b0ece6e7208117f645ca0ca71ffb1)

PR #2049, opened 2026-05-26
feat(undici-http-handler): bump undici to 7.x

trivikr:
*Issue #, if available:*

Refs: https://github.com/smithy-lang/smithy-typescript/issues/2041#issuecomment-4520459801

*Description of changes:*

A step towards using the latest version of undici in UndiciHttpHandler

The performance difference also improved. It consistently shows above 35% for parallel and near 50% for sequential.

```console
$ undici-http-handler> yarn test:bench
...
 ✓ src/undici-http-handler.bench.ts > 10 sequential GETs 1208ms
     name                     hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    1,479.54  0.5194  2.1635  0.6759  0.6634  1.6065  1.8097  2.1635  ±2.17%      740
   · UndiciHttpHandler  2,148.49  0.3613  1.3769  0.4654  0.4497  1.0650  1.1415  1.1907  ±1.80%     1075

 ✓ src/undici-http-handler.bench.ts > 50 concurrent GETs 1329ms
     name                   hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · NodeHttpHandler    374.69  2.3330  4.9398  2.6689  2.6947  4.2898  4.9398  4.9398  ±2.30%      188
   · UndiciHttpHandler  518.54  1.7406  3.6043  1.9285  1.8922  3.2063  3.2660  3.6043  ±1.76%      260

 BENCH  Summary

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 10 sequential GETs
    1.45x faster than NodeHttpHandler

  UndiciHttpHandler - src/undici-http-handler.bench.ts > 50 concurrent GETs
    1.38x faster than NodeHttpHandler
```

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2050, opened 2026-05-26
ci: skip undici-http-handler build and test on node 18

trivikr:
Experiment (if successful) to be added in https://github.com/smithy-lang/smithy-typescript/pull/2049

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,10 +117,24 @@ jobs:
         run: |
           yarn
           yarn turbo telemetry disable
+      # @smithy/undici-http-handler requires undici, which may drop Node 18
+      # support in a future major. Skip build/test of that package on Node 18
+      # so the rest of the matrix continues to validate it.
       - name: Build and validate API snapshot
+        if: matrix.node != 18
         run: node ./scripts/retry -- make api-snapshot
+      - name: Build and validate API snapshot (skip undici-http-handler)
+        if: matrix.node == 18
+        run: |
+          node ./scripts/retry -- yarn build --filter '!@smithy/undici-http-handler'
+          node scripts/validation/api-snapshot-validation.js --write
+          git diff --exit-code api-snapshot/
       - name: Run unit tests
+        if: matrix.node != 18
         run: node ./scripts/retry -- yarn test
+      - name: Run unit tests (skip undici-http-handler)
+        if: matrix.node == 18
+        run: node ./scripts/retry -- yarn g:vitest run -c vitest.config.mts --exclude='**/undici-http-handler/**'
       - name: Run integration tests
         run: |
           yarn config set enableImmutableInstalls false
--- a/scripts/validation/validate-deps-used.js
+++ b/scripts/validation/validate-deps-used.js
@@ -28,6 +28,13 @@ async function validate(packageDir) {
   const declared = new Set(Object.keys(pkgJson.dependencies || {}));
   const used = new Set();
 
+  // If none of the dist folders exist, the package has not been built.
+  // Skip rather than report every declared dep as unused.
+  const distFolders = ["dist-cjs", "dist-es", "dist-types"];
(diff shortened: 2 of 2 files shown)


trivikr:
Verified that undici-http-handler is skipped in node 18

### node 18

```console
Run node ./scripts/retry -- yarn build --filter '!@smithy/undici-http-handler'
...
Run node ./scripts/retry -- yarn g:vitest run -c vitest.config.mts --exclude='**/undici-http-handler/**'
```

https://github.com/smithy-lang/smithy-typescript/actions/runs/26472771652/job/77954558414?pr=2050

### node 20

```console
Run node ./scripts/retry -- make api-snapshot
...
Run node ./scripts/retry -- yarn test
```

https://github.com/smithy-lang/smithy-typescript/actions/runs/26472771652/job/77954558788?pr=2050

scripts/validation/validate-deps-used.js
```diff
@@ -28,6 +28,13 @@ async function validate(packageDir) {
   const declared = new Set(Object.keys(pkgJson.dependencies || {}));
   const used = new Set();
 
+  // If none of the dist folders exist, the package has not been built.
+  // Skip rather than report every declared dep as unused.
+  const distFolders = ["dist-cjs", "dist-es", "dist-types"];
+  if (!distFolders.some((d) => fs.existsSync(path.join(packageDir, d)))) {
+    return [];
```
kuhe, line 35: This is a failure mode that also presents when there is a build cache error. We should not use this condition to imply anything else.

trivikr: Should I put a hard check for packageDir `undici-http-handler`?
Or alternatively, just skip the check for package when node version is below `<20` like we're doing in CI?

kuhe: yeah, I think it should depend on the specific package too, or the node version as you say.

.github/workflows/ci.yml
```diff
@@ -117,10 +117,24 @@ jobs:
...
+        run: |
+          node ./scripts/retry -- yarn build --filter '!@smithy/undici-http-handler'
+          node scripts/validation/api-snapshot-validation.js --write
+          git diff --exit-code api-snapshot/
       - name: Run unit tests
+        if: matrix.node != 18
         run: node ./scripts/retry -- yarn test
+      - name: Run unit tests (skip undici-http-handler)
+        if: matrix.node == 18
+        run: node ./scripts/retry -- yarn g:vitest run -c vitest.config.mts --exclude='**/undici-http-handler/**'
```
kuhe, line 137: I'd prefer not to have much logic in the YAML. Can the package's build and test scripts detect node version with an execution wrapper and return exit code 0 in those cases?

This wrapper could be reused for other packages with higher node requirements too.

kuhe:
```json
"scripts": {
  "build": "node runIf.mjs 20 -- <build script here>",
  "test": "node runIf.mjs 20 -- <vitest here>",
}
```
---
```js
// runIf.mjs

if (version >= arg3) {
  await exec(...restArgs);
}

process.exit(0);
```

trivikr: This doesn't work for test in current setup, as they're run globally using root vitest config files and not individual package.json scripts.

trivikr:
Superceded by https://github.com/smithy-lang/smithy-typescript/pull/2053

.github/workflows/ci.yml
trivikr: Instead of making change in CI, I made change in Makefile in https://github.com/smithy-lang/smithy-typescript/pull/2053

scripts/validation/validate-deps-used.js
trivikr: Explicit check for `undici-http-handler` added in https://github.com/smithy-lang/smithy-typescript/pull/2053

PR #2054, opened 2026-05-28
fix(node-http-handler): use default imports for node:http and node:https

icholy:
*Issue #, if available:*

*Description of changes:*

This PR changes the node http handler to use default imports for the `node:http` and `node:https` packages.

I noticed that [MSW](https://mswjs.io/) was not intercepting any requests made by the aws sdk. MSW intercepts requests by replacing the `request` property on the default exports on `node:http` & `node:https`. This approach stops working when using anything other than a default import.

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -1,5 +1,5 @@
-import type { Agent as hAgentType, request as hRequestType } from "node:http";
-import { Agent as hsAgent, request as hsRequest, type RequestOptions } from "node:https";
+import type hNodeType from "node:http";
+import hsNode from "node:https";
 import { HttpResponse, buildQueryString, type HttpHandler, type HttpRequest } from "@smithy/core/protocols";
 import type { HttpHandlerOptions, Logger, NodeHttpHandlerOptions, Provider } from "@smithy/types";
 
@@ -16,9 +16,9 @@ import { writeRequestBody } from "./write-request-body";
 export { NodeHttpHandlerOptions };
 
 interface ResolvedNodeHttpHandlerConfig extends Omit<NodeHttpHandlerOptions, "httpAgent" | "httpsAgent"> {
-  httpAgentProvider: () => Promise<hAgentType>;
-  httpAgent?: hAgentType;
-  httpsAgent: hsAgent;
+  httpAgentProvider: () => Promise<hNodeType.Agent>;
+  httpAgent?: hNodeType.Agent;
+  httpsAgent: hsNode.Agent;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


kuhe:
node 22.15+ 23.5+ 24.0 26.0

package.json

```json
{
  "name": "mock-https-example",
  "private": true,
  "type": "module",
  "scripts": {
    "start": "node index.mjs"
  },
  "devDependencies": {
    "@smithy/node-http-handler": "*",
    "@smithy/protocol-http": "*",
    "msw": "^2.7.5"
  }
}
```

hooks.mjs
```js
import { registerHooks } from "node:module";

registerHooks({
  resolve(specifier, context, nextResolve) {
    if (
      (specifier === "node:http" || specifier === "http") &&
      !context.parentURL?.includes("?wrapped") &&
      !context.parentURL?.includes("@mswjs/interceptors")
    ) {
      return { url: "node:http?wrapped", shortCircuit: true };
    }
    return nextResolve(specifier, context);
  },
  load(url, context, nextLoad) {
    if (url === "node:http?wrapped") {
      return {
        source: `
          const http = require('http');
          exports.request = function request(...args) { return http.request(...args); };
          exports.get = function get(...args) { return http.get(...args); };
          exports.Agent = http.Agent;
          exports.globalAgent = http.globalAgent;
          exports.ClientRequest = http.ClientRequest;
          exports.IncomingMessage = http.IncomingMessage;
          exports.OutgoingMessage = http.OutgoingMessage;
          exports.ServerResponse = http.ServerResponse;
          exports.Server = http.Server;
          exports.createServer = http.createServer;
          exports.METHODS = http.METHODS;
          exports.STATUS_CODES = http.STATUS_CODES;
          exports.default = http;
        `,
        format: "commonjs",
        shortCircuit: true,
      };
    }
    return nextLoad(url, context);
  },
});
```

index.mjs

```js
import "./hooks.mjs";
import { setupServer } from "msw/node";
import { http, HttpResponse } from "msw";
import { HttpRequest } from "@smithy/protocol-http";

const server = setupServer(
  http.get("http://localhost:9999/test", () => HttpResponse.json({ mocked: true }))
);
server.listen({ onUnhandledRequest: "error" });

const handler = new NodeHttpHandler();
const { response } = await handler.handle(
  new HttpRequest({
    protocol: "http:",
    hostname: "localhost",
    port: 9999,
    method: "GET",
    path: "/test",
    headers: { host: "localhost:9999" },
  })
);

const chunks = [];
for await (const chunk of response.body) {
  chunks.push(Buffer.from(chunk));
}
console.log("✅ Static imports:", Buffer.concat(chunks).toString());

server.close();
```

Ask MSW to evaluate hooks for 22.15+. Permanently curtailing a Node.js feature for a test framework is not sustainable.

icholy:
The only advantage that named imports give you is the ability to tree shake. Both `node:http` and `node:https` are not tree-shakable. So I'm not clear on what "features" we're losing here. This is an arbitrary style choice that's causing regressions for users.

edit:

This is the de-facto standard interception point for nock, @mswjs/interceptors, OpenTelemetry’s HTTP instrumentation, and APM agents. It is not specific to any one library.

kuhe:
Open an issue here with this request. 

Instrumentation authors relying on internal implementation details should also open issues here requesting validation support.

PR #2061, opened 2026-06-01
fix(core/retry): move adaptive rate limiting to directly before requests

kuhe:
*Issue #, if available:*
JS-6872

*Description of changes:*

This moves the adaptive rate limiter to right before any request (initial or retry). 

Previously the rate limiter would fire after a request and before exponential backoff. Although this resulted in approximately the same amount of rate-limiting in aggregate, the ordering was incorrect.

### DIFF
--- a/packages/core/src/submodules/retry/middleware-retry/configurations.ts
+++ b/packages/core/src/submodules/retry/middleware-retry/configurations.ts
@@ -5,6 +5,7 @@ import type { Logger, Provider, RetryStrategy, RetryStrategyV2 } from "@smithy/t
 import { AdaptiveRetryStrategy } from "../util-retry/AdaptiveRetryStrategy";
 import { StandardRetryStrategy } from "../util-retry/StandardRetryStrategy";
 import { DEFAULT_MAX_ATTEMPTS, DEFAULT_RETRY_MODE, RETRY_MODES } from "../util-retry/config";
+import { Retry } from "../util-retry/retries-2026-config";
 
 /**
  * @internal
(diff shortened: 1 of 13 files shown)


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenUtils.java
```diff
@@ -299,6 +299,8 @@ static List<String> getFunctionParametersList(Map<String, Object> paramsMap) {
                     }
                 } else if (value instanceof String) {
                     functionParametersList.add(String.format("%s: '%s'", key, value));
+                } else if (value instanceof Number) {
+                    functionParametersList.add(String.format("%s: %s", key, value));
```
kuhe, line 303: needed later for downstream to provide function params that are numbers

packages/core/src/submodules/retry/middleware-retry/configurations.ts
```diff
@@ -86,21 +87,35 @@ export interface RetryResolvedConfig {
 /**
  * @internal
  */
-export const resolveRetryConfig = <T>(input: T & PreviouslyResolved & RetryInputConfig): T & RetryResolvedConfig => {
+export const resolveRetryConfig = <T>(
+  input: T & PreviouslyResolved & RetryInputConfig,
+  defaults?: { defaultMaxAttempts?: number; defaultBaseDelay?: number }
```
kuhe, line 92: the changes to this are to allow separate defaults for specific services

siddsriv (approved):

PR #2062, opened 2026-06-01
docs(undici-http-handler): add migrating from NodeHttpHandler section

trivikr:
*Issue #, if available:*

Requests from slack

*Description of changes:*

Add a "Migrating from NodeHttpHandler" section to the undici-http-handler README that maps each NodeHttpHandler option to its undici Dispatcher equivalent, with a before/after example.

---

### DIFF
--- a/.changeset/perfect-bottles-kick.md
+++ b/.changeset/perfect-bottles-kick.md
@@ -1,0 +1,5 @@
+---
+"@smithy/undici-http-handler": patch
+---
+
+Add a "Migrating from NodeHttpHandler" section to theundici-http-handler README that maps each NodeHttpHandler option toits undici Dispatcher equivalent, with a before/after example.
--- a/.changeset/perfect-bottles-kick.md
+++ b/.changeset/perfect-bottles-kick.md
@@ -2,4 +2,4 @@
 "@smithy/undici-http-handler": patch
 ---
 
-Add a "Migrating from NodeHttpHandler" section to theundici-http-handler README that maps each NodeHttpHandler option toits undici Dispatcher equivalent, with a before/after example.
+Add a "Migrating from NodeHttpHandler" section to the undici-http-handler README that maps each NodeHttpHandler option to its undici Dispatcher equivalent, with a before/after example.
--- a/packages/undici-http-handler/README.md
+++ b/packages/undici-http-handler/README.md
@@ -113,6 +113,45 @@ The `NodeHttpHandler` is configured with top-level timeout fields and
 `http.Agent`/`https.Agent` instances, whereas `UndiciHttpHandler` is configured
 with a single undici `Dispatcher` (or `Agent.Options`).
 
+### Before / After example
+
+This shows the most commonly configured options. See [Option mapping](#option-mapping) below for the complete list.
+
+```js
+// Before: NodeHttpHandler
+import { Agent as HttpsAgent } from "node:https";
+import { NodeHttpHandler } from "@smithy/node-http-handler";
+
+new NodeHttpHandler({
+  connectionTimeout: 3000,
+  requestTimeout: 5000,
+  socketTimeout: 4000,
+  httpsAgent: new HttpsAgent({
+    maxSockets: 50,
+    keepAlive: true,
+    keepAliveMsecs: 1000,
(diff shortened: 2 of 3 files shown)


.changeset/perfect-bottles-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/undici-http-handler": patch
+---
+
+Add a "Migrating from NodeHttpHandler" section to theundici-http-handler README that maps each NodeHttpHandler option toits undici Dispatcher equivalent, with a before/after example.
```
kuhe:
```suggestion
Add a "Migrating from NodeHttpHandler" section to the undici-http-handler README that maps each NodeHttpHandler option to its undici Dispatcher equivalent, with a before/after example.
```

packages/undici-http-handler/README.md
```diff
@@ -104,6 +104,73 @@ client.listTables({}).then(console.log);
...
+| --------------------------------- | ------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `connectionTimeout`               | `dispatcher.connect.timeout`                           | Time allowed for the connect (and TLS) phase.                                                                                                                                                                          |
+| `requestTimeout`                  | `dispatcher.headersTimeout` / `dispatcher.bodyTimeout` | undici splits this into time-to-headers and time-between-body-chunks. The per-request `requestTimeout` option also sets both.                                                                                          |
```
kuhe: requestTimeout is time to last byte in node, fetch, and xhr. 

These values are time to first byte in undici and behave differently.

packages/undici-http-handler/README.md
```diff
@@ -104,6 +104,73 @@ client.listTables({}).then(console.log);
...
+| `socketTimeout`                   | `dispatcher.bodyTimeout` / `dispatcher.headersTimeout` | Node's `socketTimeout` fires on socket inactivity during an in-flight request. undici's `bodyTimeout` (time between body chunks) and `headersTimeout` (time waiting for headers) cover the same stalled-request cases. |
+| `httpAgent` / `httpsAgent`        | `dispatcher`                                           | A single undici `Dispatcher` handles both `http:` and `https:`; there is no separate agent per protocol.                                                                                                               |
+| `httpAgent.maxSockets`            | `dispatcher.connections`                               | Maximum connections undici opens per origin. undici defaults to unlimited; `NodeHttpHandler` defaults to `50`.                                                                                                         |
```
kuhe: node http defaults to unlimited as well, yet we overrode it with 50. Is there no need to do this for undici?

kuhe (approved):

packages/undici-http-handler/README.md
trivikr: Created a feature request at https://github.com/smithy-lang/smithy-typescript/issues/2063

packages/undici-http-handler/README.md
trivikr: In NodeHttpHandler, which UndiciHttpHandler is designed as a replacement for, the `requestTimeout` is window from request start through receiving the response headers which is closer to time-to-first-byte of the response than time-to-last-byte.

The requestTimeoutId is cleared in clearTimeouts, which runs at response-headers-received time, not at response-body-complete time.

Source code:
https://github.com/smithy-lang/smithy-typescript/blob/f90b927f87675ff1246b0e2a94bb40b20f1c7b37/packages/node-http-handler/src/node-http-handler.ts#L254-L262

https://github.com/smithy-lang/smithy-typescript/blob/f90b927f87675ff1246b0e2a94bb40b20f1c7b37/packages/node-http-handler/src/node-http-handler.ts#L172-L176

The `await writeRequestBodyPromise` means "send the request body + receive the response headers."

If the timer were meant to bound last-byte, it would have to be cleared after the body stream emits `end`/`close`. The handler doesn't listen for the body's `end`; it gives the stream to the caller.

kuhe: oh right, it's the end of headers, not end of body. since headers are short, the timing difference isn't worth mentioning

kuhe (approved):

PR #2075, opened 2026-06-08
feat(undici-http-handler): support Dispatcher factory function with per-request resolution

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/issues/8081

*Description of changes:*

Allow UndiciHttpHandlerOptions.dispatcher to accept a () => Dispatcher
factory function that is invoked on each request. This enables dynamic
dispatcher resolution for use cases like rotating proxies or runtime
connection pool selection.

The factory is stored separately and called in handle() rather than
cached, so callers control the lifecycle of the returned Dispatcher.
updateHttpClientConfig also supports transitioning between all three
dispatcher forms (instance, options, factory).

---


trivikr:
We decided to support global dispatcher

https://github.com/aws/aws-sdk-js-v3/issues/8081#issuecomment-4711498685

PR #2078, opened 2026-06-08
chore(undici-http-handler): update benchmark to include Client Dispatcher

trivikr:
*Issue #, if available:*

N/A

*Description of changes:*

Add UndiciHttpHandler with Client dispatcher to benchmarks to
demonstrate the additional performance gain when all requests target
a single origin. Update README benchmarks section.

I used lower and upper limits of values shown by re-running the benchmarks 4-5 times

Example output:
```console
  UndiciHttpHandler (Client) - src/undici-http-handler.bench.ts > 10 sequential GETs
    1.15x faster than UndiciHttpHandler
    1.80x faster than NodeHttpHandler

  UndiciHttpHandler (Client) - src/undici-http-handler.bench.ts > 50 concurrent GETs
    1.07x faster than UndiciHttpHandler
    1.40x faster than NodeHttpHandler
```

---

### DIFF
--- a/.changeset/beige-pens-sniff.md
+++ b/.changeset/beige-pens-sniff.md
@@ -1,0 +1,5 @@
+---
+"@smithy/undici-http-handler": patch
+---
+
+Add Client dispatcher benchmark
--- a/packages/undici-http-handler/README.md
+++ b/packages/undici-http-handler/README.md
@@ -106,15 +106,19 @@ client.listTables({}).then(console.log);
 
 ## Benchmarks
 
-Our benchmark spin up a local HTTP server and runs two scenarios:
+Our benchmark spins up a local HTTP server and runs two scenarios:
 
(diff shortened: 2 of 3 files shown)


.changeset/beige-pens-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smithy/undici-http-handler": patch
+---
+
+Add Client dispatcher benchmark
```
kuhe, line 5: should be an empty changeset as this should not generate a unique version

kuhe (approved):

.changeset/beige-pens-sniff.md
trivikr: Actually, I want to publish this update in a patch version - as it'll appear in README for npm landing page.

PR #2080, opened 2026-06-09
fix(core/retry): limit Retry-After header delay in deprecated StandardRetryStrategy

smilkuri:
*Issue #, if available:*
Internal P441716414

*Description of changes:*
Adds an upper bound for server-side Retry-After delay value.

---

### DIFF
--- a/packages/core/src/submodules/retry/middleware-retry/retry-pre-sra-deprecated/StandardRetryStrategy.ts
+++ b/packages/core/src/submodules/retry/middleware-retry/retry-pre-sra-deprecated/StandardRetryStrategy.ts
@@ -150,8 +150,8 @@ const getDelayFromRetryAfterHeader = (response: unknown): number | undefined => 
   const retryAfter = response.headers[retryAfterHeaderName];
 
   const retryAfterSeconds = Number(retryAfter);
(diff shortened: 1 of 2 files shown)


packages/core/src/submodules/retry/middleware-retry/retry-pre-sra-deprecated/StandardRetryStrategy.ts
```diff
@@ -117,7 +117,7 @@ export class StandardRetryStrategy implements RetryStrategy {
           );
 
           const delayFromResponse = getDelayFromRetryAfterHeader(err.$response);
-          const delay = Math.max(delayFromResponse || 0, delayFromDecider);
```
kuhe, line 120: move to the return value in getDelayFromRetryAfterHeader

```ts
return Math.max(retryAfterDate.getTime() - Date.now(), 20_000);
```

smilkuri: It should be `Math.min(...)` right?

kuhe: yes

kuhe (approved):

PR #2084, opened 2026-06-12
fix(core/protocols): validate hostLabel input

smilkuri:
*Issue #, if available:*
Internal V2228441001

*Description of changes:*
Adds `isValidHostname` validation after host prefix substitution in `HttpProtocol.ts`

---


trivikr (approved):

packages/core/src/submodules/protocols/HttpProtocol.ts
```diff
@@ -157,6 +157,9 @@ export abstract class HttpProtocol extends SerdeContext implements ClientProtoco
           hostPrefix = hostPrefix.replace(`{${name}}`, replacement);
         }
         request.hostname = hostPrefix + request.hostname;
+        if (!isValidHostname(request.hostname)) {
+          throw new Error("ValidationError: prefixed hostname must be hostname compatible.");
```
kuhe: needs unit and integ test

also, this message is the same as the old one but is worded incorrectly. We can take the opportunity to update it.

> ValidationError: hostname containing input prefix is invalid

we don't know for sure that the invalidity of the hostname is caused by the prefix.

packages/core/src/submodules/protocols/HttpProtocol.ts
```diff
@@ -158,6 +158,9 @@ export abstract class HttpProtocol extends SerdeContext implements ClientProtoco
           hostPrefix = hostPrefix.replace(`{${name}}`, replacement);
         }
         request.hostname = hostPrefix + request.hostname;
+        if (!isValidHostname(request.hostname)) {
+          throw new Error("ValidationError: resolved hostname is not a valid hostname.");
```
kuhe:
```suggestion
          throw new Error(`[${request.hostname}] is not a valid hostname.`);
```

can we log this in the error so misuse is noticed quickly in logs?

kuhe (approved):

kuhe (approved):

PR #2091, opened 2026-06-16
feat(core/cbor): string caches for CBOR serde performance

kuhe:
- changes the cbor benchmark to run tasks in workers. This prevents v8 optimization from being distracted by things not being measured.

Previously, JSON serde and CBOR serde were mixed in the process. Now, each worker only runs one of: JSON ser, JSON de, CBOR ser, CBOR de.

- plus performance optimizations described in cbor.md in the diff.

### DIFF
--- a/packages/core/src/submodules/cbor/cbor-decode.ts
+++ b/packages/core/src/submodules/cbor/cbor-decode.ts
@@ -1,4 +1,4 @@
-import { nv, toUtf8 } from "@smithy/core/serde";
+import { nv } from "@smithy/core/serde";
 
 import {
   alloc,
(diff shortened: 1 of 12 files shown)


packages/core/src/submodules/cbor/cbor-decode.ts
```diff
@@ -52,6 +24,10 @@ let _offset: CborOffset = 0;
  */
 export function setPayload(bytes: Uint8Array) {
   payload = bytes;
+  // Buffer view for fast .toString("utf-8", start, end) access in Node.js.
+  if (USE_BUFFER) {
+    payloadBuffer = bytes instanceof Buffer ? bytes : Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength);
```
kuhe: this may affect string perf

siddsriv (approved):

PR #2094, opened 2026-06-16
docs(undici-http-handler): warn about single-origin Client multi-host routing

### LINKED ISSUE #2092: A single-origin undici `Client` dispatcher breaks operations routed to a different host
The [README][] recommends passing an undici `Client` as the dispatcher for a
single-origin performance gain:

> If your application only talks to a single origin, passing a `Client` as the
> dispatcher gives an additional 5%-20% improvement over the default `Agent` by
> skipping per-origin routing.

This guidance has a sharp, undocumented edge: **many AWS SDK clients route
individual operations to different hosts**, so "talks to a single origin" is
often false even when it looks true. An undici `Client` is bound to exactly one
origin, so any operation routed to a different host fails — and with a single
`Client` the failure also wedges the dispatcher, so every subsequent request
hangs (unsettled promise / "unsettled top-level await" warning).

## Why "single origin" is misleading

A single SDK client commonly contacts more than one host:

- **Host-prefixed operations.** Some operations carry `@endpoint(hostPrefix)`.
  e.g. CloudWatch Logs `StartLiveTail` uses prefix `stream-`, so it targets
  `stream-logs.<region>.amazonaws.com` while every other Logs operation uses
  `logs.<region>.amazonaws.com`.
- **S3 virtual-hosted addressing.** `ListBuckets` hits `s3.<region>.amazonaws.com`
  but any bucket-scoped operation hits `<bucket>.s3.<region>.amazonaws.com`.
- **DynamoDB account-based routing.** With `AccountIdEndpointMode=preferred`
  (the default) and an `accountId` on the resolved credentials, operations route
  to `<account-id>.ddb.<region>.amazonaws.com` instead of
  `dynamodb.<region>.amazonaws.com`. The host depends on the account ID resolved
  from credentials at request time, so it cannot even be known when statically
  constructing a single-origin `Client`.

## Minimal reproduction

`host-routing-repro.mjs`:

```js
import { CloudWatchLogs } from "@aws-sdk/client-cloudwatch-logs";
import { DynamoDB } from "@aws-sdk/client-dynamodb";
import { UndiciHttpHandler } from "@smithy/undici-http-handler";

const region = process.env.AWS_REGION || "us-east-1";
const logs = new CloudWatchLogs({
  region,
  requestHandler: new UndiciHttpHandler(),
});
const s3 = new S3({ region, requestHandler: new UndiciHttpHandler() });
const ddb = new DynamoDB({ region, requestHandler: new UndiciHttpHandler() });

// DynamoDB account-based routing engages only when credentials carry an
// accountId; attach one so the example shows the account-specific host.
const baseCreds = ddb.config.credentials;
ddb.config.credentials = async () => {
  const c = await baseCreds();
  return { ...c, accountId: c.accountId ?? process.env.AWS_ACCOUNT_ID };
};

const run = (label, fn) =>
  fn().catch((e) => console.error(`[${label}] ${e?.name}`));

await run("logs.describeLogGroups", () => logs.describeLogGroups({ limit: 1 }));
await run("logs.startLiveTail", () =>
  logs.startLiveTail({
    logGroupIdentifiers: [`arn:aws:logs:${region}:000000000000:log-group:x`],
  }),
);
await run("s3.listBuckets", () => s3.listBuckets({}));
await run("s3.headBucket", () =>
  s3.headBucket({ Bucket: `example-${region}-does-not-exist` }),
);
await run("ddb.listTables", () => ddb.listTables({}));
await run("ddb.getItem", () =>
  ddb.getItem({ TableName: "does-not-exist", Key: { id: { S: "1" } } }),
);

await logs.destroy();
await s3.destroy();
await ddb.destroy();
```

Run with undici debug logging:

```
AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text) \
  NODE_DEBUG=undici node host-routing-repro.mjs 2>&1 | grep "sending request to"
```

### Observed output (hosts differ within each client)

```
UNDICI: sending request to POST https://logs.us-east-1.amazonaws.com/
UNDICI: sending request to POST https://stream-logs.us-east-1.amazonaws.com/
UNDICI: sending request to GET  https://s3.us-east-1.amazonaws.com/?x-id=ListBuckets
UNDICI: sending request to HEAD https://example-bucket-us-east-1-does-not-exist.s3.us-east-1.amazonaws.com/
UNDICI: sending request to POST https://930809265900.ddb.us-east-1.amazonaws.com/
UNDICI: sending request to POST https://930809265900.ddb.us-east-1.amazonaws.com/
```

Distinct hosts across three services from one client each — CloudWatch Logs
(`logs.` vs `stream-logs.`), S3 (`s3.` vs `<bucket>.s3.`), and DynamoDB
(`<account-id>.ddb.`, determined dynamically from credentials).

Two distinct hosts per service, from one client each.

## Failure mode with a single-origin `Client`

When the dispatcher is a `Client` pinned to one origin (e.g.
`new Client("https://logs.<region>.amazonaws.com")`), a request routed to a
different host (`stream-logs.<region>.amazonaws.com`) does not connect to that
host. The `Client` opens its TCP connection to its fixed origin host but
validates TLS against the per-request host, producing:

```
Error [ERR_TLS_CERT_ALTNAME_INVALID]: Host: stream-logs.us-east-1.amazonaws.com
is not in the cert's altnames: DNS:logs.us-east-1.amazonaws.com, ...
```

The failed request stays in the `Client`'s `running` slot, and undici's
servername-mismatch guard in `_resume()` (`if (client[kRunning] > 0) return`)
then blocks every later request. Subsequent calls (e.g. cleanup) never settle,
surfacing as `Warning: Detected unsettled top-level await`.

## Expected behavior

At minimum, the README should warn that a single-origin `Client` must not be
used with clients whose operations may target multiple hosts (host-prefixed
endpoints like CloudWatch Logs `StartLiveTail`, S3 virtual-hosted addressing,
FIPS/dualstack variants). Recommend `Agent` (default) — or `Agent({ connections: 1 })`
for a single connection per origin — when multi-host routing is possible.

Optionally, the handler could detect a second distinct origin and fall back to
an `Agent` instead of failing/wedging.

[README]: https://github.com/smithy-lang/smithy-typescript/blob/main/packages/undici-http-handler/README.md#configuring-undici-dispatcher

trivikr:
*Issue #, if available:*

Fixes: https://github.com/smithy-lang/smithy-typescript/issues/2092

*Description of changes:*

The README recommended passing an undici Client as the dispatcher for a
single-origin performance gain, but many SDK clients route individual
operations to different hosts (host-prefixed endpoints, S3 virtual-hosted
addressing, DynamoDB account-based routing). A Client is pinned to one
origin, so such requests fail TLS validation and wedge the dispatcher.

Clarify that the constraint is scoped to the individual SDK client using
the handler, reword the ambiguous "talks to a single origin" phrasing, and
switch the example to the Lambda client to avoid contradicting the
DynamoDB caveat.

---

### DIFF
--- a/packages/undici-http-handler/README.md
+++ b/packages/undici-http-handler/README.md
@@ -125,7 +125,7 @@ client.listFunctions({}).then(console.log);
 > When a request is routed to a host other than the `Client`'s fixed origin, it
 > fails TLS validation (`ERR_TLS_CERT_ALTNAME_INVALID`) and leaves the
 > dispatcher stuck, so every subsequent request hangs without ever completing.
-> If multi-host routing is possible, use the default `Agent` — or
+> If multi-host routing is possible, use the default `Agent` or
 > `Agent({ connections: 1 })` for a single connection per origin.
 
 > **Tip:** When you pass your own `Dispatcher`, you control which version of
--- a/packages/undici-http-handler/README.md
+++ b/packages/undici-http-handler/README.md
@@ -123,10 +123,10 @@ client.listFunctions({}).then(console.log);
 > - **FIPS/dualstack variants** can also resolve to a different host.
 >
 > When a request is routed to a host other than the `Client`'s fixed origin, it
-> fails TLS validation (`ERR_TLS_CERT_ALTNAME_INVALID`) and wedges the
-> dispatcher, leaving every subsequent request unsettled. If multi-host routing
-> is possible, use the default `Agent` — or `Agent({ connections: 1 })` for a
-> single connection per origin.
+> fails TLS validation (`ERR_TLS_CERT_ALTNAME_INVALID`) and leaves the
+> dispatcher stuck, so every subsequent request hangs without ever completing.
+> If multi-host routing is possible, use the default `Agent` — or
+> `Agent({ connections: 1 })` for a single connection per origin.
 
 > **Tip:** When you pass your own `Dispatcher`, you control which version of
 > `undici` is used. For best performance, install the latest `undici` directly
--- a/packages/undici-http-handler/README.md
+++ b/packages/undici-http-handler/README.md
@@ -73,31 +73,59 @@ const client = new S3({
 client.listBuckets().then(console.log);
 ```
 
-If your application only talks to a single origin (e.g. a Lambda function
-calling one service endpoint), you can pass a `Client` for lower overhead by
-skipping the per-origin routing that `Agent` performs.
+If every operation made by the SDK client using this handler is guaranteed to
+hit the exact same host (e.g. a Lambda function with a single client calling one
+fixed service endpoint), you can pass a `Client` for lower overhead by skipping
+the per-origin routing that `Agent` performs. Note that the dispatcher is scoped
+to the one SDK client this handler is attached to, not the whole application —
+other clients get their own handler.
 
 ```js
-import { DynamoDB } from "@aws-sdk/client-dynamodb";
+import { Lambda } from "@aws-sdk/client-lambda";
 import { UndiciHttpHandler } from "@smithy/undici-http-handler";
 import { Client } from "undici";
 
 const region = "us-east-1";
-const endpoint = `https://dynamodb.${region}.amazonaws.com`;
+const endpoint = `https://lambda.${region}.amazonaws.com`;
 
 const dispatcher = new Client(endpoint, {
   pipelining: 1,
   connect: { timeout: 3000 },
 });
 
-const client = new DynamoDB({
+const client = new Lambda({
   region,
   endpoint,
   requestHandler: new UndiciHttpHandler({ dispatcher }),
 });
 
-client.listTables({}).then(console.log);
+client.listFunctions({}).then(console.log);
 ```
+
+> **Warning:** A `Client` is pinned to exactly one origin, so it must not be
+> used with an SDK client whose operations may target multiple hosts. A client
+> that appears to use one endpoint often still routes individual operations to
+> different hosts:
+>
+> - **Host-prefixed operations.** Operations carrying `@endpoint(hostPrefix)`
+>   target a different host. e.g. CloudWatch Logs `StartLiveTail` uses prefix
+>   `stream-`, hitting `stream-logs.<region>.amazonaws.com` while other Logs
+>   operations use `logs.<region>.amazonaws.com`.
+> - **S3 virtual-hosted addressing.** `ListBuckets` hits
+>   `s3.<region>.amazonaws.com` but bucket-scoped operations hit
+>   `<bucket>.s3.<region>.amazonaws.com`.
+> - **DynamoDB account-based routing.** With `AccountIdEndpointMode=preferred`
+>   (the default) and an `accountId` on the resolved credentials, operations
+>   route to `<account-id>.ddb.<region>.amazonaws.com`. The host depends on the
+>   account ID resolved at request time, so it cannot be known when statically
+>   constructing a single-origin `Client`.
+> - **FIPS/dualstack variants** can also resolve to a different host.
+>
+> When a request is routed to a host other than the `Client`'s fixed origin, it
+> fails TLS validation (`ERR_TLS_CERT_ALTNAME_INVALID`) and wedges the
+> dispatcher, leaving every subsequent request unsettled. If multi-host routing
+> is possible, use the default `Agent` — or `Agent({ connections: 1 })` for a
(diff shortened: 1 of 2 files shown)


packages/undici-http-handler/README.md
```diff
@@ -73,32 +73,60 @@ const client = new S3({
 client.listBuckets().then(console.log);
 ```
 
-If your application only talks to a single origin (e.g. a Lambda function
-calling one service endpoint), you can pass a `Client` for lower overhead by
-skipping the per-origin routing that `Agent` performs.
+If every operation made by the SDK client using this handler is guaranteed to
+hit the exact same host (e.g. a Lambda function with a single client calling one
```
kuhe: "hostname", "origin", or "authority", since "host" may be taken to mean a single EC2 instance.

Origin looks like it appears the most often in the rest of the text, so that should be used throughout.

packages/undici-http-handler/README.md
```diff
@@ -73,32 +73,60 @@ const client = new S3({
...
+client.listFunctions({}).then(console.log);
 ```
 
+> **Warning:** A `Client` is pinned to exactly one origin, so it must not be
+> used with an SDK client whose operations may target multiple hosts. A client
+> that appears to use one endpoint often still routes individual operations to
+> different hosts:
+>
+> - **Host-prefixed operations.** Operations carrying `@endpoint(hostPrefix)`
+>   target a different host. e.g. CloudWatch Logs `StartLiveTail` uses prefix
```
kuhe: "Amazon CloudWatch", since we're in Smithy context here.

S3 and DynamoDB also use the "Amazon" brand prefix instead of "AWS".

packages/undici-http-handler/README.md
```diff
@@ -73,32 +73,60 @@ const client = new S3({
...
+>   `<bucket>.s3.<region>.amazonaws.com`.
+> - **DynamoDB account-based routing.** With `AccountIdEndpointMode=preferred`
+>   (the default) and an `accountId` on the resolved credentials, operations
+>   route to `<account-id>.ddb.<region>.amazonaws.com`. The host depends on the
+>   account ID resolved at request time, so it cannot be known when statically
+>   constructing a single-origin `Client`.
+> - **FIPS/dualstack variants** can also resolve to a different host.
+>
+> When a request is routed to a host other than the `Client`'s fixed origin, it
+> fails TLS validation (`ERR_TLS_CERT_ALTNAME_INVALID`) and wedges the
```
kuhe: wedge, like a door wedge?

packages/undici-http-handler/README.md
```diff
@@ -73,32 +73,60 @@ const client = new S3({
...
+> dispatcher, leaving every subsequent request unsettled. If multi-host routing
+> is possible, use the default `Agent` — or `Agent({ connections: 1 })` for a
```
kuhe:
```suggestion
> is possible, use the default `Agent` or `Agent({ connections: 1 })` for a
```

packages/undici-http-handler/README.md
trivikr: "wedges" meant the dispatcher gets stuck in a broken state and stops making progress, so new requests pile up behind it and never complete. It was rephrased in [`10d642a` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2094/changes/10d642ac03b8064a3e91901b9b24d51dc9fbaac0)

packages/undici-http-handler/README.md
trivikr: Updated in [`b74c301` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2094/changes/b74c3016a1b1b6448f5b27bb9f98934dc1d960a4)

packages/undici-http-handler/README.md
trivikr: Used "origin" in [`6dd05a5` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2094/changes/6dd05a5ec0e8dec86ebaaf2cd8d984f7e32ba6fb)

packages/undici-http-handler/README.md
trivikr: Used official AWS/Amazon service names in [`03e65fe` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2094/changes/03e65fe93f43223726829859a2327311a88040bc)

kuhe (approved):

PR #2095, opened 2026-06-16
fix(undici-http-handler): route event streams through shared dispatcher

trivikr:
*Issue #, if available:*

Refs: https://github.com/smithy-lang/smithy-typescript/pull/2076#discussion_r3421948130

*Description of changes:*

Previously event streams over the internally-managed Agent were given a
dedicated isolated Client to keep a long-lived stream out of the shared
connection pool. Route them through the shared dispatcher instead, like
normal requests.

With HTTP/2 (the default allowH2) streams multiplex over a single
connection, and with HTTP/1.1 undici opens additional connections as
needed (unlimited by default), so a long-lived stream does not block
other traffic. Callers who cap `connections` make an explicit
pool-sizing decision that is now honored for event streams too.

---

### DIFF
--- a/packages/undici-http-handler/src/undici-http-handler.spec.ts
+++ b/packages/undici-http-handler/src/undici-http-handler.spec.ts
@@ -1,7 +1,7 @@
 import { createServer, type IncomingMessage, type Server, type ServerResponse } from "node:http";
 import type { AddressInfo } from "node:net";
 import { HttpRequest } from "@smithy/core/protocols";
-import { Agent, Pool, buildConnector, type Dispatcher } from "undici";
+import { Agent, buildConnector, type Dispatcher } from "undici";
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 
 import { UndiciHttpHandler } from "./undici-http-handler";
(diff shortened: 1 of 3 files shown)


trivikr:
E2E test done in https://github.com/aws/aws-sdk-js-v3/pull/8109

packages/undici-http-handler/src/undici-http-handler.spec.ts
```diff
@@ -495,12 +495,14 @@ describe("UndiciHttpHandler", () => {
...
-      // an isolated client should still be used for event streams.
+    it("routes event streams through the shared Agent when Agent.Options were provided", async () => {
+      // Agent.Options means the handler created the Agent internally. The event
+      // stream still goes through that shared Agent.
+      const sharedAgentRequestSpy = vi.spyOn(Agent.prototype, "request");
       handler = new UndiciHttpHandler({ dispatcher: { connections: 2 } });
       const { response } = await handler.handle(createMockRequest(), { isEventStream: true });
       expect(response.statusCode).toBe(200);
+      expect(sharedAgentRequestSpy).toHaveBeenCalled();
     });
```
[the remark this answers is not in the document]
trivikr: Restored all mocks afterEach in [`edfb5f6` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2095/changes/edfb5f6a83975ba4b72430844eff27bc54cf0ead)

packages/undici-http-handler/src/undici-http-handler.spec.ts
```diff
@@ -443,20 +443,20 @@ describe("UndiciHttpHandler", () => {
...
       handler = new UndiciHttpHandler();
-      // Internal Agent is in use here, so the isolated client path applies.
-      // Event stream goes through an isolated Client instead of the shared Agent.
+      // Event streams are routed through the shared Agent like normal requests,
+      // rather than through a dedicated isolated Client.
       const { response } = await handler.handle(createMockRequest(), { isEventStream: true });
       expect(response.statusCode).toBe(200);
-      expect(sharedAgentRequestSpy).not.toHaveBeenCalled();
+      expect(sharedAgentRequestSpy).toHaveBeenCalled();
     });
```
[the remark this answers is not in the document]
trivikr: Restored all mocks afterEach in [`edfb5f6` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2095/changes/edfb5f6a83975ba4b72430844eff27bc54cf0ead)

packages/undici-http-handler/src/undici-http-handler.ts
```diff
@@ -135,20 +89,9 @@ export class UndiciHttpHandler implements HttpHandler<UndiciHttpHandlerOptions>
 
   public async handle(
     request: HttpRequest,
-    { abortSignal, requestTimeout, isEventStream }: HttpHandlerOptions & EventStreamSignal = {}
```
kuhe, line 138: this breaks a public API

kuhe (approved):

packages/undici-http-handler/src/undici-http-handler.ts
trivikr: We receive a subset of options, as `isEventStream` property is not used.

It was removed in [`50c07a4` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2095/changes/50c07a436730ba2ac9eb41296ed4183ca807d517) but I can always re-add it and not use it.

Is it really needed?

kuhe: I'm just flagging this in case it was unintentional. 

It's unlikely anyone was using this, if you want to go ahead and remove it.

trivikr: Yeah. It should be okay as UndiciHttpHandler is very new and it's very unlikely that anyone would be calling `handle` explicitly. It was optional too.

PR #2097, opened 2026-06-18
feat(credential-provider-imds): add fromInstanceMetadataRegion provider

aBurmeseDev:
_Issue #, if available:_
https://github.com/aws/aws-sdk-js-v3/issues/6183
JS-6904

_Description of changes:_
Adds a named, exported region provider that resolves the host's region from EC2 IMDSv2. 

### DIFF
--- a/packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
+++ b/packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
@@ -1,0 +1,105 @@
+import { ProviderError } from "@smithy/core/config";
+import type { Provider } from "@smithy/types";
+
+import { httpRequest } from "./remoteProvider/httpRequest";
+import { retry } from "./remoteProvider/retry";
+import { getInstanceMetadataEndpoint } from "./utils/getInstanceMetadataEndpoint";
+
+const IMDS_REGION_PATH = "/latest/meta-data/placement/region";
+const IMDS_TOKEN_PATH = "/latest/api/token";
+const X_AWS_EC2_METADATA_TOKEN = "x-aws-ec2-metadata-token";
+const X_AWS_EC2_METADATA_TOKEN_TTL = "x-aws-ec2-metadata-token-ttl-seconds";
+const ENV_IMDS_DISABLED = "AWS_EC2_METADATA_DISABLED";
+const DEFAULT_TIMEOUT_MS = 1000;
+const DEFAULT_TOKEN_TTL_SECONDS = "21600";
+const NEG_CACHE_TTL_MS = 60_000;
+
+// Module-level so the negative result is shared across all provider invocations in a process.
+let negativeCacheUntil = 0;
+
+/**
+ * @public
+ */
+export interface InstanceMetadataRegionInit {
+  /**
+   * The connection timeout (in milliseconds).
+   */
+  timeout?: number;
+  /**
+   * The maximum number of times the HTTP connection should be retried.
+   */
+  maxRetries?: number;
+}
+
+/**
+ * Creates a region provider that sources the region from the EC2 Instance Metadata Service.
+ *
+ * @public
+ */
+export const fromInstanceMetadataRegion =
+  (init: InstanceMetadataRegionInit = {}): Provider<string> =>
+  async (): Promise<string> => {
+    if (process.env[ENV_IMDS_DISABLED]) {
+      throw new ProviderError("IMDS region provider disabled via AWS_EC2_METADATA_DISABLED", {
+        tryNextLink: true,
+      });
+    }
+    if (Date.now() < negativeCacheUntil) {
+      throw new ProviderError("IMDS region recently failed; skipping until cache expires", {
+        tryNextLink: true,
+      });
+    }
+
+    const timeout = init.timeout ?? DEFAULT_TIMEOUT_MS;
+    const maxRetries = init.maxRetries ?? 0;
+
+    try {
+      const endpoint = await getInstanceMetadataEndpoint();
+      const token = (
+        await retry(
+          () =>
+            httpRequest({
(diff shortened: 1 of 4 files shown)


packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,87 @@
...
+      throw new ProviderError("IMDS region recently failed; skipping until cache expires", {
+        tryNextLink: true,
+      });
+    }
+
+    const timeout = init.timeout ?? DEFAULT_TIMEOUT_MS;
+
+    try {
+      const endpoint = await getInstanceMetadataEndpoint();
+      const token = (
```
smilkuri, line 58: Consider wrapping get token in it's own `try/catch` block. If `PUT` request to `/latest/api/token` fails entire provider will throw. Currently [fromInstanceMetdata](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/credential-provider-imds/src/fromInstanceMetadata.ts#L122-L131) handles this by falling back to IMDSv1.

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,87 @@
...
+        await httpRequest({
+          ...endpoint,
+          path: IMDS_TOKEN_PATH,
+          method: "PUT",
+          headers: { [X_AWS_EC2_METADATA_TOKEN_TTL]: DEFAULT_TOKEN_TTL_SECONDS },
+          timeout,
+        })
+      ).toString();
+
+      const region = (
```
smilkuri, line 74: Should we wrap this region `GET` with `retry()` and also expose `maxRetries` on the init?

aBurmeseDev: Wrapped both the token PUT and the region GET in `retry()` and exposed `maxRetries`.

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
aBurmeseDev: This was intentional to have IMDSv2 only, mirroring peer SDKs. On hosts where IMDSv2 is required, I believe v1 region GET would 401 anyway so the fallback might not help. Will check with team internally.

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import { httpRequest } from "./remoteProvider/httpRequest";
+import { retry } from "./remoteProvider/retry";
+import { getInstanceMetadataEndpoint } from "./utils/getInstanceMetadataEndpoint";
+
+const IMDS_REGION_PATH = "/latest/meta-data/placement/region";
+const IMDS_TOKEN_PATH = "/latest/api/token";
+const X_AWS_EC2_METADATA_TOKEN = "x-aws-ec2-metadata-token";
+const X_AWS_EC2_METADATA_TOKEN_TTL = "x-aws-ec2-metadata-token-ttl-seconds";
+const ENV_IMDS_DISABLED = "AWS_EC2_METADATA_DISABLED";
```
kuhe, line 12: are these constants defined elsewhere?

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * The maximum number of times the HTTP connection should be retried.
+   */
+  maxRetries?: number;
+}
+
+/**
+ * Creates a region provider that sources the region from the EC2 Instance Metadata Service.
+ *
+ * @public
```
kuhe, line 37: isn't this an internal API? who would want to use this instead of the instance metadata service?

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export const fromInstanceMetadataRegion =
+  (init: InstanceMetadataRegionInit = {}): Provider<string> =>
+  async (): Promise<string> => {
+    if (process.env[ENV_IMDS_DISABLED]) {
+      throw new ProviderError("IMDS region provider disabled via AWS_EC2_METADATA_DISABLED", {
```
kuhe, line 43: ProviderError is only for credentials errors, isn't it?

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,105 @@
...
+        tryNextLink: true,
+      });
+    }
+    if (Date.now() < negativeCacheUntil) {
+      throw new ProviderError("IMDS region recently failed; skipping until cache expires", {
+        tryNextLink: true,
```
kuhe, line 49: all of these have tryNextLink=true, but what would even sit after this?

packages/credential-provider-imds/src/fromInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,105 @@
...
+   */
+  maxRetries?: number;
+}
+
+/**
+ * Creates a region provider that sources the region from the EC2 Instance Metadata Service.
+ *
+ * @public
+ */
+export const fromInstanceMetadataRegion =
```
kuhe, line 39: this function is duplicated by inferPhysicalRegion, part of the defaults mode provider. there should only be one, possibly in core/config instead of here.

aBurmeseDev:
Closing in favor of https://github.com/smithy-lang/smithy-typescript/pull/2103.

PR #2102, opened 2026-06-25
feat(core): add S3 wire checksum download stream

trivikr:
*Issue #, if available:*

Internal JS-6752

*Description of changes:*

Add createWireChecksumStream with Node.js and browser/web-stream
implementations that validate an S3 wire checksum over a GetObject
response body without buffering the full payload.

The checksum is delivered in the aws-chunked trailer, so the stream
delegates framing decode to a new incremental AwsChunkedDecoder: it
strips chunk framing, surfaces only decoded object data to the caller,
computes the checksum over those bytes, and reads the expected value
from the trailer. Validation runs at end-of-stream and optionally
verifies the decoded byte count against x-amz-decoded-content-length.

---

### DIFF
--- a/.changeset/stale-rockets-heal.md
+++ b/.changeset/stale-rockets-heal.md
@@ -1,0 +1,5 @@
+---
+"@smithy/core": minor
+---
+
+Add createWireChecksumStream to validate S3 wire checksums delivered in the aws-chunked response trailer.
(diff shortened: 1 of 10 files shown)


trivikr:
Replaced by https://github.com/smithy-lang/smithy-typescript/pull/2250

PR #2103, opened 2026-06-25
feat(core/config): IMDSv2 region fallback in default Node region chain

aBurmeseDev:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/6183
JS-6904

*Description of changes:*
- Wires `NODE_REGION_CONFIG_OPTIONS.default` to try IMDSv2 before throwing "Region is missing"
- Consolidates with `inferPhysicalRegion` in `defaults-mode`, which previously inlined its own IMDS helpers. Both now share getInstanceMetadataRegion.

---

### DIFF
--- a/packages/core/src/submodules/config/config-resolver/regionConfig/getInstanceMetadataRegion.ts
+++ b/packages/core/src/submodules/config/config-resolver/regionConfig/getInstanceMetadataRegion.ts
@@ -1,0 +1,128 @@
+import type { IncomingMessage } from "node:http";
+
+import {
+  ENV_IMDS_DISABLED,
+  IMDS_REGION_PATH,
+  IMDS_TOKEN_PATH,
+  X_AWS_EC2_METADATA_TOKEN,
+  X_AWS_EC2_METADATA_TOKEN_TTL,
+} from "../../defaults-mode/constants";
+
+const TIMEOUT_MS = 1000;
+const NEG_CACHE_TTL_MS = 60_000;
+
+let negativeCacheUntil = 0;
+
+interface ImdsEndpoint {
+  hostname: string;
+  port?: number;
+}
+
+/**
+ * Returns the region of the host from the EC2 Instance Metadata Service (IMDSv2),
+ * or undefined if unavailable.
+ *
+ * @internal
+ */
+export const getInstanceMetadataRegion = async (): Promise<string | undefined> => {
+  if (process.env[ENV_IMDS_DISABLED]) {
+    return undefined;
+  }
+  if (Date.now() < negativeCacheUntil) {
+    return undefined;
+  }
+  try {
+    const endpoint = resolveImdsEndpoint();
+    const token = (
+      await imdsRequest({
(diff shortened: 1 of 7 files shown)


packages/core/src/submodules/config/config-resolver/regionConfig/getInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,129 @@
...
+import {
+  ENV_IMDS_DISABLED,
+  IMDS_REGION_PATH,
+  IMDS_TOKEN_PATH,
+  X_AWS_EC2_METADATA_TOKEN,
+  X_AWS_EC2_METADATA_TOKEN_TTL,
+} from "../../defaults-mode/constants";
+
+const TIMEOUT_MS = 1000;
+const TOKEN_TTL_SECONDS = "21600";
```
kuhe: this can be inlined because it is a value where the key already explains what it is

packages/core/src/submodules/config/config-resolver/regionConfig/getInstanceMetadataRegion.ts
```diff
@@ -0,0 +1,129 @@
...
+  const { request } = await import("node:http");
+  return new Promise<Buffer>((resolve, reject) => {
+    const req = request({
+      hostname: options.hostname,
+      port: options.port,
+      path: options.path,
+      method: options.method,
+      headers: options.headers,
+      timeout: TIMEOUT_MS,
+      signal: AbortSignal.timeout(TIMEOUT_MS),
```
kuhe, line 102: are these two redundant? timeout + signal?

kuhe: although it was already there, I think the defaultsMode region inference code was called a lot less than what this is going to do, so it has to be newly evaluated

packages/core/src/submodules/config/defaults-mode/constants.ts
```diff
@@ -22,3 +22,15 @@ export const DEFAULTS_MODE_OPTIONS = ["in-region", "cross-region", "mobile", "st
  * @internal
  */
 export const IMDS_REGION_PATH = "/latest/meta-data/placement/region";
+/**
```
siddsriv, line 25: this might be out of scope for this change, but perhaps we can consolidate these here; i think these constants (or a subset of it) were also declared in other IMDS packages (`ec2-metadata-service` I think)

kuhe (dismissed):

packages/core/src/submodules/config/config-resolver/regionConfig/getInstanceMetadataRegion.ts
aBurmeseDev: Inlined

kuhe (approved):

siddsriv (approved):

PR #2096, opened 2026-06-17
feat(server): add interceptor support to generated SSDK handlers

vismayku:
*Issue #, if available:*

*Description of changes:*

 Adds a server-side interceptor API for the Smithy TypeScript SSDK.

  Generated service and operation handlers now expose `withAuth()` and
  `addInterceptor()` so service teams can plug in auth schemes and interceptors.
  The interceptor pipeline is emitted inline into each generated handler by the
  codegen (`ServerGenerator`), and `@aws-smithy/server-common` provides the shared
  interceptor types.

  Interceptors expose read and modify hooks around each step (auth, deserialize,
  validate, invoke, serialize) and fire in reverse registration order. Auth schemes
  are tried in order and the first one to authenticate wins; if none do, the request
  fails with `UnauthenticatedException` (401).

  Interceptors and auth are opt-in. The handler constructor and `handle()` signature
  are unchanged, so existing services keep working without any changes — with no
  interceptors and no auth registered, the generated handler behaves as before.

  Also in `@aws-smithy/server-common`:
  - New `UnauthenticatedException` (401), wired into the framework error union and
    the generated `serializeFrameworkException`.
  - `isFrameworkException` now safely returns `false` for non-object values instead
    of throwing.

  Notes for reviewers:
  - This touches `smithy-typescript-codegen`, so it is codegen-affecting. The change
    is additive/opt-in, so handlers generated without interceptors or auth produce
    equivalent responses to before. Happy to open the companion aws-sdk-js-v3 sync PR
    if you'd like.
  - No changeset: the modified package is `@aws-smithy/server-common` (under
    `smithy-typescript-ssdk-libs/`, not `/packages`), and `@aws-smithy/*` is in the
    changeset `ignore` list.

If one or more of the packages in the `/packages` directory has been modified, be sure `yarn changeset add` has been run and its output has
been committed and included in this pull request. See CONTRIBUTING.md.

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2104, opened 2026-06-26
feat(ci): auto-version @aws-smithy/* SSDK libs on release

vismayku:
*Issue #, if available:* Follow-up to #2096 (SSDK interceptor support)

*Description of changes:*

Adds automatic versioning for the @aws-smithy/* SSDK libraries to the release-npm-ssdk-libs workflow, mirroring the Version → commit → push pattern in the core release-npm-packages workflow.

 - New scripts/bump-ssdk-alpha.js: bumps each @aws-smithy/* package's prerelease number (alpha.N → alpha.N+1).
 - release-npm-ssdk-libs.yml: runs the bump, then commits & pushes the version change before staging/publishing.
 
 The SSDK libs remain in the changesets ignore list, so the GA @smithy/* release flow is unaffected — their versioning is handled independently by this script. 

---

### DIFF
--- a/.github/workflows/release-npm-ssdk-libs.yml
+++ b/.github/workflows/release-npm-ssdk-libs.yml
@@ -18,6 +18,11 @@ jobs:
           cache: "yarn"
       - name: Install dependencies
         run: yarn install
+      - name: Version
+        id: version
+        run: |
+          node scripts/bump-ssdk-alpha.js
+          echo "porcelain=$(git status --porcelain | wc -l)" >> $GITHUB_OUTPUT
       - name: Configure AWS Credentials
         id: credentials
         uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/release-npm-ssdk-libs.yml
kuhe: I don't think NPM_TOKEN is available anymore, and this script has to be updated to match the release script for the GA packages.

vismayku: Thanks. 
I believe I need to do this : [ref](https://github.com/smithy-lang/smithy-typescript/blob/22bfd5ef220a046b018972ea020c028e5a999a19/.github/workflows/release-npm-packages.yml#L28-L69)

vismayku: @kuhe 
I am not moving the ssdk release to changeset yet. Given it's for the whole project, I would like to maintain separate script to bump versions during release till GA.
Let me know what do you think.

kuhe: that's fine

trivikr: We moved to use npm trusted publishing in https://github.com/smithy-lang/smithy-typescript/pull/2100. Details are in internal issue JS-6577. We plan to use pull request approvals for publishing in the coming weeks, instead of starting a GitHub actions. SSDKs can follow the PR and internal issue once that happens.

kuhe (approved):

kuhe (approved):

PR #2106, opened 2026-06-26
refactor(core/serde): make ChecksumStream a Readable

trivikr:
*Issue #, if available:*

Noticed while working on JS-6752

*Description of changes:*

Change ChecksumStream's base class from Duplex to Readable and consume
the source directly instead of piping it in. The source is attached via
its data/end/error events, the checksum is updated per chunk before
forwarding it with push(), and _read drives backpressure by pausing and
resuming the source.

This removes the writable side that only existed to serve as the pipe
target, so the source is no longer buffered on both a writable and a
readable side. Behavior is unchanged: checksum mismatches and
update/digest errors surface as stream errors during read, the source
is torn down on destroy, and instanceof Readable still holds.

---

### DIFF
--- a/.changeset/selfish-donuts-build.md
+++ b/.changeset/selfish-donuts-build.md
@@ -1,0 +1,5 @@
+---
+"@smithy/core": patch
+---
+
+Make ChecksumStream a Readable that consumes its source directly
--- a/packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.spec.ts
+++ b/packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.spec.ts
@@ -186,6 +186,51 @@ describe(ChecksumStream.name, () => {
 
       await expect(collect(checksumStream)).rejects.toThrow("digest failed");
     });
+
+    it("should surface an error emitted by the source", async () => {
+      const source = makeSource();
+      const checksumStream = new ChecksumStream({
+        expectedChecksum: canonicalBase64,
+        checksum: new Appender(),
+        checksumSourceLocation: "my-header",
+        source,
+      });
+
+      const sourceError = new Error("source failure");
+      const streamError = new Promise<Error>((resolve) => checksumStream.once("error", resolve));
+
+      source.emit("error", sourceError);
+
+      expect(await streamError).toBe(sourceError);
+      expect(checksumStream.destroyed).toBe(true);
+      expect(source.destroyed).toBe(true);
+    });
+  });
+
+  describe("lazy start", () => {
+    it("should not read from the source until the stream is consumed", async () => {
+      const checksum = new Appender();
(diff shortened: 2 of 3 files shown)


trivikr:
This can be rebased and marked for ready when https://github.com/smithy-lang/smithy-typescript/pull/2105 is merged.

kuhe (approved):
Lgtm, I suggest commenting that the Checksum stream is acting as a duplex, but should not be rewritten back into a duplex. Manual control is being used for performance. 

What do benchmarks show?

trivikr:
The Duplex-like design of ChecksumStream was explained in [`ed175b0` (this PR)](https://github.com/smithy-lang/smithy-typescript/pull/2106/changes/ed175b0614200425dd984197bf4698b883d3d786)

trivikr:
# ChecksumStream Benchmark: Duplex vs Readable

Compares `ChecksumStream` before (`main`, extends `Duplex`) and after
(`checksumstream-readable`, extends `Readable`) the refactor.

## Method

- Benchmark: [checksum-stream-perf.ts](https://github.com/user-attachments/files/29610605/checksum-stream-perf.ts)
- Checksum: no-op, to isolate stream plumbing (push/pause/resume/backpressure) rather than hashing cost
- Source: synthetic in-memory `Readable` reusing one pre-allocated buffer (allocation kept out of the measurement)
- Environment: Node.js v24.16.0 (linux x64), `--expose-gc` enabled for RSS deltas

Run:

```console
    NODE_OPTIONS=--expose-gc yarn tsx packages/core/scripts/checksum-stream-perf.ts
```

## Results

| chunk x count | main (Duplex)           | this branch (Readable)  | speedup |
| ------------- | ----------------------- | ----------------------- | ------- |
| 1KB x 4096    | 2.08 GB/s, +19.9 MB RSS | 3.32 GB/s, +6.7 MB RSS  | ~1.60x  |
| 16KB x 1024   | 5.79 GB/s, +1.0 MB RSS  | 7.25 GB/s, +2.3 MB RSS  | ~1.25x  |
| 64KB x 1024   | 27.63 GB/s, +3.2 MB RSS | 39.94 GB/s, +0.0 MB RSS | ~1.45x  |
| 256KB x 512   | 179.66 GB/s, +0.3 MB    | 200.35 GB/s, +0.3 MB    | ~1.12x  |

## Takeaways

- The `Readable` version is faster across all chunk sizes.
- The biggest win is with many small chunks (~1.6x at 1KB), where per-chunk
  stream overhead dominates. This is also where memory drops the most
  (+6.7 MB vs +19.9 MB peak RSS), consistent with removing the double-buffering
  of the old Duplex write side.
- Gains shrink as chunks grow, since fewer chunks means less per-chunk plumbing.

Relative micro-benchmark on a synthetic in-memory source. Real network sources
are far slower and real hashing will dominate total time.

PR #2107, opened 2026-06-29
feat(server): add metrics recorder interface and handler wiring

vismayku:
Add MetricsRecorder/MetricsRecorderFactory to @smithy/types and wire a withMetrics() registration plus begin/recordRequestOutcome/end lifecycle into generated SSDK handlers.

*Issue #, if available:*

*Description of changes:*

 Adds a backend-agnostic metrics abstraction for the SSDK.

  - New MetricsRecorder / MetricsRecorderFactory interfaces in @smithy/types.
  smithy-typescript exposes only the contract; concrete recorders (e.g. an EMF
  implementation) live outside this repo.
  - ServerGenerator emits a chainable withMetrics(factory) on generated handlers. When a
  factory is registered, the handler creates a per-request recorder, calls begin(),
  attaches it to context.metricsRecorder for use in business logic, and records the
  outcome and duration in a finally block so end() always runs, including on the error
  path. Recorder calls are guarded so a misbehaving recorder cannot fail the request.
  - Regenerated the interceptor-example-ssdk handlers so the generated output is
  reviewable in the diff.
  - Added server-common/src/metrics.spec.ts covering the emitted lifecycle:
  begin-before-operation, Success/Fault outcomes, the finally guarantee on the error
  path, context attachment, and recorder errors being swallowed.

A changeset is included for @smithy/types. @aws-smithy/* SSDK libs remain in the changesets ignore list and are versioned independently (#2104).

---


packages/types/src/index.ts
```diff
@@ -16,6 +16,7 @@ export * from "./http";
 export * from "./http/httpHandlerInitialization";
 export * from "./identity";
 export * from "./logger";
+export * from "./metrics";
```
kuhe: no `export *` in new code, see `./CODE_REVIEW.md`

packages/types/src/metrics.ts
```diff
@@ -0,0 +1,128 @@
...
+ * recorders (e.g. an EMF recorder, or a future OpenTelemetry recorder) translate
+ * these abstract calls into their own primitives.
+ *
+ * @typeParam TNativeInstance - the concrete backend handle returned by
+ *   {@link MetricsRecorder.getMetrics}. Defaults to `unknown` so backend-agnostic
+ *   callers must narrow before reaching for backend features.
+ *
+ * @public
+ */
+export interface MetricsRecorder<TNativeInstance = unknown> {
```
kuhe: is the type parameter always going to be supplied in code generation?

I think in that case there should be no default type parameter. What does it mean when native instance is an unknown type?

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
```diff
@@ -176,70 +185,88 @@ export class GetItemHandler<Context> implements __ServiceHandler<Context> {
...
+    const safeRecord = (fn: (recorder: __MetricsRecorder) => void): void => {
+      if (!recorder) {
+        return;
       }
+      try {
+        fn(recorder);
+      } catch {}
+    };
+    safeRecord((r) => r.begin());
+    (context as { metricsRecorder?: __MetricsRecorder }).metricsRecorder = recorder;
```
kuhe: why isn't this on the type of `context`?

private/interceptor-example-ssdk/src/server/operations/Ping.ts
```diff
@@ -25,6 +25,10 @@ import {
 import { HttpRequest as __HttpRequest, HttpResponse as __HttpResponse } from "@smithy/core/protocols";
 import { fromBase64, fromUtf8, toBase64, toUtf8 } from "@smithy/core/serde";
 import { NodeHttpHandler, streamCollector } from "@smithy/node-http-handler";
+import {
+  MetricsRecorder as __MetricsRecorder,
+  MetricsRecorderFactory as __MetricsRecorderFactory,
```
kuhe, line 33: why are these aliased? are there code generated sections that can potentially use these names in this scope?

private/interceptor-example-ssdk/src/server/operations/Ping.ts
```diff
@@ -175,70 +184,88 @@ export class PingHandler<Context> implements __ServiceHandler<Context> {
...
     const entered = new Set<__ServerInterceptor<Context>>();
 
-    try {
-      for (const interceptor of this.interceptors) {
-        if (interceptor.readBeforeExecution) {
-          interceptor.readBeforeExecution(base);
-        }
-        entered.add(interceptor);
+    const recorder: __MetricsRecorder | undefined = this.metricsRecorderFactory?.create();
+    const safeRecord = (fn: (recorder: __MetricsRecorder) => void): void => {
```
kuhe: safeRecord can be a lib function

private/interceptor-example-ssdk/src/server/operations/Ping.ts
```diff
@@ -175,70 +184,88 @@ export class PingHandler<Context> implements __ServiceHandler<Context> {
...
-      let authScheme: string | undefined;
-      if (this.authSchemes.length > 0) {
-        for (const scheme of this.authSchemes) {
-          const result = await scheme.authenticate(request, context);
-          if (result) {
-            caller = result;
-            authScheme = scheme.name;
-            break;
+    try {
+      try {
```
kuhe, line 205: why are there two `try` open blocks here?

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
```diff
@@ -176,70 +185,88 @@ export class GetItemHandler<Context> implements __ServiceHandler<Context> {
...
-        }
-        entered.add(interceptor);
+    const recorder: __MetricsRecorder | undefined = this.metricsRecorderFactory?.create();
+    const safeRecord = (fn: (recorder: __MetricsRecorder) => void): void => {
+      if (!recorder) {
+        return;
       }
+      try {
+        fn(recorder);
+      } catch {}
```
kuhe: add comment in the catch block explaining why ignoring all errors is okay

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
```diff
@@ -176,70 +185,88 @@ export class GetItemHandler<Context> implements __ServiceHandler<Context> {
...
+    };
+    safeRecord((r) => r.begin());
+    (context as { metricsRecorder?: __MetricsRecorder }).metricsRecorder = recorder;
+    const __metricsStart = Date.now();
```
kuhe: performance.now()?

packages/types/src/metrics.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * Outcome of a request, recorded by the framework via {@link MetricsRecorder.recordRequestOutcome}.
+ *
+ * @public
+ */
+export type RequestOutcome = "Success" | "Fault";
```
kuhe, line 6: no Fault distinctions? 4xx/5xx?

smithy-typescript-ssdk-libs/server-common/src/metrics.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  context: { metricsRecorder?: MetricsRecorder };
+  responded: boolean;
+}
+
+/**
+ * Mirrors the metrics lifecycle emitted into a generated ServiceHandler.handle
+ * by ServerGenerator: create a recorder, begin(), attach it to the context, run
+ * the operation guarded by try/finally, then record the outcome and end().
+ */
+async function handleWithMetrics(
```
kuhe: this looks like the generated code. this is going to drift and then fail to cover. can you test the actual generated code?

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
vismayku: Ack

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
vismayku: Ack. Will use it.

packages/types/src/metrics.ts
vismayku: That will be in the follow up PR I will be raising.

smithy-typescript-ssdk-libs/server-common/src/metrics.spec.ts
vismayku: Ack. Will switch to it.

packages/types/src/index.ts
vismayku: Ack

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
vismayku: Context is generic right now where user can pass anything. In future PRs, I will be introducing a concrete typed context. For now,  I haven't touched context parameter and have kept as is.

private/interceptor-example-ssdk/src/server/operations/Ping.ts
vismayku: Ack. can move to server-common

private/interceptor-example-ssdk/src/server/operations/Ping.ts
vismayku: Will fix it.

private/interceptor-example-ssdk/src/server/operations/Ping.ts
vismayku: I followed existing convention for framework import like Authscheme, Caller etc. 
Edit : I believe they are used for conflict avoidance.

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
kuhe: why not introduce typed context first? or else add a todo here

packages/types/src/metrics.ts
vismayku: Intention is to use generics so that we can support swapping backends.

private/interceptor-example-ssdk/src/server/operations/GetItem.ts
vismayku: I can add todo.
Given Context is integrated everywhere, I wanted to keep it separate effort.
Instead of generic type, I would like to introduce a class with parameters as context which we can control.

packages/types/src/metrics.ts
kuhe: generics are ok, my question is about the use of unknown as a default. What does it mean for a native instance to be unknown?

packages/types/src/metrics.ts
```diff
@@ -0,0 +1,123 @@
...
+   * Attaches a property to the request's metrics. A nullish or empty value
+   * removes the property.
+   */
+  setProperty(name: string, value: string | null | undefined): void;
+
+  /**
+   * Returns the recorder's native backend handle for library-specific features.
+   * Callers narrow the result to the concrete recorder's type before use.
+   */
+  getMetrics(): unknown;
```
kuhe: Do you really care how the implementing class makes the native component accessible? If the class itself is unknown type there doesn't need to be an interface method enforcing its availability.

These classes offer a way to retrieve "native" handle with more type information than the interface offers.
```ts
class A implements MetricsRecorder {
  public getMetrics() { throw }
  public getNativeA(): NativeA;
}

class B implements MetricsRecorder {
  public getMetrics() { throw }
  public getNativeB(): NativeB;
}
```

This goes back to why I was asking `unknown` was the _default_ type parameter. The type parameter should be required instead of having a default. Then you get:

```ts
class A implements MetricsRecorder<NativeA> {
  public getMetrics(): NativeA;
}

class A implements MetricsRecorder<NativeB> {
  public getMetrics(): NativeB;
}
```

vismayku: We need native metric access so that user don't need to learn internals of our interface.
For ex, If we have backend implementing Otel, user can directly access Meter object and use it's internal methods.

packages/types/src/metrics.ts
vismayku: My intention is to allow bare use of MetricsRecorder where concrete implementation isn't present. Isn't it better to provide unknown default instead of having/facing undefined?

packages/types/src/metrics.ts
kuhe: use required parametric (generic) type instead of default=unknown or untyped

vismayku: Just to make sure I understand correctly, It's fine to have TNativeInstance type and we should remove unknown. So signature will look like

`export interface MetricsRecorder<TNativeInstance = unknown> { .... }`

and 

`getMetrics(): TNativeInstance;`

kuhe:
```ts
export interface MetricsRecorder<Native> {
  getMetrics(): Native
}
```

Remove `= unknown` to avoid accidentally leaving a known interface untyped.

If there is a real reason to use unknown, and there shouldn't be, it can still be done with `MetricsRecorder<unknown>`.

vismayku: Ack. will make the change

packages/types/src/metrics.ts
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Records a discrete data point where percentiles matter (e.g. per-call
+   * latency, item size).
+   */
+  addMetric(name: string, value: number, unit?: MetricUnit): void;
+
+  /**
+   * Records a unitless ratio (e.g. cache hit rate, throttle rate).
+   */
+  addRatio(name: string, value: number, unit?: MetricUnit): void;
```
kuhe: there should be no unit arg here?

packages/types/src/metrics.ts
```diff
@@ -0,0 +1,132 @@
+/**
```
kuhe, line 1: please move these types to server-common unless you actually need to use them in multiple packages that are downstream of `@smithy/types`

kuhe: I think client and shared types can be in `types`, but SSDK-only types should be in the most upstream SSDK package, which I think is server-common.

vismayku: I will be raising a new PR after this to integrate metric support in client side. We will need this available on client side.

kuhe (approved):

trivikr (approved):

PR #2129, opened 2026-07-07
fix(core): convert string to buffer in checksumStream

smilkuri:
*Issue #, if available:*

Internal JS-7042

*Description of changes:*

Convert string chunks to Buffer in `onSourceData` before passing to `checksum.update()`. Checksum [implementation](https://github.com/aws/aws-sdk-js-v3/blob/main/packages-internal/checksums/src/submodules/crc/crc32c/Crc32cJs.ts#L22) expects buffer and would miscompute when a string is given. Added unit test for string chunk handling.

---

### DIFF
--- a/packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.ts
+++ b/packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.ts
@@ -89,14 +89,13 @@ export class ChecksumStream extends Readable {
     if (this.destroyed) {
       return;
     }
-    const data = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
     try {
-      this.checksum.update(data);
+      this.checksum.update(typeof chunk === "string" ? Buffer.from(chunk) : chunk);
     } catch (e: unknown) {
       this.destroy(e as Error);
       return;
     }
-    if (!this.push(data)) {
+    if (!this.push(chunk)) {
       this.source.pause();
     }
   };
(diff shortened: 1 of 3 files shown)


packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.ts
```diff
@@ -85,17 +85,18 @@ export class ChecksumStream extends Readable {
...
     }
+    const data = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
     try {
-      this.checksum.update(chunk);
+      this.checksum.update(data);
     } catch (e: unknown) {
       this.destroy(e as Error);
       return;
     }
-    if (!this.push(chunk)) {
```
trivikr, line 98: If we're getting string from source, we should not pass buffer to destination.

trivikr (changes requested):

packages/core/src/submodules/serde/util-stream/checksum/ChecksumStream.ts
smilkuri: But wouldn't there be a disparity between for what checksum was computed vs what is actually passed if we do that?

trivikr: Checksum computation happens on line 93. The value is passed to consumer on line 98.
If source passes us a string, it should be passed to the destination as-is.

The `update()` API expects `Uint8Array` so it's the responsibility of the callee, i.e. this code, to convert it to Buffer.

trivikr: Buffer and string are just representations of the data. The checksum value should ideally remain the same.

trivikr (approved):

kuhe (approved):

PR #2144, opened 2026-07-13
smithy-typescript-codegen 0.51.0

smilkuri:
Bump codegen version to 0.51.0

PR #2146, opened 2026-07-13
Revert "chore(codegen): bump codegen version to 0.51.0 (#2144)"

smilkuri:
revert codegen commit https://github.com/smithy-lang/smithy-typescript/pull/2144


siddsriv (approved):

trivikr (changes requested):
Revert is not needed unless there's some issue in generating clients which can't be fixed in v3. Is there one?

PR #2149, opened 2026-07-15
feat(client): add MetricsRecorder support to client middleware

vismayku:
*Issue #, if available:*

*Description of changes:*

Expose a call-site `MetricsRecorder` to Smithy client middleware.

- `client.send(command, { metricsRecorder })` accepts a recorder alongside existing request-handler options.
- `Command.resolveMiddlewareWithContext` transfers it to `handlerExecutionContext[SMITHY_CONTEXT_KEY].metricsRecorder`.
- The recorder is removed before options are passed to the transport request handler.
- Aggregated clients expose a named `<ClientName>RequestOptions` type instead of repeated composed types.
- Existing permissive and disjoint-union handler option types remain supported.

The recorder is intentionally supplied per call because it is request scoped. This PR adds only API and context plumbing; it does not install a metrics middleware or emit metrics. A reusable metrics plugin can be introduced separately.

Tests cover request-option typing, recorder and Smithy-context precedence, transport option filtering, and generated aggregate-client request options.

If one or more of the packages in the `/packages` directory has been modified, be sure `yarn changeset add` has been run and its output has been committed and included in this pull request. See CONTRIBUTING.md.

---

### DIFF
--- a/packages/core/src/submodules/client/smithy-client/command.ts
+++ b/packages/core/src/submodules/client/smithy-client/command.ts
@@ -91,6 +91,7 @@ export abstract class Command<
         ...smithyContext,
       },
       ...additionalContext,
+      recorder: options?.recorder,
     };
     const { requestHandler } = configuration;
     let requestOptions = options ?? {};
--- a/packages/types/src/middleware.ts
+++ b/packages/types/src/middleware.ts
@@ -4,6 +4,7 @@ import type { Command } from "./command";
 import type { EndpointV2 } from "./endpoint";
 import type { SmithyFeatures } from "./feature-ids";
 import type { Logger } from "./logger";
+import type { MetricsRecorder } from "./metrics";
 import type { UserAgent } from "./util";
 
 /**
@@ -583,6 +584,14 @@ export interface HandlerExecutionContext {
      */
     [key: string]: unknown;
   };
+
+  /**
+   * A per-request {@link MetricsRecorder} supplied at the `client.send(command, { recorder })`
+   * call site. When present, a metrics middleware records into this recorder instead of one
+   * minted from a factory, letting a caller (e.g. a server handler) fold the client's request
+   * metrics into its own.
+   */
+  recorder?: MetricsRecorder<any>;
 
   /**
    * Set by some operations which instructs the retry behavior to backoff
--- a/private/my-local-model-schema/src/XYZService.ts
+++ b/private/my-local-model-schema/src/XYZService.ts
@@ -2,6 +2,7 @@
 import { type WaiterResult, createAggregatedClient } from "@smithy/core/client";
 import type {
   HttpHandlerOptions as __HttpHandlerOptions,
+  MetricsRecorder as __MetricsRecorder,
   PaginationConfiguration,
   Paginator,
   WaiterConfiguration,
@@ -58,13 +59,20 @@ const waiters = {
   waitUntilNumbersWhatDoTheyDoAnyway,
 };
 
+/**
+ * @public
+ */
+export interface XYZServiceRequestOptions extends __HttpHandlerOptions {
+  recorder?: __MetricsRecorder<any>;
+}
+
 export interface XYZService {
   /**
    * @see {@link HttpLabelCommandCommand}
    */
   httpLabelCommand(
     args: HttpLabelCommandCommandInput,
-    options?: __HttpHandlerOptions
+    options?: XYZServiceRequestOptions
   ): Promise<HttpLabelCommandCommandOutput>;
   httpLabelCommand(
     args: HttpLabelCommandCommandInput,
@@ -72,7 +80,7 @@ export interface XYZService {
   ): void;
   httpLabelCommand(
     args: HttpLabelCommandCommandInput,
-    options: __HttpHandlerOptions,
+    options: XYZServiceRequestOptions,
     cb: (err: any, data?: HttpLabelCommandCommandOutput) => void
   ): void;
 
@@ -82,7 +90,7 @@ export interface XYZService {
   camelCaseOperation(): Promise<CamelCaseOperationCommandOutput>;
   camelCaseOperation(
     args: CamelCaseOperationCommandInput,
-    options?: __HttpHandlerOptions
+    options?: XYZServiceRequestOptions
   ): Promise<CamelCaseOperationCommandOutput>;
   camelCaseOperation(
     args: CamelCaseOperationCommandInput,
@@ -90,7 +98,7 @@ export interface XYZService {
   ): void;
   camelCaseOperation(
     args: CamelCaseOperationCommandInput,
-    options: __HttpHandlerOptions,
+    options: XYZServiceRequestOptions,
     cb: (err: any, data?: CamelCaseOperationCommandOutput) => void
   ): void;
 
@@ -100,7 +108,7 @@ export interface XYZService {
   getNumbers(): Promise<GetNumbersCommandOutput>;
   getNumbers(
     args: GetNumbersCommandInput,
-    options?: __HttpHandlerOptions
+    options?: XYZServiceRequestOptions
   ): Promise<GetNumbersCommandOutput>;
   getNumbers(
     args: GetNumbersCommandInput,
@@ -108,7 +116,7 @@ export interface XYZService {
   ): void;
   getNumbers(
     args: GetNumbersCommandInput,
-    options: __HttpHandlerOptions,
+    options: XYZServiceRequestOptions,
     cb: (err: any, data?: GetNumbersCommandOutput) => void
   ): void;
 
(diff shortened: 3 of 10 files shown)


packages/types/src/middleware.ts
```diff
@@ -4,8 +4,21 @@ import type { Command } from "./command";
...
+ * Invocation-scoped options made visible to every middleware when a command is
+ * resolved. Carries per-call concerns supplied at the `client.send(command, options)`
+ * call site — currently a {@link MetricsRecorder} — so a middleware can read them
+ * without the value being buried in the transport closure.
+ *
+ * @public
+ */
+export interface InvocationOptions<Native> {
+  recorder?: MetricsRecorder<Native>;
+}
```
kuhe: what else do you expect to add to `InvocationOptions`?

This type does not need to exist. Other comments explain why.

packages/types/src/middleware.ts
```diff
@@ -4,8 +4,21 @@ import type { Command } from "./command";
 import type { EndpointV2 } from "./endpoint";
 import type { SmithyFeatures } from "./feature-ids";
 import type { Logger } from "./logger";
+import type { MetricsRecorder } from "./metrics";
 import type { UserAgent } from "./util";
 
+/**
+ * Invocation-scoped options made visible to every middleware when a command is
+ * resolved. Carries per-call concerns supplied at the `client.send(command, options)`
+ * call site — currently a {@link MetricsRecorder} — so a middleware can read them
```
kuhe:
```suggestion
 * call site, currently a {@link MetricsRecorder}, so a middleware can read them
```

packages/types/src/middleware.ts
```diff
@@ -4,8 +4,21 @@ import type { Command } from "./command";
...
+ * without the value being buried in the transport closure.
+ *
+ * @public
+ */
+export interface InvocationOptions<Native> {
```
kuhe: `Native` is too ambiguous as the type parameter name here at the consumer site.

It's ok in the type definition of `MetricsRecorder`, but here it should be `NativeMetrics`.

packages/types/src/middleware.ts
```diff
@@ -161,7 +174,11 @@ export interface InitializeMiddleware<Input extends object, Output extends objec
    *
    * @param context - Invariant data and functions for use by the handler.
    */
-  (next: InitializeHandler<Input, Output>, context: HandlerExecutionContext): InitializeHandler<Input, Output>;
+  <Native>(
+    next: InitializeHandler<Input, Output>,
+    context: HandlerExecutionContext,
+    invocationOptions?: InvocationOptions<Native>
+  ): InitializeHandler<Input, Output>;
```
kuhe: no change should be made to the middleware stack interface.

I will explain the alternative in other comments.

private/my-local-model-schema/src/XYZService.ts
```diff
@@ -64,15 +65,15 @@ export interface XYZService {
    */
   httpLabelCommand(
     args: HttpLabelCommandCommandInput,
-    options?: __HttpHandlerOptions
+    options?: __HttpHandlerOptions & __InvocationOptions<any>
```
kuhe: Repeated composed types are hard to read. And this is the most user-facing part of any service client API.

Introduce a new type called `RequestOptions` or `InvocationOptions` that extends (is superset of) `__HttpHandlerOptions`, and generate that here instead.

#### also,

what was the point of the `Native` type parameter if clients don't infer it at the public call site?

packages/core/src/submodules/client/smithy-client/command.ts
```diff
@@ -102,7 +102,8 @@ export abstract class Command<
     }
     return stack.resolve(
       (request: FinalizeHandlerArguments<any>) => requestHandler.handle(request.request as HttpRequest, requestOptions),
-      handlerExecutionContext
+      handlerExecutionContext,
+      options
```
kuhe: add the recorder into the `handlerExecutionContext[SMITHY_CONTEXT_KEY]` instead of creating a new object for the request lifecycle.

the handler execution context is already matched to the request lifecycle.

packages/core/src/submodules/client/middleware-stack/MiddlewareStack.ts
```diff
@@ -4,6 +4,7 @@ import type {
   Handler,
```
kuhe, line 4: this file must not be changed for this feature

kuhe (changes requested):
What is the proposed public API for using a smithy client combined with a MetricsRecorder?

It is purely via `client.send(args, { recorder })` and nowhere else? Why is there no client level config?

packages/core/src/submodules/client/middleware-stack/MiddlewareStack.ts
vismayku: Ack

packages/core/src/submodules/client/smithy-client/command.ts
vismayku: Ack. Can switch to it.

packages/types/src/middleware.ts
vismayku: Ack. Can switch to NativeMetrics

packages/types/src/middleware.ts
vismayku: I am adding this interface to basically add anything in future and we can avoid any signature changes. I'm fine switching to context for passing metrics recorder.

vismayku:
> What is the proposed public API for using a smithy client combined with a MetricsRecorder?
> 
> It is purely via `client.send(args, { recorder })` and nowhere else? Why is there no client level config?

Yeah, Users basically passing the metric instance to the call. Open to suggestions if you want me to switch this as a metric middleware step while creating client.

vismayku:
Also, to add one more thing.
This PR just adds public api. I will have separate PR where I propose some common metric and actually using a separate metric to avoid metric clash with existing service metrics.

packages/types/src/middleware.ts
```diff
@@ -576,6 +577,10 @@ export interface HandlerExecutionContext {
     commandInstance?: Command<any, any, any, any, any>;
     selectedHttpAuthScheme?: SelectedHttpAuthScheme;
     features?: SmithyFeatures;
+    /**
+     * The request-scoped metrics recorder selected for this client call.
+     */
+    metricsRecorder?: MetricsRecorder<unknown>;
```
kuhe, line 583: The comment is unnecessary. Everything in handler execution context is request scoped, and it's clear by name that it's a metricsRecorder.

The comment can say instead that the MetricsRecorder is transferred from request options to the context.

packages/core/src/submodules/client/smithy-client/command.ts
```diff
@@ -80,20 +80,30 @@ export abstract class Command<
...
       logger,
       clientName,
       commandName,
       inputFilterSensitiveLog,
       outputFilterSensitiveLog,
+      ...additionalContext,
       [SMITHY_CONTEXT_KEY]: {
         commandInstance: this,
         ...smithyContext,
+        ...additionalSmithyContext,
```
kuhe: now user provided context can override the codegenerated smithyContext values, that probably shouldn't be allowed, so move smithyContext to the end?

packages/core/src/submodules/client/smithy-client/command.ts
```diff
@@ -80,20 +80,30 @@ export abstract class Command<
...
+        ...(options?.metricsRecorder === undefined ? {} : { metricsRecorder: options.metricsRecorder }),
       },
-      ...additionalContext,
     };
     const { requestHandler } = configuration;
     let requestOptions = options ?? {};
+    if (
+      (typeof requestOptions === "object" || typeof requestOptions === "function") &&
+      "metricsRecorder" in requestOptions
```
kuhe: just use 

```ts
if (requestOptions.metricsRecorder) {
  // ...
}
```

if the caller passes a falsy type it won't matter that it was passed to the requestHandler.

packages/core/src/submodules/client/smithy-client/command.ts
vismayku: Reordered the merge so additional context is applied first, followed by the actual command instance and codegenerated Smithy context. The call-site recorder remains authoritative. Added regression coverage for conflicting values.

packages/types/src/middleware.ts
vismayku: Updated the comment to clarify that the MetricsRecorder is transferred from request options into the Smithy context.

packages/core/src/submodules/client/smithy-client/command.ts
vismayku: Updated to the suggested direct truthy check before cloning and removing metricsRecorder from the transport options.

kuhe (approved):

PR #2150, opened 2026-07-15
refactor: use satisfies operator for config and middleware objects

trivikr:
*Issue #, if available:*

N/A

*Description of changes:*

Replace wider type annotations (`const x: T = {...}`) with the
TypeScript 4.9 `satisfies` operator (`const x = {...} satisfies T`)
on config-option objects, middleware-option objects, and the
DEFAULT_PORTS constant map across core submodules and
middleware-compression.

This validates each object against its type while preserving the
literal/narrow types that a plain annotation would widen away.
Contextual typing of inner selectors is retained, and the change is
compile-time only with zero runtime impact.

---


kuhe:
what was the use case for this?

if the inferred type has more information than the interface, it could mean the interface was defined too loosely.

kuhe (approved):

trivikr:
> what was the use case for this?

There was no use case. I was just exploring TypeScript features which we can use.

Closing since satisfies for objects isn't helpful.

PR #2158, opened 2026-07-20
feat(codegen): add types-only codegen

JordonPhillips:
This adds support for types-only codegen driven by Smithy shape closures. This does not add support for mixed-mode generation, but it does take steps in that direction by centralizing generation in a single codegen plugin that takes a mode parameter.

---


kuhe:
😲

smithy-typescript-codegen-test/smithy-build.json
```diff
@@ -144,6 +144,20 @@
...
+          "package": "@smithy/types-example",
+          "packageVersion": "0.0.1",
+          "packageJson": {
+            "license": "Apache-2.0",
+            "private": true
+          },
+          "modes": ["types"],
+          "closure": "example.types#typesClosure"
+        }
+      }
```
kuhe, line 160: in addition, the ./codegen/smithy-typescript-protocol-test-codegen folder should receive a new types-only model, with corresponding declaration in its smithy-build.json. 

Please have the `make generate-protocol-tests` target include deleting and regenerating the sample model(s). 

That app/target generates into `./private/<pkg>`, which is versioned. This gives a reviewable diff of the consequences of the codegen change. It currently has one SSDK (newly added), and two client SDKs in both schema-serde and legacy modes.

JordonPhillips: Protocol tests aren't really appropriate since there are no protocol tests for types-only mode. I'll get something together at least.

kuhe: The name `make generate-protocol-tests` was perhaps misleading. 

It is generating SDKs, some of which come with protocol tests, namely the Smithy RPCv2 CBOR protocol test client and perhaps more in the future, but the other things it generates are for visual diffing, which is why I want to have a types-only package in the group.

yasmewad (approved):

kuhe (approved):

kuhe:
Having the generated client versioned for visual diff is something I can set up.

> This is in draft

Since it's not in draft and it's approved, do you want to merge it now, or make more changes?

JordonPhillips:
@kuhe I was going to add an entry like you suggested. Sorry for the delay on that. It's my next priority

kuhe (approved):

PR #2170, opened 2026-07-21
feat(node-http-handler): pass client logger to request handlers

aBurmeseDev:
_Issue #, if available:_
https://github.com/aws/aws-sdk-js-v3/issues/6130
JS-7027

_Description of changes:_

Passes the client's logger to request handlers as a fallback. Handlers use it for socket exhaustion warnings and timeout messages only if they don't already have their own logger. When the client logger is a `NoOpLogger`, the call is skipped entirely, so handlers keep their console defaults when the client hasn't configured logging.

Also fixes a latent bug in [`httpExtensionConfig.ts`](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts) where `getHttpHandlerExtensionConfig` and `resolveHttpHandlerRuntimeConfig` were reading/writing `runtimeConfig.httpHandler` which clients never populate. Changed to `requestHandler`. Both affected types are `@internal`.

_Testing_

- [x] unit tests for logger injection in `getHttpHandlerExtensionConfig`
- [x] unit tests for `NodeHttpHandler.updateHttpClientConfig` nullish
- [x] unit tests for `UndiciHttpHandler.updateHttpClientConfig`
---


packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
```diff
@@ -25,6 +25,8 @@ export type HttpHandlerExtensionConfigType<HandlerConfig extends object = {}> =
 export const getHttpHandlerExtensionConfiguration = <HandlerConfig extends object = {}>(
   runtimeConfig: HttpHandlerExtensionConfigType<HandlerConfig>
 ) => {
+  runtimeConfig.httpHandler?.updateHttpClientConfig("logger" as keyof HandlerConfig, (runtimeConfig as any).logger);
```
trivikr: Add a truthy check?
```suggestion
  if ((runtimeConfig as any).logger) {
    runtimeConfig.httpHandler?.updateHttpClientConfig("logger" as keyof HandlerConfig, (runtimeConfig as any).logger);
  }
```

packages/undici-http-handler/src/undici-http-handler.spec.ts
```diff
@@ -561,15 +561,22 @@ describe("UndiciHttpHandler", () => {
       expect(configs.logger).toBe(logger);
     });
 
-    it("updates config", async () => {
+    it("does not overwrite an existing logger via updateHttpClientConfig", async () => {
```
trivikr: Why should `updateHttpClientConfig` skip overriding existing logger?

packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
kuhe: the method call must be checked too. `?.`

try requiring `HandlerConfig extends { logger?: Logger }` instead of casting a literal to something it isn't.

This should be backwards compatible.

https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgDIHsDmnrIN4BQyxyANlgBQB0NcUmAzgFzICuIA1iOgO4gDaAXQCULAG7pgAEwDcBAL4ECoSLEQoAShACOrCAzAAJOCCmloAHgDCAPnxESrAA5S4kChwgBPFp6-oYZCsAGmQxOFIWdi5eEFEwyVkFJRh2BDBgdBBkBAjSAFUXNwhrZAgAD0hTBmQKPDIsHCgAfhYMbFx5YRsKAAsTM2gWLV19IwHzKGsbYXsSZH7TSapnV3cAInIOqHXQkFZSUmE5RQIELIMFiaHkEb0DYyXLPHk7AF5kfcPkOBqTLzkZzyhTWED61ygxyAA

```ts
interface Logger {
    log(...args: unknown[]): void;
}

interface RequestHandler<C> {
    update(key: keyof C, val: unknown): void;
}

function callUpdate<C extends ({ logger?: Logger })>(handler: RequestHandler<C>) {
    handler.update("logger", null);
}

const handler: RequestHandler<{}> = null as any;

callUpdate(handler);
```

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -325,6 +325,9 @@ or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler conf
   public updateHttpClientConfig(key: keyof NodeHttpHandlerOptions, value: NodeHttpHandlerOptions[typeof key]): void {
     this.config = undefined;
     this.configProvider = this.configProvider.then((config) => {
+      if (key === "logger") {
+        return { ...config, logger: config.logger ?? (value as NodeHttpHandlerOptions["logger"]) };
+      }
```
kuhe, line 332: this makes it impossible to update the logger once it's set

kuhe (changes requested):
look at additional requestHandler implementations in AWS SDK JS.

at this time, handlers which resolve config asynchronously don't offer a way to query whether there is a higher priority logger set on the requestHandler at requestHandler initialization.

We need a plan to account for this problem.

packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
aBurmeseDev: added the truthy check so we don't call `updateHttpClientConfig` with undefined/null

packages/undici-http-handler/src/undici-http-handler.spec.ts
aBurmeseDev: `updateHttpClientConfig` should always overwrites, removed the nullish assignment.

packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
aBurmeseDev: added `?.` on the method and Logger import with the constraint `(HandlerConfig extends { logger?: Logger })`. 
Also wrapped in a truthy check, so we skip the call entirely when there's no logger.

packages/node-http-handler/src/node-http-handler.ts
aBurmeseDev: removed the nullish assignment. `updateHttpClientConfig` now does a plain overwrite for all keys including logger.

packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
```diff
@@ -22,9 +24,13 @@ export type HttpHandlerExtensionConfigType<HandlerConfig extends object = {}> =
  *
  * @internal
  */
-export const getHttpHandlerExtensionConfiguration = <HandlerConfig extends object = {}>(
+export const getHttpHandlerExtensionConfiguration = <HandlerConfig extends { logger?: Logger }>(
   runtimeConfig: HttpHandlerExtensionConfigType<HandlerConfig>
 ) => {
+  if ((runtimeConfig as any).logger) {
+    runtimeConfig.httpHandler?.updateHttpClientConfig("logger" as keyof HandlerConfig, (runtimeConfig as any).logger);
```
kuhe: This is an unconditional overwrite of the handler logger with the client logger, which is not always intended.

Here are the design requirements:

H = handler, C = client

- if H has no explicit logger and C has a logger (default or otherwise), provide C logger to H logger
  - in other words, if H has an explicit logger, do not overwrite it
- if H resolves its logger asynchronously (NodeHttpHandler), it must still accept the C logger synchronously and save it for later use.
- the system must work with all 5 team-authored hanlders: Fetch, Node, Undici, Xhr, and WebsocketFetch
- no handler level method should be added, such as `setFallbackLogger(Logger)`.

We may want to customize `updateHttpClientConfig` implementations in all 5 handlers such that receiving a special key sets the fallback logger, but do not define this key on the public input config of the handler itself.

Example:

```ts
updateHttpClientConfig(key: symbol, value: unknown): void;
updateHttpClientConfig(key: keyof XhrHttpHandlerOptions, value: XhrHttpHandlerOptions[typeof key]): void;
```

aBurmeseDev: The client now passes its logger to the request handler through `updateHttpClientConfig`, under a new `FALLBACK_LOGGER` key exported from `@smithy/core/protocols`

`node-http-handler` uses it for its socket exhaustion and request timeout warnings when it has no `logger` of its own, replacing `config.logger ?? console` with `config.logger ?? fallbackLogger`

aBurmeseDev: found that this never actually runs, the client offers its logger but `updateHttpClientConfig` call is skipped. I think it's because `getHttpHandlerExtensionConfig` expects the handler at `runtimeConfig.httpHandler` but clients [populate](https://github.com/aws/aws-sdk-js-v3/blob/d6c0ea36225581e94befb619bc5df166fde4ba00/clients/client-s3/src/runtimeConfig.ts#L72) `requestHandler` instead. And since `httpHandler` is never set, the guard `httpHandler?.` on `updateHttpClientConfig` skips the call entirely. Should it move to a point that actually reads requestHandler?

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -200,10 +210,19 @@ export class FetchHttpHandler implements HttpHandler<FetchHttpHandlerOptions> {
...
 
-  updateHttpClientConfig(key: keyof FetchHttpHandlerOptions, value: FetchHttpHandlerOptions[typeof key]): void {
+  updateHttpClientConfig(key: typeof FALLBACK_LOGGER, value: Logger): void;
+  updateHttpClientConfig(key: keyof FetchHttpHandlerOptions, value: FetchHttpHandlerOptions[typeof key]): void;
+  updateHttpClientConfig(
+    key: keyof FetchHttpHandlerOptions | typeof FALLBACK_LOGGER,
+    value: FetchHttpHandlerOptions[keyof FetchHttpHandlerOptions] | Logger
+  ): void {
+    if (key === FALLBACK_LOGGER) {
+      this.fallbackLogger = value as Logger;
```
aBurmeseDev: fallbackLogger gets set here but isn't referenced anywhere else, same for undici. Only `node-http-handler` consumes it.

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -42,6 +48,10 @@ export class NodeHttpHandler implements HttpHandler<NodeHttpHandlerOptions> {
   private configProvider: Promise<ResolvedNodeHttpHandlerConfig>;
   private socketWarningTimestamp = 0;
   private externalAgent = false;
+  /**
+   * Client logger, used only when this handler has no logger of its own.
+   */
+  private fallbackLogger?: Logger;
```
kuhe: don't use a separate field. the logger can be assigned if receiving a fallback logger and no logger exists

packages/core/src/submodules/protocols/protocol-http/fallbackLogger.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Key with which a client offers its logger to an HttpHandler via
+ * `updateHttpClientConfig`, to be used only if the handler has no logger
+ * of its own.
+ *
+ * A symbol keeps this off the handlers' public options types. `Symbol.for`
+ * makes the key shared between duplicate copies of this package.
+ *
+ * @internal
+ */
+export const FALLBACK_LOGGER: unique symbol = Symbol.for("smithy.httpHandler.fallbackLogger");
```
kuhe: `Symbol.for` is the opposite of `unique`. Remove the `unique` type keyword here.

packages/core/src/submodules/protocols/protocol-http/httpHandler.ts
```diff
@@ -16,8 +19,15 @@ export type HttpHandler<HttpHandlerConfig extends object = {}> = RequestHandler<
 > & {
   /**
    * @internal
+   *
+   * The key may also be {@link FALLBACK_LOGGER}, with which a client offers its
+   * logger for use only when the handler has no logger of its own. Handlers
+   * that predate that key ignore it.
```
kuhe: description goes first, above the `@internal` annotation

packages/core/src/submodules/protocols/protocol-http/httpHandler.ts
```diff
@@ -16,8 +19,15 @@ export type HttpHandler<HttpHandlerConfig extends object = {}> = RequestHandler<
...
    */
-  updateHttpClientConfig(key: keyof HttpHandlerConfig, value: HttpHandlerConfig[typeof key]): void;
+  updateHttpClientConfig(
```
kuhe: don't change the signature of this method, I believe this could be breaking for consuming implementers.

packages/core/src/submodules/protocols/index.ts
```diff
@@ -16,6 +16,7 @@ export { SerdeContext } from "./SerdeContext";
 export { Field } from "./protocol-http/Field";
 export { Fields, type FieldsOptions } from "./protocol-http/Fields";
 export { type HttpHandler, type HttpHandlerUserInput } from "./protocol-http/httpHandler";
+export { FALLBACK_LOGGER } from "./protocol-http/fallbackLogger";
```
kuhe: let's not export this, `Symbol.for("string")` can be equality-compared across contexts.

The value can be `Symbol.for("logger")`, since being a symbol is already differentiated from `"logger"` as a normal string.

packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts
aBurmeseDev: `getHttpHandlerExtensionConfig` and `resolveHttpHandlerRuntimeConfig` now use `requestHandler` (the field clients actually set) instead of `httpHandler` (which was always undefined)

This was already broken where `runtimeConfig.httpHandler!` non-null assertion has been silently reading undefined since it was introduced. Nothing surfaced it because nothing called these methods until now.

Both `HttpHandlerExtensionConfigType` and `resolveHttpHandlerRuntimeConfig` return type change shape but both are `@internal`. Noted in the changeset.

packages/core/src/submodules/protocols/protocol-http/fallbackLogger.ts
aBurmeseDev: done, now `Symbol.for("logger")` with no `unique`.

packages/core/src/submodules/protocols/index.ts
aBurmeseDev: no longer exported. Each handler declares its own `Symbol.for("logger")` now.

packages/core/src/submodules/protocols/protocol-http/httpHandler.ts
aBurmeseDev: reverted. The fallback logger key is now handled at each handler's call site.

packages/node-http-handler/src/node-http-handler.ts
aBurmeseDev: `fallbackLogger` field removed, now assigns directly into `config.logger`.

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -16,6 +16,14 @@ import { writeRequestBody } from "./write-request-body";
 
 export type { NodeHttpHandlerOptions };
 
+/**
+ * Key with which a client offers its logger, to be used only if this handler
+ * has no logger of its own. `Symbol.for` makes this equal to the client's copy.
+ *
+ * @internal
+ */
+const FALLBACK_LOGGER: symbol = Symbol.for("logger");
```
kuhe: inline this

aBurmeseDev: inlined in both handlers.

aBurmeseDev: Also inlined it at the offer site in `httpExtensionConfig.ts` which let me drop `fallbackLogger.ts` entirely since it only existed to export that one.

aBurmeseDev (commented):
recap for reviewers: 
passes the client's logger to request handlers as a fallback. Handlers use it for socket exhaustion warnings and timeout messages only if they don't already have their own logger. When the client logger is a `NoOpLogger`, the call is skipped entirely, so handlers keep their console defaults when the client hasn't configured logging.

Also fixes a latent bug in [`httpExtensionConfig.ts`](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/core/src/submodules/protocols/protocol-http/extensions/httpExtensionConfiguration.ts) where `getHttpHandlerExtensionConfig` and `resolveHttpHandlerRuntimeConfig` were reading/writing `runtimeConfig.httpHandler` which clients never populate. Changed to `requestHandler`. Both affected types are `@internal`.

[`FetchHttpHandlerOptions`](https://github.com/smithy-lang/smithy-typescript/blob/eda8e9796c241eef879074440b8fb275e459aa85/packages/types/src/http/h%20%20%20ttpHandlerInitialization.ts#L86-L136) has no logger field and the handler doesn't log anything. So `fetch-http-handler` is unchanged. If we want consistency across, I can add a `logger` field to `FetchHttpHandlerOptions` as a public option.

kuhe (approved):

siddsriv (approved):

Issue #959, opened 2023-09-25
Error: Smithy TypeScript SSDK codegen produces un-buildable code with event streaming


MYoung25:
Minimal Repro: https://github.com/MYoung25/smithy-ts-async-iterable-repro

Steps to reproduce:
1) Clone repo
2) Install Smithy CLI
3) `yarn build`

When using the Smithy CLI Template and having a model with event streaming, the SSDK fails to build with error:
```
src/models/models_0.ts: error TS2345: Argument of type 'AsyncIterable<PublishEvents> | undefined' is not assignable to parameter of type 'PublishEvents | null | undefined'.
   Type 'AsyncIterable<PublishEvents>' is not assignable to type 'PublishEvents | null | undefined'.
```

The [Smithy Model](https://github.com/MYoung25/smithy-ts-async-iterable-repro/blob/main/models/weather.smithy) is the example in the [smithy docs for event streaming](https://smithy.io/2.0/spec/streaming.html?highlight=event%20streaming#event-streams) with added `http` bindings

PR #2175, opened 2026-07-22
fix(codegen): support event streams in server SDK serde context

goelakash:
**The problem:** make a Smithy model with an event stream, generate a server from it, and the code you get is broken. Two reasons:
1. The response side of the stream is typed as `any` — which tells the compiler "don't check this." So that error was there, just hidden. (The request side was typed correctly, which is why the package as a whole already failed to compile.)
2. The server is never handed the thing that actually encodes and decodes stream events (the marshaller). So even at runtime there's nothing to do the work.

Related to #959.

**The fix** — two small changes, both in the code generator. No npm packages touched:
1. The generated response serializer now uses the real type, `__EventStreamSerdeContext`, instead of `any`. Now the compiler checks it for real. (The request side already used the right type.)
2. The generated server now gets a marshaller, built with `eventStreamSerdeProvider`. All it needs is the utf8 helpers the server already had, so nothing new at runtime.

**Who this affects:** only servers generated from models that have event streams — and their code was already broken, so this is strictly an improvement. Everyone else gets byte-for-byte the same output as before. `smithy-aws-typescript-codegen` keeps compiling unchanged; restJson1 will need its own follow-up PR in aws-sdk-js-v3.

**How it was tested:** added the first-ever server event stream codegen tests, the full gradle build passes, and a sample of the new generated output compiles clean under `tsc --strict`. The old output fails that same check with `eventStreamMarshaller is missing` — exactly the error the `any` was hiding.

**Not fixed here:** the rpcv2Cbor-based test in this diff only exercises the handler wiring — that protocol's own server codegen is still stubbed. Stream validation is still skipped (events arrive after up-front validation runs, so supporting it needs a design decision). And there's no live end-to-end test yet — that needs a concrete protocol, which lives in aws-sdk-js-v3.

**Why I'm doing this / where it's heading:** I want to build Smithy services that stream to clients over Server-Sent Events (`text/event-stream`), which smithy-typescript has no story for today (smithy-java gained SSE handling for MCP; TypeScript has nothing). The marshaller is injected through the handler's serde context on purpose: all framing runs through that one object, so a protocol generator that wires in an SSE marshaller instead gets `text/event-stream` framing with the rest of the machinery unchanged. (To be precise: the plug-in point is the protocol-generator layer at codegen time, not a runtime hook in the emitted server.) I'm planning to prototype that as a third-party protocol generator and bring it back as an RFC — this PR is the first step, but it also stands on its own as a bugfix for #959.

### DIFF
--- a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/DirectedTypeScriptCodegen.java
+++ b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/DirectedTypeScriptCodegen.java
@@ -44,6 +44,7 @@ import software.amazon.smithy.model.traits.PaginatedTrait;
 import software.amazon.smithy.model.validation.ValidationEvent;
 import software.amazon.smithy.typescript.codegen.auth.http.HttpAuthSchemeProviderGenerator;
 import software.amazon.smithy.typescript.codegen.endpointsV2.EndpointsV2Generator;
+import software.amazon.smithy.typescript.codegen.integration.AddEventStreamDependency;
 import software.amazon.smithy.typescript.codegen.integration.ProtocolGenerator;
 import software.amazon.smithy.typescript.codegen.integration.RuntimeClientPlugin;
 import software.amazon.smithy.typescript.codegen.integration.TypeScriptIntegration;
@@ -258,9 +259,10 @@ final class DirectedTypeScriptCodegen
         }
 
         if (settings.generateServerSdk()) {
+            boolean hasEventStream = AddEventStreamDependency.hasEventStream(directive.model(), service);
             for (OperationShape operation : directive.operations()) {
                 delegator.useShapeWriter(operation, w -> {
-                    ServerGenerator.generateOperationHandler(symbolProvider, service, operation, w);
+                    ServerGenerator.generateOperationHandler(symbolProvider, service, operation, w, hasEventStream);
                 });
             }
         }
@@ -736,7 +738,13 @@ final class DirectedTypeScriptCodegen
             .useShapeWriter(service, writer -> {
                 ServerGenerator.generateOperationsType(symbolProvider, service, operations, writer);
                 ServerGenerator.generateServerInterfaces(symbolProvider, service, operations, writer);
-                ServerGenerator.generateServiceHandler(symbolProvider, service, operations, writer);
+                ServerGenerator.generateServiceHandler(
+                    symbolProvider,
+                    service,
+                    operations,
+                    writer,
+                    AddEventStreamDependency.hasEventStream(directive.model(), service)
+                );
             });
     }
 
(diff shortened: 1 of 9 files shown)


Miles123K:
For this PR or the next [EventStreamGenerator](https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/EventStreamGenerator.java) will also require updates. It currently only supports
```
generateEventStreamSerializers: calls hasEventStreamInput and getEventStreamInputShape
generateEventStreamDeserializers: calls hasEventStreamOutput and getEventStreamOutputShape
```

We would need to make a change like

```
public enum EventStreamDirection {
    INPUT,
    OUTPUT
}

/**
 * Ensure backward compatible public interface
 */
public void generateEventStreamSerializers(
    GenerationContext context,
    ServiceShape service,
    String documentContentType,
    Runnable serializeEventDocumentPayload,
    Set<Shape> documentShapesToSerialize
) {
    generateEventStreamSerializers(
        context,
        service,
        EventStreamDirection.INPUT,
        documentContentType,
        serializeEventDocumentPayload,
        documentShapesToSerialize
    );
}

/**
 * Generates serializers for event streams in the selected operation direction.
 */
public void generateEventStreamSerializers(
    GenerationContext context,
    ServiceShape service,
    EventStreamDirection direction,
    String documentContentType,
    Runnable serializeEventDocumentPayload,
    Set<Shape> documentShapesToSerialize
) {
    Model model = context.getModel();
    Set<OperationShape> operations =
        TopDownIndex.of(model).getContainedOperations(service);

    TreeSet<UnionShape> eventUnionsToSerialize = new TreeSet<>();
    TreeSet<Pair<String, StructureShape>> eventShapesToMarshall =
        new TreeSet<>(
            (a, b) -> Objects.compare(
                a.getRight(),
                b.getRight(),
                StructureShape::compareTo
            )
        );

    for (OperationShape operation : operations) {
        boolean hasEventStream = switch (direction) {
            case INPUT -> hasEventStreamInput(context, operation);
            case OUTPUT -> hasEventStreamOutput(context, operation);
        };

        if (!hasEventStream) {
            continue;
        }

        UnionShape eventsUnion = switch (direction) {
            case INPUT -> getEventStreamInputShape(context, operation);
            case OUTPUT -> getEventStreamOutputShape(context, operation);
        };

        eventUnionsToSerialize.add(eventsUnion);

        eventsUnion.members().forEach(member -> {
            StructureShape event = model.expectShape(
                member.getTarget(),
                StructureShape.class
            );

            eventShapesToMarshall.add(
                Pair.of(member.getMemberName(), event)
            );
        });
    }

    eventUnionsToSerialize.forEach(eventsUnion ->
        generateEventStreamSerializer(context, eventsUnion)
    );

    SerdeElisionIndex serdeElisionIndex = SerdeElisionIndex.of(model);

    eventShapesToMarshall.forEach(memberAndEvent ->
        generateEventMarshaller(
            context,
            memberAndEvent.getLeft(),
            memberAndEvent.getRight(),
            documentContentType,
            serializeEventDocumentPayload,
            documentShapesToSerialize,
            serdeElisionIndex
        )
    );
}
```

SSDK response generation uses:

```java
generateEventStreamSerializers(
    context,
    service,
    EventStreamDirection.OUTPUT,
    contentType,
    serializer,
    shapes
);
```

kuhe:
event streams are going to be handled differently going forward, as JS runtime code and not codegen.

PR #2189, opened 2026-07-27
ci: cache npm package existence record

trivikr:
_Issue #, if available:_

Noticed in JS-7089

_Description of changes:_

Replace the checked-in `published-packages.json`, which had to be edited
by hand whenever a package was added, with a GitHub Actions cache entry.
A push to main refreshes the record and the release PR check restores it
read-only, because a cache written by a pull_request run is scoped to
that PR's merge ref and cannot be read by the next release PR.

The cache is only an optimization. On a miss every package to publish is
verified against the registry, so the check never depends on it being
present or complete. The key hashes the `CHANGELOG.md` files under
`packages/` and `smithy-typescript-ssdk-libs/`, which change when a package
is added, removed or renamed but stay put for dependency bumps and other
`package.json` edits.

Move the check out of ci.yml into `npm-package-existence.yml`, alongside
the refresh job that shares its cache configuration, and split the script
into `scripts/npm-packages/ `with one entrypoint per job over a shared
module.

---


kuhe (approved):

trivikr:
Record was created on merge

```console
✅ Recorded 62 of 62 publishable package(s) as published on npm (62 newly verified via registry) in /home/runner/work/_temp/published-packages/record.json.
```

https://github.com/smithy-lang/smithy-typescript/actions/runs/30384905211/job/90361637695

Verified that the cache `published-packages-v1-53dfd0b6e0ea8966c28c5683120782736ec57754135fe4fa2767e61e3749f44d` exists on https://github.com/smithy-lang/smithy-typescript/actions/caches?query=sort%3Acreated-desc

PR #2183, opened 2026-07-24
feat(core/cbor): add single-pass cbor serde

kuhe:
Add single-pass ShapeSerializer and ShapeDeserializer for CBOR, combining Smithy data-model prep with the encoding pass.

## Serialization Results

| Test Case                         | CBOR Size | JSON Size | CBOR multi ms | CBOR single ms | JSON ms | Single vs Multi | Single vs JSON |
| --------------------------------- | --------: | --------: | ------------: | -------------: | ------: | --------------: | -------------: |
| `list<string(0,180)>`             |    469 kb |    482 kb |         53 ms |          24 ms |   68 ms |           2.19x |          2.79x |
| `list<float>`                     |    270 kb |    675 kb |        220 ms |          27 ms |  308 ms |           8.24x |         11.55x |
| `list<int>`                       |    253 kb |    458 kb |        693 ms |          92 ms |  277 ms |           7.54x |          3.01x |
| `list<long int>`                  |    250 kb |    549 kb |        360 ms |          56 ms |  245 ms |           6.46x |          4.40x |
| `map<string, string>`             |    438 kb |    455 kb |         42 ms |          30 ms |   75 ms |           1.41x |          2.54x |
| `map<string, long int>`           |     38 kb |     49 kb |         25 ms |          12 ms |   22 ms |           2.08x |          1.81x |
| `list<struct> PutMetricData-like` |    363 kb |    459 kb |        344 ms |         121 ms |  385 ms |           2.84x |          3.17x |
| `struct PutMetricData realistic`  |    1.7 mb |    2.2 mb |       1846 ms |         459 ms | 1814 ms |           4.02x |          3.95x |
| `list<struct> non-ASCII keys`     |    399 kb |    483 kb |        239 ms |          93 ms |  280 ms |           2.56x |          3.00x |
| `list<struct> with blobs`         |    263 kb |    353 kb |        140 ms |          70 ms |  116 ms |           2.01x |          1.66x |
| `list<struct> with timestamps`    |    229 kb |    345 kb |        227 ms |          83 ms |  167 ms |           2.75x |          2.02x |

## Deserialization Results

| Test Case                         | Data Size | CBOR multi ms | CBOR single ms | JSON ms | Single vs Multi | Single vs JSON |
| --------------------------------- | --------: | ------------: | -------------: | ------: | --------------: | -------------: |
| `list<string(0,180)>`             |    468 kb |         34 ms |          24 ms |   25 ms |           1.45x |          1.08x |
[...]

PR #2190, opened 2026-07-28
feat(core/cbor): default to v2 cbor codec

kuhe:
following https://github.com/smithy-lang/smithy-typescript/pull/2183, this sets the default CBOR codec to v2.


packages/core/src/submodules/schema/schemas/NormalizedSchema.ts
```diff
@@ -504,17 +504,6 @@ export class NormalizedSchema implements INormalizedSchema {
...
-  /**
-   * CBOR-optimized struct iteration. Returns a cache of parallel arrays
-   * (memberNames, memberSchemas, encodedKeys) for direct indexed iteration.
-   * Implementation is patched by the cbor submodule on load.
-   *
-   * @internal
-   */
-  public structIteratorCbor(): { memberNames: string[]; memberSchemas: NormalizedSchema[]; encodedKeys: Uint8Array[] } {
-    throw new Error("@smithy/core/schema - structIteratorCbor not loaded.");
-  }
```
kuhe, line 517: ok to remove - was only here for one minor version, internal, and had no implementation.

trivikr (dismissed):

trivikr (approved):

PR #2191, opened 2026-07-28
ci: check superseded versions exist on npm (attempt 1)

trivikr:
_Issue #, if available:_

Internal JS-7089

_Description of changes:_

The release workflow publishes only the versions a release PR adds. If a
previous release never reached the registry, merging the next release PR
bumps past the versions it should have published: they stay in git, in the
changelogs and in the tags, but never appear on npm, and no later release
will publish them.

Extend the release PR check so that, for every package it publishes, the
version at the branch's fork point from main must exist on npm. Failures
are reported in two groups, because they need different fixes: packages
never published at all still need their one-time manual first publish,
while a published package missing the version on main means the last
release did not finish.

The cached record now maps each confirmed package name to the version of
it most recently confirmed published, so both halves of the check can skip
registry queries. Only main's current version is ever looked up, so one
version per package is enough. Bump the cache key prefix to v2 for the new
shape.

---

### DIFF
--- a/.github/workflows/npm-package-existence.yml
+++ b/.github/workflows/npm-package-existence.yml
@@ -2,14 +2,16 @@ name: npm-package-existence
 
 on:
   # Verify the changesets release PR: every package it will publish must already
-  # exist on npm, since trusted publishing (OIDC) cannot create a new name.
+  # exist on npm, since trusted publishing (OIDC) cannot create a new name, and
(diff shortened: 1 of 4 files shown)


trivikr:
There was a bug in this PR, where the record was updated before packages were published on npm.

It's fixed in https://github.com/smithy-lang/smithy-typescript/pull/2192

PR #2192, opened 2026-07-28
ci: check superseded versions exist on npm

trivikr:
_Issue #, if available:_

Internal JS-7089

_Description of changes:_

The release workflow publishes only the versions a release PR adds. If a
previous release never reached the registry, merging the next release PR
bumps past the versions it should have published: they stay in git, in the
changelogs and in the tags, but never appear on npm, and no later release
will publish them.

Extend the release PR check so that, for every package it publishes, the
version at the branch's fork point from main must exist on npm. Failures
are reported in two groups, because they need different fixes: packages
never published at all still need their one-time manual first publish,
while a published package missing the version on main means the last
release did not finish.

The cached record now maps each confirmed package name to the version of
it most recently confirmed published, so both halves of the check can skip
registry queries. Only main's current version is ever looked up, so one
version per package is enough, and the cache key prefix moves to v2 for
the new shape. Refresh the record when release-npm-packages completes
rather than on push to main: a push refresh runs alongside that release
and would record the versions it is publishing as absent.

---

### DIFF
--- a/.github/workflows/npm-package-existence.yml
+++ b/.github/workflows/npm-package-existence.yml
@@ -2,17 +2,27 @@ name: npm-package-existence
 
 on:
   # Verify the changesets release PR: every package it will publish must already
-  # exist on npm, since trusted publishing (OIDC) cannot create a new name.
+  # exist on npm, since trusted publishing (OIDC) cannot create a new name, and
+  # every version it bumps past must already be published, since merging it means
+  # no later release will ever publish those.
   pull_request:
     branches:
       - main
-  # Refresh the cached record of packages known to exist on npm. Only push and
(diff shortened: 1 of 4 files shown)


kuhe (approved):

trivikr:
Verified that the check works without cache

```console
Cache not found for input keys: published-packages-v2-1813ec8ac5d1157a53c0d378310bbd63e462f1f0b6848559f7a9feca078bc78d
...
✅ All 59 package(s) to publish exist on npm, and the 59 version(s) they supersede are published: 0 check(s) from the cached record, 118 verified via registry.
```

https://github.com/smithy-lang/smithy-typescript/actions/runs/30398811559/job/90409024557?pr=2193

PR #2194, opened 2026-07-28
ci: fold npm name check into the version check

trivikr:
_Issue #, if available:_

Refs: https://github.com/smithy-lang/smithy-typescript/pull/2192#issuecomment-5109680020

_Description of changes:_

The release PR check asked the registry two questions per package: does
the name exist, and is the version it supersedes published. A published
version already implies a published name, so the name probe was
redundant on every passing check - 118 requests for 59 packages.

Collapse both into one list of checks, where a null version means the
package is new in this PR and only its name can be checked. The name is
probed on its own only when a version comes back missing, to keep the
two failures distinguishable: a package that was never published needs a
manual first publish plus trusted publisher setup, while an absent
version means an earlier release never finished. partitionByExistence
now covers both cases and reports missingNames separately from
missingVersions; partitionVersionsByExistence is gone, and the record
refresh merges the two buckets since it leaves anything unconfirmed
unrecorded either way.

Replace isPublishable with readPublishablePackage, which reads a
manifest and requires a name and version unless it is private. A
publishable package missing either was previously skipped, dropping it
silently from both the record and the release check; it now fails with
the manifest path and what to fix.

---

### DIFF
--- a/scripts/npm-packages/ensure-packages-exist-on-npm.mts
+++ b/scripts/npm-packages/ensure-packages-exist-on-npm.mts
@@ -33,9 +33,10 @@
  * the one being superseded; a package the branch adds supersedes nothing, so only
  * its name is checked.
  *
- * Packages and versions already recorded as published (see shared.mts) skip their
- * registry check; the rest are queried. Runs directly via Node type stripping
- * (Node >= 24, no build step).
+ * Each package costs one registry check: the version it supersedes, since a
+ * published version implies a published name. Checks already answered by the
+ * record (see shared.mts) are skipped; the rest are queried. Runs directly via
+ * Node type stripping (Node >= 24, no build step).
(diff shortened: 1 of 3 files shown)


siddsriv (approved):

trivikr:
Validated that one call is sent to npm per package, if cache is not available

```console
Cache not found for input keys: published-packages-v2-1813ec8ac5d1157a53c0d378310bbd63e462f1f0b6848559f7a9feca078bc78d
...
✅ All 59 package(s) to publish exist on npm, and the 59 version(s) they supersede are published: 0 check(s) from the cached record, 59 verified via registry.
```
https://github.com/smithy-lang/smithy-typescript/actions/runs/30408077428/job/90438625176?pr=2193

PR #2196, opened 2026-07-29
ci: hand off npm record refresh via dispatch

trivikr:
_Issue #, if available:_

The refresh of the cached record of published npm packages ran on workflow_run and could no longer save the cache:
```console
Failed to save: Unable to reserve cache with key published-packages-v2-1813ec8ac5d1157a53c0d378310bbd63e462f1f0b6848559f7a9feca078bc78d. More details: cache write denied: token has no writable scopes
```
Workflow run: https://github.com/smithy-lang/smithy-typescript/actions/runs/30466471816

_Description of changes:_

GitHub now issues read-only cache tokens to runs that write to the default branch's cache scope from an untrusted trigger, and workflow_run is untrusted because a fork pull request can cascade into it. The save was denied, so the record was never written and every release PR check fell back to probing the registry for every package.

Have `release-npm-packages.yml` raise a `repository_dispatch` event once it has published instead. That trigger keeps read-write access to the scope, `GITHUB_TOKEN` may raise it with the `contents: write` the release job already holds, and it always runs the default branch - so the ordering the refresh depends on, running only after the publish finished, is preserved. The hand-off is best-effort, since failing it would fire the release failure notification for a release that succeeded.

---

### DIFF
--- a/.github/workflows/npm-package-existence.yml
+++ b/.github/workflows/npm-package-existence.yml
@@ -13,16 +13,19 @@ on:
   # on the registry by the time they are probed. Refreshing on push to main instead
   # would race that release and record its versions as absent.
   #
-  # workflow_run also happens to be the trigger this needs for cache access: it
-  # runs in the context of the default branch, and only trusted triggers like it
-  # may write to the default branch's cache scope, which is the scope pull request
-  # runs restore from. The record is therefore always written from main and never
-  # from a pull request.
-  workflow_run:
-    workflows:
-      - release-npm-packages
+  # release-npm-packages.yml raises this event, rather than this workflow listening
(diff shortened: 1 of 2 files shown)


kuhe (approved):

trivikr:
The `repository_dispatch` was run using API

```console
gh api "repos/smithy-lang/smithy-typescript/dispatches" -f event_type=npm-release-published
```

The record was populated
```console
✅ Recorded 62 of 62 publishable package(s) as published on npm (62 newly verified via registry) in /home/runner/work/_temp/published-packages/record.json.
```
Workflow run: https://github.com/smithy-lang/smithy-typescript/actions/runs/30478383438

PR #2205, opened 2026-08-04
chore(server-common): bring SSDK code into packages folder

kuhe:
move SSDK to `./packages`, applying standards and validators from the main package codebase


packages/core/src/submodules/checksum/hash-stream-node/readableStreamHasher.spec.ts
```diff
@@ -81,7 +81,7 @@ describe(readableStreamHasher.name, () => {
     const expectedError = new Error("Unable to calculate hash for flowing readable stream");
     try {
       readableStreamHasher(mockHashCtor, readableStream);
-      fail(`expected ${expectedError}`);
+      expect.fail(`expected ${expectedError}`);
```
kuhe, line 84: `fail` was never working, it was a global type from jest that passed test typechecking due to jest being installed.

trivikr (approved):

PR #2220, opened 2026-08-10
chore: upgrade to typescript 7

kuhe:
use typescript 7

- removes the unused service client doc generator because it uses typedoc
- removes references to typedoc in packages
- make typescriptVersion configurable by codegen consumers


smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/PackageJsonGenerator.java
```diff
@@ -81,34 +80,6 @@ static void writePackageJson(
             );
         }
 
-        if (settings.generateTypeDoc()) {
```
trivikr, line 84: The `generateTypeDoc` setting was added in https://github.com/smithy-lang/smithy-typescript/pull/1091

It had hard dependency on older TypeScript versions, so we've decided to drop it.

We do not know of any internal or external customers which are using the configuration. It's not documented in our README. Since smithy-typescript code generators are not GA yet, we can remove this. Customer who want to use typedoc can always configure it in their `smithy-build.json` configuration.

trivikr (approved):

trivikr (dismissed):
Blocking this PR merge as the previous release needs update in changesets https://github.com/smithy-lang/smithy-typescript/pull/2227#pullrequestreview-4932367889

Will unblock once release is complete.

trivikr (approved):

trivikr (changes requested):
Blocking since changesets v3 resulted in release failure in https://github.com/smithy-lang/smithy-typescript/actions/runs/31754700865/job/94627877936?pr=2227

Will unblock once release is successful.

trivikr (approved):
Release blocker (from changesets v3) is now fixed. This PR will need a rebase though.

PR #2222, opened 2026-08-11
chore(deps-dev): migrate changesets to v3

trivikr:
_Issue #, if available:_

https://changesets.dev/blog/announcing-changesets-v3

_Description of changes:_

Upgrade @changesets/cli to v3 and changesets/action to v2, which is the version of the action compatible with Changesets v3.

The v3 config needed no changes beyond the schema pin, which now points at the copy in node_modules so it tracks the installed version. `access: "public"` is still valid, and `baseBranch` and `privatePackages` now match v3's defaults; they are kept because `npm-package-existence.yml` cites `privatePackages.version` by name.

The v2 action renames its inputs and outputs: version, publish, commit and title become version-script, publish-script, commit-message and pr-title, and pullRequestNumber becomes pr-number. It also no longer reads the GITHUB_TOKEN environment variable, so the token is passed through the github-token input. Its branching is unchanged, so both release workflows behave as before.

---

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "$schema": "../node_modules/@changesets/config/schema.json",
   "changelog": "@changesets/cli/changelog",
   "commit": false,
   "fixed": [],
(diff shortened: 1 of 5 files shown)


trivikr:
Posted as draft, since `@changesets/cli@3.0.0` is newer than the repository's 24 hour npmMinimalAgeGate.
It'll be rebased and marked ready for review on 8/12.

kuhe (approved):

siddsriv (approved):

Issue #2202, opened 2026-08-04
NodeHttp2Handler: isolated HTTP/2 sessions for event-stream requests are never destroyed, pinning the request's async context


taoche:
### Checkboxes for prior research

- [x] I've gone through Developer Guide and API reference
- [x] I've checked AWS Forums and StackOverflow.
- [x] I've searched for previous similar issues (`http2 session`, `isolated session`, `ClientHttp2Session` in this repo and aws/aws-sdk-js-v3) and didn't find any solution.

### Describe the bug

`NodeHttp2Handler` creates an **isolated HTTP/2 session per event-stream request** (`useIsolatedSession = disableConcurrentStreams || isEventStream`). In our production workload (Bedrock `ConverseStreamCommand` via `@aws-sdk/client-bedrock-runtime`), these isolated sessions are **never destroyed**: exactly one live `ClientHttp2Session` + `TLSSocket` pair accumulates per streaming call, for the lifetime of the process.

The impact is far larger than the socket itself. Node keeps a strong Global handle to the `AsyncContextFrame` that was current when a native handle was created (Node 24, `AsyncLocalStorage` on AsyncContextFrame). The undestroyed session therefore pins the **entire async context of the request** — in our case each LangGraph agent run's full message graph (~250KB/run), producing a steady process-level heap leak that survived every application-level fix until we found this anchor.

Heap-snapshot retainer chain (Chrome DevTools, `--expose-gc`, double GC before snapshot):

```
(GC roots) → (Global handles) --internal-->
  AsyncContextFrame --table-->
    <application ALS entries: langsmith RunTree → config → closures → PregelLoop → messages>
```

and per leaked session:

```
Node / Environment → Node / PrincipalRealm → Node / BaseObjectList
  → Node / Http2Session [native] → ClientHttp2Session {_host: "bedrock-runtime.us-east-1.amazonaws.com"}
```

Repro accounting: after 13 streaming calls, heap snapshot diff shows exactly **+13 `ClientHttp2Session`**, all with `servername: bedrock-runtime.us-east-1.amazonaws.com`, all reachable only from `BaseObjectList` (i.e. still-open native handles), plus +13 of each object class belonging to the per-request async context.

[...]

PR #2223, opened 2026-08-11
fix(node-http-handler): destroy isolated HTTP/2 sessions and type timeout errors

siddsriv:
_Issue #, if available:_
#2202 

_Description of changes:_
- Add stream 'end' listener to destroy isolated sessions when body ends
- Type close-without-response error as TimeoutError when session was destroyed by timeout
- Arm a default 5-minute idle timeout on isolated sessions as a fallback

---

### DIFF
--- a/packages/node-http-handler/src/node-http2-connection-manager.ts
+++ b/packages/node-http-handler/src/node-http2-connection-manager.ts
@@ -116,9 +116,11 @@ export class NodeHttp2ConnectionManager implements ConnectionManager<ClientHttp2
     session.on("frameError", ensureDestroyed);
     session.on("close", ensureDestroyed);
 
-    if (connectionConfiguration.requestTimeout) {
-      session.setTimeout(connectionConfiguration.requestTimeout, ensureDestroyed);
-    }
+    // Arm a session timeout as a safety net to reap orphaned sessions.
+    // Uses the configured sessionTimeout if provided, otherwise defaults to
+    // 5 minutes to prevent indefinite leaks when other cleanup paths fail.
+    const timeout = connectionConfiguration.requestTimeout ?? 300_000;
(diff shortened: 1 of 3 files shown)


packages/node-http-handler/src/node-http2-connection-manager.ts
```diff
@@ -116,9 +116,11 @@ export class NodeHttp2ConnectionManager implements ConnectionManager<ClientHttp2
     session.on("frameError", ensureDestroyed);
     session.on("close", ensureDestroyed);
 
-    if (connectionConfiguration.requestTimeout) {
-      session.setTimeout(connectionConfiguration.requestTimeout, ensureDestroyed);
-    }
+    // Arm a session timeout as a safety net to reap orphaned sessions.
+    // Uses the configured sessionTimeout if provided, otherwise defaults to
+    // 5 minutes to prevent indefinite leaks when other cleanup paths fail.
+    const timeout = connectionConfiguration.requestTimeout ?? 300_000;
```
siddsriv, line 122: lmk if any objections to a 5 min limit, or if you have recommendations

kuhe: is this safe w.r.t. the upcoming default timeouts specification?

siddsriv: yeah i think so, going to re-confirm

siddsriv: it should not be a problem in terms of interfering with upcoming updates. For reviewers: If you can think of a case/service, where this timeout might be short enough to interrupt legitimate operations, please flag it to me. Otherwise, this should be ready.

trivikr (approved):

PR #2225, opened 2026-08-11
ci: wait out npm's publish-time malware scan

trivikr:
_Issue #, if available:_

https://github.blog/changelog/2026-07-28-npm-publish-time-malware-scanning-and-dual-use-metadata/

Noticed in previous run logs where six versions weren't done by malware scanning, and thus not visible on npm view
```console
ℹ️  6 version(s) on main are not on npm, either because the release publishing them has not finished or because the package has never been published at all:
  @smithy/chunked-blob-reader@5.5.0
  @smithy/hash-stream-node@4.5.0
  @smithy/signature-v4@5.7.0
  @smithy/util-buffer-from@4.5.0
  @smithy/util-defaults-mode-browser@4.6.0
  @smithy/util-retry@4.6.0
✅ Recorded 62 of 62 publishable package(s) as published on npm (56 newly verified via registry) in /home/runner/work/_temp/published-packages/record.json.
```
Details: https://github.com/smithy-lang/smithy-typescript/actions/runs/31518075558/job/93868079304

_Description of changes:_

npm now scans every publish for malware before making the version installable, so for several minutes after a release the registry answers 404 for versions it did publish, indistinguishable from versions never published at all.

The record refresh runs inside that window, so it would record none of what the release just published and leave every check on the next release PR falling back to the registry. It now re-checks anything the registry does not have yet until it appears or a 25 minute budget runs out (`NPM_PUBLISH_SCAN_WAIT_MS` overrides it, 0 to skip waiting), with the workflow job capped at 40 minutes so a hang is cut short. The existence check's failure message now tells maintainers to re-run before treating a fresh release's versions as lost.

---


siddsriv (approved):

trivikr:
Last [release](https://github.com/smithy-lang/smithy-typescript/pull/2224) published only three versions, none of which had to wait for malware scanning to complete.
The changes in this PR can be validated when we publish 50+ versions in a release.

It did refresh the six versions which missed in the record though.

```console
✅ Recorded 62 of 62 publishable package(s) as published on npm (9 newly verified via registry) in /home/runner/work/_temp/published-packages/record.json.
```
Logs: https://github.com/smithy-lang/smithy-typescript/actions/runs/31624774326/job/94208252745

Issue #2182, opened 2026-07-24
Error in toEndpointV1 method


Nefcanto:
I came here because of this bug:

```
 2026-07-22T14:17:59.244Z Command {
     "Bucket": "someProject",
     "Key": "post/205ef9cd-54ac-4cdf-b400-d985d6315a02.webp",
     "ContentType": "image/webp",
     "ACL": "public-read"
 }
 2026-07-22T14:17:59.269Z Upload failed: post/205ef9cd-54ac-4cdf-b400-d985d6315a02.webp
 2026-07-22T14:17:59.269Z TypeError: endpoint.headers[name].join is not a function
     at toEndpointV1 (/home/someUser/someProject/adminApi/node_modules/@smithy/core/dist-cjs/submodules/transport/index.js:164:85)
     at Object.endpoint (/home/someUser/someProject/adminApi/node_modules/@smithy/core/dist-cjs/submodules/schema/index.js:85:23)
     at S3RestXmlProtocol.serializeRequest (/home/someUser/someProject/adminApi/node_modules/@smithy/core/dist-cjs/submodules/protocols/index.js:189:40)
     at S3RestXmlProtocol.serializeRequest (/home/someUser/someProject/adminApi/node_modules/@aws-sdk/core/dist-cjs/submodules/protocols/index.js:1821:37)
     at S3RestXmlProtocol.serializeRequest (/home/someUser/someProject/adminApi/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/submodules/s3/index.js:495:37)
     at /home/someUser/someProject/adminApi/node_modules/@smithy/core/dist-cjs/submodules/schema/index.js:87:43
     at /home/someUser/someProject/adminApi/node_modules/@aws-sdk/checksums/dist-cjs/submodules/flexible-checksums/index.js:293:12
     at async /home/someUser/someProject/adminApi/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/submodules/s3/index.js:62:28
     at async /home/someUser/someProject/adminApi/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/submodules/s3/index.js:89:20
     at async /home/someUser/someProject/adminApi/node_modules/@aws-sdk/core/dist-cjs/submodules/client/index.js:125:26
     at async Object.upload (file:///home/someUser/someProject/adminApi/node_modules/core/cloud/aws/storage.js:75:13)
[...]

PR #2230, opened 2026-08-14
fix(core): use Object.keys() in toEndpointV1 to avoid inherited property iteration

aBurmeseDev:
_Issue #, if available:_
https://github.com/aws/aws-sdk-js-v3/issues/8115
https://github.com/smithy-lang/smithy-typescript/issues/2182

_Description of changes:_

The `for...in` loop in `toEndpointV1` iterates inherited enumerable properties from `Object.prototype`, causing a TypeError when any library extends `Object.prototype` with enum properties.
  
Reverts to `Object.keys(`) which only iterates own properties, matching the original behavior before the perf optimization in #1961.

---

### DIFF
--- a/.changeset/fix-toEndpointV1-prototype.md
+++ b/.changeset/fix-toEndpointV1-prototype.md
@@ -1,0 +1,5 @@
+---
+"@smithy/core": patch
+---
+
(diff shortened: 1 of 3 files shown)


kuhe:
this will be covered by https://github.com/smithy-lang/smithy-typescript/pull/2232

PR #2238, opened 2026-08-17
feat(dynamic-client): add package for creating clients from Smithy JSON AST

kuhe:
this is an experiment


kuhe:
moved to https://github.com/smithy-lang/smithy-typescript/pull/2242

PR #2241, opened 2026-08-17
ci: skip the ci workflow on the release PR

trivikr:
_Issue #, if available:_

N/A

_Description of changes:_

The changesets release PR only bumps versions, rewrites the inter-package dependency ranges to match, writes changelogs and deletes the consumed changesets. It touches no Java or TypeScript source, and `release-dry-run.yml` already installs, builds and packs that exact tree.

Guard every job rather than the trigger, since a pull_request event's branch filters match the base branch and there is no workflow-level if. This makes the step-level guard on the changeset status check redundant, so it moves up to the job.

The run is still created and still queued for approval before any of these conditions are evaluated, so ci stays in EXPECTED_WORKFLOWS in create-release-pr.yml, which is what approves it. It now completes as skipped rather than running the Java matrix, five Node test legs, protocol tests and docs extraction.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,10 +9,15 @@ on:
 permissions:
   contents: read
 
+# Every job is guarded with `if: github.head_ref != 'changeset-release/main'` so
+# that none of them run on the release PR, which only bumps versions and writes
+# changelogs; release-dry-run.yml installs, builds and packs that tree already.
+
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


trivikr:
During post-scrum on 8/18, we decided to retain `ci` as it takes max 7 mins.
If publishing is urgent, we can always not wait for `ci` to succeed/fail before merging.

PR #2243, opened 2026-08-19
fix(core/cbor): handle exponent notation when serializing NumericValue containers

kuhe:
fixes CBOR serialization to handle when the NumericValue string container has a value like `1.23E100` (exponential). 

It also fixes precision handling when the exponent is itself a bigInteger (for excessively large numbers).

### DIFF
--- a/packages/core/src/submodules/cbor/cbor-decode.ts
+++ b/packages/core/src/submodules/cbor/cbor-decode.ts
@@ -377,25 +377,41 @@ function decodeTagValue(
     return minor === 3 ? -b - BigInt(1) : b;
   } else if (minor === 4) {
     const decimalFraction = decode(at + offset, to);
-    const [exponent, mantissa] = decimalFraction;
+    const [rawExponent, mantissa] = decimalFraction;
     const normalizer = mantissa < 0 ? -1 : 1;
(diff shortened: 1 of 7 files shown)


siddsriv (approved):

packages/core/src/submodules/cbor/cbor-decode.ts
```diff
@@ -377,25 +377,41 @@ function decodeTagValue(
...
-    numericString = numericString.replace(/^0+/g, "");
-    if (numericString === "") {
-      numericString = "0";
-    }
-    if (numericString[0] === ".") {
-      numericString = "0" + numericString;
+    const isSmallExponent = typeof rawExponent === "number" && Math.abs(rawExponent) <= 1e15;
+    if (isSmallExponent) {
+      const exponent = rawExponent as number;
+      const mantissaStr = "0".repeat(Math.abs(exponent) + 1) + mantissaDigits;
```
smilkuri, line 391: Wouldn't this throw rangeError if it's greater than 2^29, and 1e15 is greater than that right?

kuhe: boundaries set to 2 ** 28

smilkuri (approved):

